PR #23, opened 2024-02-07
backslash-menu

### LINKED ISSUE #16: Backslash Popup Menu
When the user types a backslash, popup a menu to pick the marker:

image

samueljd:
Fixes #16  
- Backslash menu for platform
- currently opens the `BlockFormatDropDown` toolbar but any component can be passed to it as a child.

### DIFF
--- a/packages/platform/src/editor/plugins/BackslashPlugin.tsx
+++ b/packages/platform/src/editor/plugins/BackslashPlugin.tsx
@@ -1,0 +1,83 @@
+import { computePosition } from "@floating-ui/dom";
+import { useLexicalComposerContext } from "@lexical/react/LexicalComposerContext";
+import { $getSelection, $isRangeSelection } from "lexical";
+import { useCallback, useEffect, useRef, useState } from "react";
+import { createPortal } from "react-dom";
+import { BackslashMenu, BackslashMenuCoords } from "../components/BackslashMenu/BackslashMenu";
+import { useBackslashMenu } from "../components/BackslashMenu/useBackslashMenu";
+
+const DOM_ELEMENT = document.body;
+
+export default function FloatingMenuPlugin() {
+  const ref = useRef<HTMLDivElement>(null);
+  const [coords, setCoords] = useState<BackslashMenuCoords>(undefined);
+  const [editor] = useLexicalComposerContext();
+  const [backslashOpen, setBackslashOpen] = useState(false);
+
+  const { isKeyDown, isKeyReleased } = useBackslashMenu();
+
+  const calculatePosition = useCallback(() => {
+    const domSelection = getSelection();
+    const domRange = domSelection?.rangeCount !== 0 && domSelection?.getRangeAt(0);
+    if (!domRange || !ref.current) {
+      return setCoords(undefined);
+    }
+    computePosition(domRange, ref.current, { placement: "bottom-start" })
+      .then((pos) => {
+        setCoords({ x: pos.x, y: pos.y - 10 });
+      })
+      .catch(() => {
+        setCoords(undefined);
+      });
+  }, [isKeyDown]);
+
+  const closeMenu = () => {
+    setBackslashOpen(false);
+  };
+
+  useEffect(() => {
+    const onKeyDown = (event: KeyboardEvent) => {
+      if (event.key === "\\") {
+        event.preventDefault();
+        calculatePosition();
+        setBackslashOpen((prev) => !prev);
+      }
+    };
+    return editor.registerRootListener(
+      (rootElement: HTMLElement | null, prevRootElement: HTMLElement | null) => {
+        if (prevRootElement !== null) {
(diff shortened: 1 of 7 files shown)


packages/platform/src/editor/plugins/toolbar/BlockFormatDropDown.tsx
```diff
@@ -59,10 +59,12 @@ export default function BlockFormatDropDown({
   editor,
   blockType,
   disabled = false,
+  closeMenu,
```
irahopkinson, line 62: I suggest renaming this to `onSelect` which lets users of this component know when it will be called. So in the calling component (BackslashMenu L47) the property would be `onSelect={closeMenu}` which reads well, i.e. when we select in BlockFormatDropDown it will close the BackslashMenu.

packages/platform/src/editor/plugins/toolbar/BlockFormatDropDown.tsx
```diff
@@ -71,6 +73,9 @@ export default function BlockFormatDropDown({
         $setBlocksType(selection, () => $createParaNode(typeToStyle(selectedBlockType)));
       }
     });
+    if (closeMenu) {
+      closeMenu();
+    }
```
irahopkinson, line 78: FYI there is a shortcut syntax you can use here `closeMenu.?();` which says call this method if it is defined. That way it's only 1 line instead of 3. Nothing wrong with using the `if` block as it is but I think this other syntax is a little cleaner.

packages/platform/src/editor/plugins/BackslashPlugin.tsx
irahopkinson, line 1: Take note of the 3 GitHub Action warnings about React dependencies in this file.

packages/platform/src/editor/plugins/BackslashPlugin.tsx
```diff
@@ -0,0 +1,83 @@
...
+import { BackslashMenu, BackslashMenuCoords } from "../components/BackslashMenu/BackslashMenu";
+import { useBackslashMenu } from "../components/BackslashMenu/useBackslashMenu";
+
+const DOM_ELEMENT = document.body;
+
+export default function FloatingMenuPlugin() {
+  const ref = useRef<HTMLDivElement>(null);
+  const [coords, setCoords] = useState<BackslashMenuCoords>(undefined);
+  const [editor] = useLexicalComposerContext();
+  const [backslashOpen, setBackslashOpen] = useState(false);
```
irahopkinson, line 15: nit: for boolean variables it's a good idea to use status words like `didFinish`, `isInitialized`, or `hasPolicy`. So this could be renamed `isBackslashOpen` or in context it could simply be `isMenuOpen`.

FYI I use `nit` to indicate this is nit-picky. The code is OK as is and doesn't necessarily need to be changed.

packages/platform/src/editor/plugins/BackslashPlugin.tsx
```diff
@@ -0,0 +1,83 @@
...
+
+  const { isKeyDown, isKeyReleased } = useBackslashMenu();
+
+  const calculatePosition = useCallback(() => {
+    const domSelection = getSelection();
+    const domRange = domSelection?.rangeCount !== 0 && domSelection?.getRangeAt(0);
+    if (!domRange || !ref.current) {
+      return setCoords(undefined);
+    }
+    computePosition(domRange, ref.current, { placement: "bottom-start" })
```
irahopkinson, line 25: If you hit _Enter/Return_ to get a new blank line and then type '\\' the menu pops up at the top left of the page (i.e. coord 0, 0). Can we improve that?

packages/platform/src/editor/plugins/BackslashPlugin.tsx
```diff
@@ -0,0 +1,83 @@
...
+      .then((pos) => {
+        setCoords({ x: pos.x, y: pos.y - 10 });
```
irahopkinson, line 27: nit: it can be helpful to give this magic number (10) a const name to explain what it is being used for, e.g. `const MENU_HORIZONTAL_INDENT = 10;`. I don't know if this is actaully an indent value but the name would explain that.

packages/platform/src/editor/plugins/BackslashPlugin.tsx
```diff
@@ -0,0 +1,83 @@
...
+      });
+  }, [isKeyDown]);
+
+  const closeMenu = () => {
+    setBackslashOpen(false);
+  };
+
+  useEffect(() => {
+    const onKeyDown = (event: KeyboardEvent) => {
+      if (event.key === "\\") {
```
irahopkinson, line 40: To help improve UX, could ESC and BACKSPACE close the context menu if it's already open?

irahopkinson (changes requested):
Thanks for this good work. Also a few requests here.

irahopkinson:
Closing in favor of PR #199.

PR #24, opened 2024-02-07
cut/copy/paste context menu

### LINKED ISSUE #13: Cut/Copy/Paste
Add keyboard shortcuts and context menu to do cut/copy/paste.

Checkout the context menu in the [lexical playground](https://playground.lexical.dev/). It's likely best to adapt that.

**Note:** keyboard shortcuts should be cross-platform, i.e work in Windows/Linux/macOS. You will likely want to use [lexical/environment.ts](../blob/main/packages/shared/lexical/environment.ts) to help with this.

samueljd:
Fixes #13 
- Added context menu for cut/copy/paste.
- I've only added this plugin to `perf-react` package since it breaks the platform currently.


packages/perf-react/src/lexical/plugins/ContextMenuPlugin.tsx
```diff
@@ -0,0 +1,189 @@
...
+            const event = new ClipboardEvent("paste", {
+              clipboardData: data,
+            });
+
+            editor.dispatchCommand(PASTE_COMMAND, event);
+          });
+        },
+      }),
+    ];
+  }, [editor]);
```
irahopkinson, line 139: Is there some reason you didn't include the **Paste as Plain Text** option? If it works properly then please add it in. If it caused some issue then leave it out.

packages/perf-react/src/lexical/plugins/ContextMenuPlugin.tsx
```diff
@@ -0,0 +1,189 @@
...
+    this.onSelect = options.onSelect.bind(this);
+  }
+}
+
+export default function ContextMenuPlugin(): JSX.Element {
+  const [editor] = useLexicalComposerContext();
+
+  const options = useMemo(() => {
+    return [
+      new ContextMenuOption(`Copy`, {
```
irahopkinson, line 98: The usual order is Cut-Copy-Paste. Can we swap these first two to match that order?

packages/perf-react/src/lexical/plugins/ContextMenuPlugin.tsx
```diff
@@ -0,0 +1,189 @@
...
+        onSelect: (_node) => {
+          editor.dispatchCommand(COPY_COMMAND, null);
+        },
+      }),
+      new ContextMenuOption(`Cut`, {
+        onSelect: (_node) => {
+          editor.dispatchCommand(CUT_COMMAND, null);
+        },
+      }),
+      new ContextMenuOption(`Paste`, {
```
irahopkinson, line 108: Curiously in [Codesandbox](https://codesandbox.io/p/github/BiblioNexus-Foundation/scripture-editors/context-menu) (which I think is Linux) the context menu _Paste_ item doesn't paste (_Ctrl + V_ works fine though). Running locally on my Win11 the context paste works just fine. Can you investigate this?

packages/perf-react/src/index.css
irahopkinson, line 1: If you first scroll down to the first verse and then test there, it doesn't appear to display the context menu. It turns out that it is displaying the menu further down the page so if you scroll up (I was using the mouse scroll wheel so that didn't close the menu by clicking) you can see the menu further up the screen. This may just be a CSS issue, can you investigate?

irahopkinson (changes requested):
Thanks for your good work on this. I have a couple of requests.

packages/perf-react/src/index.css
irahopkinson: Done.

packages/perf-react/src/lexical/plugins/ContextMenuPlugin.tsx
irahopkinson: Done.

packages/perf-react/src/lexical/plugins/ContextMenuPlugin.tsx
irahopkinson: This appears to be working now.

irahopkinson (approved):

PR #48, opened 2024-03-07
uncontrolled USJ props

### LINKED ISSUE #44: Uncontrolled Scripture data property
The nature of an editor means it's meant to be an uncontrolled React component since edits don't need to fire a re-render since **Lexical** has its own editor state and takes care of rendering the edits. See end of this page in [Lexical Docs](https://lexical.dev/docs/getting-started/react):

> "_Lexical is generally meant to be uncontrolled..._"

A React component can have controlled and uncontrolled properties. See [React Docs](https://react.dev/learn/referencing-values-with-refs#example-building-a-stopwatch):

> "_You can combine refs and state in a single component._"

When loading the whole book of Psalms, the render takes ~1.8s. Typing a single character and then waiting for 1.8s before you can type the next character is not practical (which is likely as a controlled React component). What is the best practice when using **Lexical**?

This is a **BREAKING CHANGE**.

irahopkinson:
- defer USJ updates to platform editor.
- fix round trip.
  - removed properties that are undefined.
  - remove empty content arrays.
- also move `lexical` to dependencies (no longer - bundled in editor package) .
- also clear timeout on `UpdateStatePlugin` exit.
- fixes #44.

### DIFF
--- a/packages/platform/src/App.tsx
+++ b/packages/platform/src/App.tsx
@@ -1,5 +1,5 @@
 import { RefSelector, ScriptureReference } from "papi-components";
-import { useMemo, useState } from "react";
+import { useCallback, useMemo, useRef, useState } from "react";
 import { usxStringToUsj } from "shared/converters/usj/usx-to-usj";
 import { Usj } from "shared/converters/usj/usj.model";
 import { WEB_PSA_USX as usx } from "shared/data/WEB-PSA.usx";
@@ -9,21 +9,27 @@ import { UsjNodeOptions } from "shared-react/nodes/scripture/usj/usj-node-option
 import { getViewOptions } from "./editor/adaptors/view-options.utils";
 import { formattedViewMode as defaultViewMode } from "./editor/plugins/toolbar/view-mode.model";
 import ViewModeDropDown from "./editor/plugins/toolbar/ViewModeDropDown";
-import Editor from "./editor/Editor";
+import Editor, { EditorRef } from "./editor/Editor";
 import "./App.css";
 
+const defaultUsj = usxStringToUsj('<usx version="3.0" />');
 const defaultScrRef: ScriptureReference = { /* PSA */ bookNum: 19, chapterNum: 1, verseNum: 1 };
-
-const usj = usxStringToUsj(usx);
-
 const nodeOptions: UsjNodeOptions = { [immutableNoteCallerNodeName]: { onClick: () => undefined } };
 
-const onChange = (usj: Usj) => console.log({ usj });
-
 export default function App() {
+  const editorRef = useRef<EditorRef>(null);
   const [viewMode, setViewMode] = useState(defaultViewMode);
   const [scrRef, setScrRef] = useState(defaultScrRef);
   const viewOptions = useMemo(() => getViewOptions(viewMode), [viewMode]);
+
+  setTimeout(() => {
+    editorRef.current?.setUsj(usxStringToUsj(usx));
+  }, 1000);
+
+  const onChange = useCallback((usj: Usj) => {
+    console.log({ usj });
+    editorRef.current?.setUsj(usj);
+  }, []);
 
   return (
     <>
@@ -32,7 +38,8 @@ export default function App() {
       </div>
       <ViewModeDropDown viewMode={viewMode} handleSelect={setViewMode} />
       <Editor
-        usj={usj}
+        defaultUsj={defaultUsj}
+        ref={editorRef}
         viewOptions={viewOptions}
         scrRef={scrRef}
         setScrRef={setScrRef}
--- a/packages/platform/src/App.tsx
+++ b/packages/platform/src/App.tsx
@@ -17,7 +17,9 @@ const defaultScrRef: ScriptureReference = { /* PSA */ bookNum: 19, chapterNum: 1
 const nodeOptions: UsjNodeOptions = { [immutableNoteCallerNodeName]: { onClick: () => undefined } };
 
 export default function App() {
-  const editorRef = useRef<EditorRef>(null);
+  // This ref becomes defined when passed to the editor.
+  // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
+  const editorRef = useRef<EditorRef>(null!);
   const [viewMode, setViewMode] = useState(defaultViewMode);
   const [scrRef, setScrRef] = useState(defaultScrRef);
   const viewOptions = useMemo(() => getViewOptions(viewMode), [viewMode]);
--- a/packages/platform/src/editor/Editor.tsx
+++ b/packages/platform/src/editor/Editor.tsx
@@ -1,12 +1,13 @@
-import { ScriptureReference } from "papi-components";
-import { JSX, useCallback } from "react";
+import { EditorRefPlugin } from "@lexical/react/LexicalEditorRefPlugin";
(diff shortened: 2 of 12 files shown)


packages/platform/src/App.tsx
```diff
@@ -9,30 +9,37 @@ import { UsjNodeOptions } from "shared-react/nodes/scripture/usj/usj-node-option
...
   const [scrRef, setScrRef] = useState(defaultScrRef);
   const viewOptions = useMemo(() => getViewOptions(viewMode), [viewMode]);
 
+  setTimeout(() => {
+    editorRef.current?.setUsj(usxStringToUsj(usx));
+  }, 1000);
+
+  const onChange = useCallback((usj: Usj) => {
+    console.log({ usj });
+    editorRef.current?.setUsj(usj);
```
tjcouch-sil, line 31: I could be misunderstanding, but I thought the point of making it uncontrolled was so you didn't need to set the usj when you received notice of change from the editor. Shouldn't we remove this line?

packages/platform/src/editor/Editor.tsx
```diff
@@ -66,7 +71,10 @@ function Placeholder(): JSX.Element {
  * Scripture Editor for USJ. Created for use in Platform.Bible.
  * @see https://github.com/usfm-bible/tcdocs/blob/usj/grammar/usj.js
  *
- * @param props.usj - USJ Scripture data.
+ * @param props.defaultUsj - Default USJ Scripture data.
```
tjcouch-sil, line 74: Does putting all these `param` comments in jsdocs on the object types themselves carry the comments around everywhere? Seems it might be nice to have the comments on the types so the comments are with the properties. Of course, feel free to ignore this. Just a thought for your consideration.

packages/platform/src/editor/Editor.tsx
```diff
@@ -80,24 +88,46 @@ function Placeholder(): JSX.Element {
...
+    nodeOptions,
+    isReadonly,
+    onChange,
+    logger,
+  }: EditorProps<TLogger>,
+  ref: React.ForwardedRef<EditorRef>,
+): JSX.Element {
+  const editorRef = useRef<LexicalEditor>(null);
+  const [usj, setUsj] = useState(defaultUsj);
+  const [loadedUsj, , setEditedUsj] = useDeferredState(usj);
```
tjcouch-sil, line 106: Do we still need this if we stop setting the USJ up above in the `onChange`? I appreciate that deep equaling the state is nice to avoid having to rerender the editor unnecessarily. However, this pattern seems to deviate somewhat significantly from normal React patterns and thereby increase mental complexity of this component. It also seems to add a render and a duplicate of this state (probably neither of which are huge deals; just considering the consequences). If it is still very important to do a deep equals check even after making the component uncontrolled (meaning we can likely assume people won't be setting this state as a result of the `onChange` running unless they're doing it wrong), could we do that somewhere in a `memo` function? Maybe in `UpdateStatePlugin`? Maybe there are too many other props in there; dunno. Just a thought.

packages/platform/src/editor/Editor.tsx
```diff
@@ -80,24 +88,46 @@ function Placeholder(): JSX.Element {
...
-  scrRef,
-  setScrRef,
-  nodeOptions,
-  isReadonly,
-  onChange,
-  logger,
-}: EditorProps<TLogger>): JSX.Element {
+const Editor = forwardRef(function Editor<TLogger extends LoggerBasic>(
+  {
+    defaultUsj,
```
tjcouch-sil, line 93: I might suggest wrapping this in a `memo` function and ignoring changes to `defaultUsj` to avoid extra renders, but it also isn't a big deal; feel free to ignore or make a note somewhere.

tjcouch-sil (changes requested):
Awesome :) looking forward to having this in! Thanks for the hard work.

packages/platform/src/App.tsx
```diff
@@ -9,30 +9,37 @@ import { UsjNodeOptions } from "shared-react/nodes/scripture/usj/usj-node-option
...
 
-const onChange = (usj: Usj) => console.log({ usj });
-
 export default function App() {
+  const editorRef = useRef<EditorRef>(null);
   const [viewMode, setViewMode] = useState(defaultViewMode);
   const [scrRef, setScrRef] = useState(defaultScrRef);
   const viewOptions = useMemo(() => getViewOptions(viewMode), [viewMode]);
 
+  setTimeout(() => {
```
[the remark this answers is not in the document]
irahopkinson: This is just to simulate USJ being set sometime after mount. Note that `Editor.tsx` is bundled to the NPM package and `App.tsx` is not but is used to test the bundled editor when running in this repo.

packages/platform/src/App.tsx
```diff
@@ -9,30 +9,37 @@ import { UsjNodeOptions } from "shared-react/nodes/scripture/usj/usj-node-option
...
 const defaultScrRef: ScriptureReference = { /* PSA */ bookNum: 19, chapterNum: 1, verseNum: 1 };
-
-const usj = usxStringToUsj(usx);
-
 const nodeOptions: UsjNodeOptions = { [immutableNoteCallerNodeName]: { onClick: () => undefined } };
 
-const onChange = (usj: Usj) => console.log({ usj });
-
 export default function App() {
+  const editorRef = useRef<EditorRef>(null);
```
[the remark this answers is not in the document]
irahopkinson: Done.

packages/platform/src/editor/Editor.tsx
irahopkinson: Since `Editor.tsx` is the main component that is exported in the npm package, I wanted the usage information to come up for the developer when they hover over the component. That doesn't happen when the JSDoc is only on the types.

packages/platform/src/editor/Editor.tsx
irahopkinson: Good suggestion. Like this?

packages/platform/src/editor/Editor.tsx
irahopkinson: No, we don't need it if we use the uncontrolled component properly. This still needs some work however when you update USX in P.B it currently spits out a new USX value so we need this logic either in the editor component or in every extension that uses it (or in the USX PDP). I elected to put it in the editor for now but expect to remove it later.

packages/platform/src/editor/Editor.tsx
tjcouch-sil: Oh... I wonder if we should fix this in Platform.Bible too then 😬

packages/platform/src/editor/Editor.tsx
tjcouch-sil: Sorry, I was talking about using the [`memo`](https://react.dev/reference/react/memo) function. You could make an `arePropsEqual` function that checks all the props and ignores the `defaultUsj` prop. But it really doesn't matter much and might just be a pre-optimization. Doing it like you had it before is fine.

tjcouch-sil (changes requested):
I'd recommend switching the `defaultUsj` back (or using the `memo` function as I mentioned in the follow-up comment, but it is probably overkill right now) as it now causes unnecessary extra cycles. `useState` already uses the value at first and ignores it from then on. Sorry for the miscommunication.

PR #91, opened 2024-05-23
Fix eslint with prettier

irahopkinson:
This is not currently causing a problem but if other eslint plugins are added it is a problem waiting to happen.


samueljd (approved):

samueljd (commented):
Looks good. Thanks.

PR #166, opened 2024-09-03
135 onchange fired when loading new text

abelpz:
fixes https://github.com/BiblioNexus-Foundation/scripture-editors/issues/135 by adding tags to editor updates, and ignoring those updates that are not supposed to trigger the onChange callback.


irahopkinson (approved):
👍 LGTM, thanks for this.

PR #190, opened 2024-10-23
Add a separate plugin with a filter input for the nodes selection menu

abelpz:
This PR incorporates code from the **Typeahead node selection menu plugin** into a new **Node selection menu plugin**. The new menu activates when the user presses the trigger key, ensuring that no text is entered into the editor during its active state.

The menu includes an input field that displays user input and dynamically updates the eligible node markers list as the query changes, mimicking Paratext's behavior. The user can navigate through the nodes list using the up and down arrow keys, and can select a node by clicking on it or pressing the enter or tab keys as the desired node is highlighted.

The menu can be hidden by pressing the escape key, backspace when the input is empty, or moving the cursor to a different position in the editor.


packages/shared/styles/perf-editor.css
irahopkinson, line 1: Run `pnpm prettier --write .` at the repo root to fix this file's formatting (and the CI build errors).

irahopkinson (changes requested):
Sorry, only partial review. I started reviewing this and then realized it is still a draft PR.

packages/shared-react/plugins/NodesMenu/NodeNavigation.tsx
```diff
@@ -0,0 +1,42 @@
+import { COMMAND_PRIORITY_HIGH } from "lexical";
+import { KEY_DOWN_COMMAND } from "lexical";
```
irahopkinson, line 2: NIT these imports can be combined.

packages/shared-react/plugins/NodesMenu/NodeNavigation.tsx
```diff
@@ -0,0 +1,42 @@
...
+    };
+
+    const unregisterCommand = editor.registerCommand(
+      KEY_DOWN_COMMAND,
+      handleEvent,
+      COMMAND_PRIORITY_HIGH,
+    );
+    return () => {
+      unregisterCommand();
+    };
```
irahopkinson, line 38: NIT pretty sure this can be simplified to `return unregisterCommand;`

packages/shared-react/plugins/NodesMenu/NodeSelectionMenu.tsx
```diff
@@ -0,0 +1,98 @@
...
+  const [editor] = useLexicalComposerContext();
+  const isControlled = controlledQuery !== undefined;
+  const [query, setQuery] = useState("");
+  const localQuery = isControlled ? controlledQuery : query;
+
+  const filteredOptions = useFilteredItems({ query: localQuery, items: options, filterBy: "name" });
+
+  const handleOptionSelection = (option: NodeOption) => {
+    onSelectOption ? onSelectOption(option) : option.action(editor);
+    onClose && onClose();
```
irahopkinson, line 39: NIT this line can be simplified to `onClose?.();`

packages/shared-react/plugins/NodesMenu/NodeSelectionMenu.tsx
```diff
@@ -0,0 +1,98 @@
...
+            return true;
+          }
+        }
+        return false;
+      },
+      COMMAND_PRIORITY_HIGH,
+    );
+    return () => {
+      unregisterCommand();
+    };
```
irahopkinson, line 77: Can be simplified to `return unregisterCommand`.

packages/shared-react/plugins/NodesMenu/index.tsx
```diff
@@ -0,0 +1,75 @@
...
+      document.addEventListener("keydown", hideMenu);
+      return () => {
+        document.removeEventListener("keydown", hideMenu);
+      };
+    }
+  }, [isOpen, hideMenu]);
+
+  // Open the menu when the trigger key is pressed
+  useEffect(() => {
+    const showMenu = (e: KeyboardEvent) => {
```
irahopkinson, line 33: BTW I'm curious why this `showMenu` function is defined inside the `useEffect` whereas `hideMenu` is defined outside its `useEffect`?

packages/shared-react/plugins/NodesMenu/index.tsx
```diff
@@ -0,0 +1,75 @@
...
+    const unregisterRootListener = editor.registerRootListener((root) => {
+      if (!root) return;
+      root.addEventListener("keydown", showMenu);
+      return () => {
+        root.removeEventListener("keydown", showMenu);
+      };
+    });
+    return () => {
+      unregisterRootListener();
+    };
```
irahopkinson, line 48: This can be simplified to `return unregisterRootListener;`

irahopkinson (changes requested):
Thanks for all this good work.

- There is a bug if you select text and then type `\pm` + `enter`. It recurring fails and locks up the editor. I'm not sure if this is meant to work but at least it shouldn't lock up the editor.
- if you go to the end of a paragraph and type `\p` + `enter` you get 2 paragraphs instead of 1.
- if you create a new paragraph and then add a note (`\f` + `enter`) you can't delete the note. (This is likely an unrelated problem and out of scope for this PR.)

packages/shared-react/plugins/NodesMenu/Menu/Option.tsx
```diff
@@ -0,0 +1,56 @@
...
+        onSelect?.(index);
+        setSelectedIndex(-1);
+      }
+    }, [index, selectedIndex, onSelect]);
+
+    const handleClick = useCallback(() => {
+      if (index !== undefined) {
+        setSelectedIndex(index);
+      }
+    }, [index, setSelectedIndex, onClick]);
```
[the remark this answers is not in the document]
abelpz: Thanks! It was a leftover from previous implementation

packages/shared-react/plugins/NodesMenu/index.tsx
abelpz: Probably because showMenu is only used inside this useEffect whereas hideMenu is used in two separate useEffect calls.

packages/shared-react/plugins/useLexicalMenuNavigation.ts
```diff
@@ -0,0 +1,101 @@
...
+  COMMAND_PRIORITY_LOW,
+  KEY_ARROW_DOWN_COMMAND,
+  KEY_ARROW_UP_COMMAND,
+  KEY_ENTER_COMMAND,
+  KEY_TAB_COMMAND,
+  LexicalEditor,
+} from "lexical";
+import { useEffect, useState } from "react";
+
+export default function useLexicalMenuNavigation<
```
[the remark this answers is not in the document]
abelpz: Thanks, just leftovers from previous implementation

irahopkinson (dismissed):
LGTM, just one NIT comment.

Also, if you add a new paragraph (CR at end of paragraph with no content) and then press `\` the menu is at top-left (probably because it has no content to anchor to).

irahopkinson:
There was another issue but I can't find consistently reproduceable steps. Something like select text and try `\pm + enter`, then `\f + enter`, then create a new paragraph and try `\pm + enter`, some times the new `pm` paragraph would contain an empty note.

abelpz:
> There was another issue but I can't find consistently reproduceable steps. Something like select text and try `\pm + enter`, then `\f + enter`, then create a new paragraph and try `\pm + enter`, some times the new `pm` paragraph would contain an empty note.

This should be fixed after the latest commit

PR #199, opened 2024-11-17
Reuse contextual nodes selection menu in platform

abelpz:


packages/platform/src/editor/Editor.tsx
```diff
@@ -262,10 +272,20 @@ const Editor = forwardRef(function Editor<TLogger extends LoggerBasic>(
...
             />
           )}
+          <UsjNodesMenuPlugin
+            trigger="\"
+            scriptureReference={{
+              book: Canon.bookNumberToId(scrRef?.bookNum ?? 0),
+              chapter: scrRef?.chapterNum ?? 0,
+              verse: scrRef?.verseNum ?? 0,
+            }}
+            contextMarker="p"
```
abelpz, line 286: @irahopkinson I will need help setting this variable, this is supposed to be the marker selected, so it is a string with the current usfm marker of the node where the cursor is located.

abelpz:
@irahopkinson, this is still a draft. However, if you have some time, I would appreciate it if you could take a look at some of the code that the platform relies on, which I had to move to shared. And other such things.

packages/platform/src/editor/Editor.tsx
irahopkinson: Yep, lets talk on your Monday.

irahopkinson (dismissed):
Thanks for all you good work on this. I think there are some concepts I'm missing. Can we talk on your Monday?

pnpm-lock.yaml
irahopkinson, line 1: When I did a `pnpm i` on this branch it further modified this file so you likely need to do that also.

packages/platform/src/App.tsx
```diff
@@ -16,6 +16,7 @@ import Marginal, { MarginalRef } from "./marginal/Marginal";
 import TextDirectionDropDown from "./TextDirectionDropDown";
 import ViewModeDropDown from "./ViewModeDropDown";
 import "./App.css";
+import "shared/styles/nodes-menu.css";
```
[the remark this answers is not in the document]
abelpz: I can do that, just didn't mix it with the plugin, so you can use your own CSS if you wanted.

abelpz (commented):
Looking pretty good, and working great on perf-react too. Good work @irahopkinson. I cannot approve since this is my PR, but you can.

PR #206, opened 2024-11-26
fix `getMarkerAction` for mark nodes

irahopkinson:
- use the right `$is...` functions


irahopkinson:
@abelpz when working on #205 I noticed `getMarkerAction` was using `MarkerNode` (I moved this node so I noticed) when I'm pretty sure from your comments it should be `MarkNode` instead. This PR fixes that. Please review and check if this is right.

abelpz (approved):
LGTM

PR #340, opened 2025-07-16
update: major package updates and compatibility fixes

samueljd:
## Major Package Updates
- TypeScript: ~5.5.4 -> ~5.8.3
- @types/react: Updated to v19.x
- Vite: ^5.4.14 -> ^7.0.4
- TailwindCSS: v3 -> v4.1.11
- ESLint: ^9.25.1 -> ^9.31.0
- Prettier: ^3.5.3 -> ^3.6.2
- Nx: 19.8.14 -> 21.2.3
- Lexical: ^0.33.0
- vite-plugin-dts: 4.5.3 -> ^4.5.4

## TypeScript & React Compatibility Fixes
- Replace JSX.Element with ReactElement imports from 'react'
- Fix useRef() calls to include initial values (required by @types/react v19)
- Update RefObject<T> to RefObject<T | null> for stricter null checking
- Add proper type imports across platform, shared-react packages

## Vite 7.x Configuration Updates
- Update vite.config.ts in perf-react and perf-vanilla packages
- Add explicit rollupOptions with input resolution for index.html
- Fix "Could not resolve entry module" build errors
## Scribe Tailwind update
- Add @tailwindcss/postcss v4.1.11 as devDependency for scribe
- Resolve PostCSS/TailwindCSS integration issues
- Update CSS imports from @tailwind directives to @import "tailwindcss"
- Convert theme configuration from JavaScript to CSS-based @theme directive
- Migrate custom colors and height variables to CSS custom properties
- Update PostCSS configuration to use @tailwindcss/postcss plugin
- Remove autoprefixer dependency (handled by TailwindCSS v4's Lightning CSS)
- Remove old tailwind.config.js file (replaced by CSS configuration)


packages/shared/package.json
```diff
@@ -15,17 +15,17 @@
...
-    "lexical": "^0.33.0",
+    "lexical": "^0.33.1",
     "open-patcher": "1.0.0-beta.1"
   },
   "devDependencies": {
     "json-difference": "^1.16.1",
-    "proskomma-core": "^0.10.10",
-    "proskomma-json-tools": "^0.8.21"
+    "proskomma-core": "^0.11.0",
+    "proskomma-json-tools": "^0.9.0"
```
irahopkinson, line 28: I have not been upgrading these because I don't know that Abel wants that. If you have checked with him then by all means go ahead.

packages/shared-react/package.json
```diff
@@ -18,17 +18,17 @@
...
   },
   "devDependencies": {
-    "@types/react": "^18.3.18",
-    "@types/react-dom": "^18.3.5",
-    "react": "^18.3.1",
-    "react-dom": "^18.3.1",
+    "@types/react": "^19.1.8",
+    "@types/react-dom": "^19.1.6",
+    "react": "^19.1.0",
+    "react-dom": "^19.1.0",
```
irahopkinson, line 31: We don't want to move up to React v19 yet. Please change this back to v18.

irahopkinson (changes requested):
The GHA build is failing because of an issue with the PNPM lock file. Given the change to the lock file when I did `pnpm i`, if you commit that it will likely get past the current error. Also doing that gave me this error: 
```sh
.
└─┬ @nx/vite 21.2.3
  └── ✕ unmet peer vite@"^5.0.0 || ^6.0.0": found 7.0.4
```
You'll need to move vite back to the latest v6 (rather than use v7) to make it compatible with NX.

Also running `nx run-many -t typecheck test lint build` gave me only one test failure so looking good, but we'll need to know why that is failing.

I also noticed that when you type '\\' for the marker menu the '\\' now appears as the first character in the menu whereas before it would swallow that character and only show the characters you type after.

When I open say `packages/platform/src/editor/Editor.tsx` in this branch I get a few errors in VS Code. This is likely due to the `ReactElement` changes that need a bit more work.

I haven't reviewed everything yet but I thought I would at least give you this so far.

irahopkinson (approved):

PR #468, opened 2026-04-13
update dependencies

irahopkinson:
- update Lexical v0.43.0
- update NX v22.6.5
- also update Claude git ignore
- also fix MCP config on Windows
- also clean-up vitest migration instructions

---

<a href="https://app.devin.ai/review/eten-tech-foundation/scripture-editors/pull/468" target="_blank">
  <picture>
    <source media="(prefers-color-scheme: dark)" srcset="https://static.devin.ai/assets/gh-open-in-devin-review-dark.svg?v=1">
    Open with Devin
  </picture>
</a>

- - -
This change is [Reviewable](https://reviewable.io/reviews/eten-tech-foundation/scripture-editors/468)


katherinejensen00 (commented):
Ira, thanks for upgrading the versions and improving the context menu, the Windows MCP config, and the clean-up vitest migration instructions! 
As I was asking Devin AI questions about the code to help me understand it, Devin mentioned some potential improvements that I included. They are AI comments so feel free to take or leave them. I made one blocking so this PR doesn't merge before you have a chance to look at them.

> @katherinejensen00 reviewed 19 files and all commit messages, and made 7 comments.
> *[Reviewable](https://reviewable.io/reviews/eten-tech-foundation/scripture-editors/468#-:-Oq6TWvg2bpip6fp1VXq:b1rmorm)* status: all files reviewed, 6 unresolved discussions (waiting on irahopkinson).

___
*[`libs/shared-react/src/plugins/usj/ContextMenuPlugin.tsx` line 7 at r1](https://reviewable.io/reviews/eten-tech-foundation/scripture-editors/468#-Oq6S6efAFblQ9pNqsF9:-Oq6S6efAFblQ9pNqsFA:bdmb4it) ([raw file](https://github.com/eten-tech-foundation/scripture-editors/blob/51b6a50c020a7cc9e0161b42a2d1a6dfb10dc968/libs/shared-react/src/plugins/usj/ContextMenuPlugin.tsx#L7)):*
> ```XML
> import { pasteSelection, pasteSelectionAsPlainText } from "./clipboard.utils";
> import { useLexicalComposerContext } from "@lexical/react/LexicalComposerContext";
> import { LexicalContextMenuPlugin, MenuOption } from "@lexical/react/LexicalContextMenuPlugin";
> ```

Devin AI Comment:

### 3\. No keyboard navigation

The old `LexicalContextMenuPlugin` (via `MenuOption`) provided arrow-key navigation through menu items. The new implementation only supports mouse hover to highlight and Escape to close. There's no up/down arrow key support. This is an accessibility regression — `role="option"` and `aria-selected` are present but keyboard users can't actually navigate the options.
___
*[`libs/shared-react/src/plugins/usj/ContextMenuPlugin.tsx` line 48 at r1](https://reviewable.io/reviews/eten-tech-foundation/scripture-editors/468#-Oq6SJo9AO7AEbaQmV9A:-Oq6SJo9AO7AEbaQmV9B:bmddqfx) ([raw file](https://github.com/eten-tech-foundation/scripture-editors/blob/4502898e0fc2e80c4a856b7c0d392736e4385ad3/libs/shared-react/src/plugins/usj/ContextMenuPlugin.tsx#L48)):*
> ```XML
>   return (
>     <li
>       key={option.title}
> ```

Devin AI Comment:

### 4. `key={option.title}` could collide

At `libs/shared-react/src/plugins/usj/ContextMenuPlugin.tsx:48` and `84`, `option.title` is used as the React `key`. If a consumer passes two extra options with the same title, React will warn and behave incorrectly. The old code used `option.key` from `MenuOption` which was guaranteed unique. Consider adding a unique identifier to `ContextMenuOption` or using the array index (which is already available).
___
*[`libs/shared-react/src/plugins/usj/ContextMenuPlugin.tsx` line 184 at r1](https://reviewable.io/reviews/eten-tech-foundation/scripture-editors/468#-Oq6R_yJ3TvdKoX4ErKi:-Oq6R_yK2HKh8qQud3Mi:bw5l5ny) ([raw file](https://github.com/eten-tech-foundation/scripture-editors/blob/4502898e0fc2e80c4a856b7c0d392736e4385ad3/libs/shared-react/src/plugins/usj/ContextMenuPlugin.tsx#L184)):*
> ```XML
>   // Register context menu event on editor root
>   useEffect(() => {
>     return editor.registerRootListener((rootElement) => {
> ```

Devin AI Comment:

### 1\. Event listener cleanup on root element swap (potential bug)

At ContextMenuPlugin.tsx:183-203, the `registerRootListener` callback receives the new root element but is also called with `null` when the previous root is removed. The current code returns early on `null` (`line 185`), which means it never cleans up the listener on the old root element. The cleanup function returned on `line 201` only runs when `registerRootListener` calls the callback again for the _next_ root, but the Lexical `registerRootListener` API actually passes `(rootElement, prevRootElement)` — you should also remove the listener from `prevRootElement`. As written, if the root element changes (e.g., during editor re-mounting), the old listener would leak.

```
// Suggested fix:return editor.registerRootListener((rootElement, prevRootElement) => {  if (prevRootElement) {    prevRootElement.removeEventListener("contextmenu", handleContextMenu);  }  if (!rootElement) return;  rootElement.addEventListener("contextmenu", handleContextMenu);});
```

Note this requires hoisting `handleContextMenu` out of the callback, or using a stable ref.
___
*[`libs/shared-react/src/plugins/usj/ContextMenuPlugin.tsx` line 210 at r1](https://reviewable.io/reviews/eten-tech-foundation/scripture-editors/468#-Oq6Ry2v6Z3hUmLcG40y:-Oq6Ry2v6Z3hUmLcG40z:b99mpar) ([raw file](https://github.com/eten-tech-foundation/scripture-editors/blob/4502898e0fc2e80c4a856b7c0d392736e4385ad3/libs/shared-react/src/plugins/usj/ContextMenuPlugin.tsx#L210)):*
> ```XML
>       closeMenu();
>     };
>     globalThis.addEventListener("scroll", handleScroll, true);
> ```

Devin AI Comment:

### 2\. Scroll listener is always active

At `libs/shared-react/src/plugins/usj/ContextMenuPlugin.tsx:206-212`, the scroll listener is registered globally at all times (capturing phase), even when the menu is closed. The click-outside and Escape handlers correctly guard with `if (!menuState.isOpen) return;`, but the scroll handler doesn't. This is minor but inconsistent — consider guarding it the same way for consistency and to avoid unnecessary `closeMenu()` calls (which are no-ops but still trigger state setter calls).
___
*[`libs/shared-react/src/plugins/usj/ContextMenuPlugin.tsx` line 244 at r1](https://reviewable.io/reviews/eten-tech-foundation/scripture-editors/468#-Oq6SYobEtFTPqHfA16M:-Oq6SYobEtFTPqHfA16N:b-gxsvct) ([raw file](https://github.com/eten-tech-foundation/scripture-editors/blob/4502898e0fc2e80c4a856b7c0d392736e4385ad3/libs/shared-react/src/plugins/usj/ContextMenuPlugin.tsx#L244)):*
> ```XML
>   if (!menuState.isOpen) return null;
> 
>   return ReactDOM.createPortal(
> ```

Devin AI Comment:

### 5\. Menu can render off-screen

At `libs/shared-react/src/plugins/usj/ContextMenuPlugin.tsx:247-253`, the menu is positioned at the exact click coordinates with `position: fixed`. If the user right-clicks near the bottom or right edge of the viewport, the menu will overflow off-screen. The old implementation relied on Lexical's anchor positioning which may have handled this. Consider adding viewport boundary clamping.
___
*[`scripts/mcp-launcher.js` line 9 at r1](https://reviewable.io/reviews/eten-tech-foundation/scripture-editors/468#-Oq6TCM09v_3qbL1rcLi:-Oq6TCM09v_3qbL1rcLj:by4g4j) ([raw file](https://github.com/eten-tech-foundation/scripture-editors/blob/4502898e0fc2e80c4a856b7c0d392736e4385ad3/scripts/mcp-launcher.js#L9)):*
> ```JavaScript
> 
> const args = process.argv.slice(2);
> const child = spawn("npx", args, { stdio: "inherit", shell: true });
> ```

NIT Devin AI Comment:

No error handling on spawn failure. If `npx` isn't found at all, the `child.on("error", ...)` event isn't handled. This would result in an unhandled error. Adding a `child.on("error", (err) => { ... })` handler would be more robust.

katherinejensen00 (commented):
Looks great! Thanks for looking into those changes
:lgtm:

> @katherinejensen00 reviewed 1 file and all commit messages, made 1 comment, and resolved 4 discussions.
> *[Reviewable](https://reviewable.io/reviews/eten-tech-foundation/scripture-editors/468#-:-Oq85Wq7AW2mDPZSrQ2X:b-49s7pc)* status: all files reviewed, 4 unresolved discussions (waiting on irahopkinson).

katherinejensen00 (commented):
> @katherinejensen00 resolved 2 discussions.
> *[Reviewable](https://reviewable.io/reviews/eten-tech-foundation/scripture-editors/468)* status: all files reviewed, 2 unresolved discussions (waiting on irahopkinson).

irahopkinson (commented):
I wasn't trying to improve the context menu. The next version of Lexcial removed the deprecated `@lexical/react/LexicalContextMenuPlugin` that it was using so we had to re-write the context menu plugin without it.

Thanks for the review items, all worth considering. I fixed all but one as you'll see.

> @irahopkinson made 7 comments.
> *[Reviewable](https://reviewable.io/reviews/eten-tech-foundation/scripture-editors/468#-:-Oq7jae81WcwTrVZ91Uy:br3qyx8)* status: 18 of 19 files reviewed, all discussions resolved (waiting on katherinejensen00).

___
*[`libs/shared-react/src/plugins/usj/ContextMenuPlugin.tsx` line 7 at r1](https://reviewable.io/reviews/eten-tech-foundation/scripture-editors/468#-Oq6S6efAFblQ9pNqsF9:-Oq81iS76crhqiiPXpad:b-896fix) ([raw file](https://github.com/eten-tech-foundation/scripture-editors/blob/51b6a50c020a7cc9e0161b42a2d1a6dfb10dc968/libs/shared-react/src/plugins/usj/ContextMenuPlugin.tsx#L7)):*

Devin AI Comment:

### 3\. No keyboard navigation

The old `LexicalContextMenuPlugin` (via `MenuOption`) provided arrow-key navigation through menu items. The new implementation only supports mouse hover to highlight and Escape to close. There's no up/down arrow key support. This is an accessibility regression — `role="option"` and `aria-selected` are present but keyboard users can't actually navigate the options.
</blockquote></details>

Done.
___
*[`libs/shared-react/src/plugins/usj/ContextMenuPlugin.tsx` line 48 at r1](https://reviewable.io/reviews/eten-tech-foundation/scripture-editors/468#-Oq6SJo9AO7AEbaQmV9A:-Oq81kfFA5i-d6oXkPYT:b-896fix) ([raw file](https://github.com/eten-tech-foundation/scripture-editors/blob/4502898e0fc2e80c4a856b7c0d392736e4385ad3/libs/shared-react/src/plugins/usj/ContextMenuPlugin.tsx#L48)):*

Devin AI Comment:

### 4. `key={option.title}` could collide

At `libs/shared-react/src/plugins/usj/ContextMenuPlugin.tsx:48` and `84`, `option.title` is used as the React `key`. If a consumer passes two extra options with the same title, React will warn and behave incorrectly. The old code used `option.key` from `MenuOption` which was guaranteed unique. Consider adding a unique identifier to `ContextMenuOption` or using the array index (which is already available).
</blockquote></details>

Done.
___
*[`libs/shared-react/src/plugins/usj/ContextMenuPlugin.tsx` line 184 at r1](https://reviewable.io/reviews/eten-tech-foundation/scripture-editors/468#-Oq6R_yJ3TvdKoX4ErKi:-Oq81dPw-mlNDYob23FI:b-896fix) ([raw file](https://github.com/eten-tech-foundation/scripture-editors/blob/4502898e0fc2e80c4a856b7c0d392736e4385ad3/libs/shared-react/src/plugins/usj/ContextMenuPlugin.tsx#L184)):*

Devin AI Comment:

### 1\. Event listener cleanup on root element swap (potential bug)

At ContextMenuPlugin.tsx:183-203, the `registerRootListener` callback receives the new root element but is also called with `null` when the previous root is removed. The current code returns early on `null` (`line 185`), which means it never cleans up the listener on the old root element. The cleanup function returned on `line 201` only runs when `registerRootListener` calls the callback again for the _next_ root, but the Lexical `registerRootListener` API actually passes `(rootElement, prevRootElement)` — you should also remove the listener from `prevRootElement`. As written, if the root element changes (e.g., during editor re-mounting), the old listener would leak.

```
// Suggested fix:return editor.registerRootListener((rootElement, prevRootElement) => {  if (prevRootElement) {    prevRootElement.removeEventListener("contextmenu", handleContextMenu);  }  if (!rootElement) return;  rootElement.addEventListener("contextmenu", handleContextMenu);});
```

Note this requires hoisting `handleContextMenu` out of the callback, or using a stable ref.
</blockquote></details>

Done.
___
*[`libs/shared-react/src/plugins/usj/ContextMenuPlugin.tsx` line 210 at r1](https://reviewable.io/reviews/eten-tech-foundation/scripture-editors/468#-Oq6Ry2v6Z3hUmLcG40y:-Oq81fX33aOHSTLVcdY-:b-896fix) ([raw file](https://github.com/eten-tech-foundation/scripture-editors/blob/4502898e0fc2e80c4a856b7c0d392736e4385ad3/libs/shared-react/src/plugins/usj/ContextMenuPlugin.tsx#L210)):*

Devin AI Comment:

### 2\. Scroll listener is always active

At `libs/shared-react/src/plugins/usj/ContextMenuPlugin.tsx:206-212`, the scroll listener is registered globally at all times (capturing phase), even when the menu is closed. The click-outside and Escape handlers correctly guard with `if (!menuState.isOpen) return;`, but the scroll handler doesn't. This is minor but inconsistent — consider guarding it the same way for consistency and to avoid unnecessary `closeMenu()` calls (which are no-ops but still trigger state setter calls).
</blockquote></details>

Done.
___
*[`libs/shared-react/src/plugins/usj/ContextMenuPlugin.tsx` line 244 at r1](https://reviewable.io/reviews/eten-tech-foundation/scripture-editors/468#-Oq6SYobEtFTPqHfA16M:-Oq81nAf14uXqB6VxqwY:b-896fix) ([raw file](https://github.com/eten-tech-foundation/scripture-editors/blob/4502898e0fc2e80c4a856b7c0d392736e4385ad3/libs/shared-react/src/plugins/usj/ContextMenuPlugin.tsx#L244)):*

Devin AI Comment:

### 5\. Menu can render off-screen

At `libs/shared-react/src/plugins/usj/ContextMenuPlugin.tsx:247-253`, the menu is positioned at the exact click coordinates with `position: fixed`. If the user right-clicks near the bottom or right edge of the viewport, the menu will overflow off-screen. The old implementation relied on Lexical's anchor positioning which may have handled this. Consider adding viewport boundary clamping.
</blockquote></details>

Done.
___
*[`scripts/mcp-launcher.js` line 9 at r1](https://reviewable.io/reviews/eten-tech-foundation/scripture-editors/468#-Oq6TCM09v_3qbL1rcLi:-Oq81slZCLE3UwPk85vm:b-e6zebx) ([raw file](https://github.com/eten-tech-foundation/scripture-editors/blob/4502898e0fc2e80c4a856b7c0d392736e4385ad3/scripts/mcp-launcher.js#L9)):*

NIT Devin AI Comment:

No error handling on spawn failure. If `npx` isn't found at all, the `child.on("error", ...)` event isn't handled. This would result in an unhandled error. Adding a `child.on("error", (err) => { ... })` handler would be more robust.
</blockquote></details>

**Cons**
                                                                                                                                                                                                                
  - With `shell: true`, the shell itself (cmd.exe / sh) is what's spawned — that almost never fails. If `npx` isn't found, the shell starts fine, runs `npx ...`, prints its own "not recognized" / "not found" error to stderr, and exits with a non-zero code. The `child.on("error")` event fires for OS-level spawn failures, not for the child process exiting with an error code. So the scenario the reviewer is worried about is already handled correctly by the existing `exit` handler — you'd just get a non-zero exit code and the shell's own error message on stderr.
  - Adds a handler for an event that is practically unreachable in normal use (`shell: true` + `npx` is a very stable surface).
  - Could mislead future readers into thinking this is the primary error path, when in practice errors surface through `exit`.

**Bottom line**

The suggestion is technically correct as defensive coding, but with `shell: true` it covers a near-impossible case — the shell binary itself missing. The real failure mode (npx not on PATH) already produces the right behavior.


<!-- Sent from Reviewable.io -->

katherinejensen00 (commented):
That makes sense that that is out scope. Thanks for looking into all of those and deciding which ones were worth fixing or not. I was able to learn some things from that.
<img class="emoji" title=":lgtm:" alt=":lgtm:" align="absmiddle" src="https://reviewable.io/lgtm.png" height="20" width="61"/>

> @katherinejensen00 reviewed 1 file and all commit messages, and made 2 comments.
> *[Reviewable](https://reviewable.io/reviews/eten-tech-foundation/scripture-editors/468#-:-Oq8ANHF3YBvW09x4SRV:b-u9570m)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved.

___
*[`scripts/mcp-launcher.js` line 9 at r1](https://reviewable.io/reviews/eten-tech-foundation/scripture-editors/468#-Oq6TCM09v_3qbL1rcLi:-Oq8AjWG-dZaFEnAUvW0:b-5i9scf) ([raw file](https://github.com/eten-tech-foundation/scripture-editors/blob/4502898e0fc2e80c4a856b7c0d392736e4385ad3/scripts/mcp-launcher.js#L9)):*

**Cons**
                                                                                                                                                                                                                
  - With `shell: true`, the shell itself (cmd.exe / sh) is what's spawned — that almost never fails. If `npx` isn't found, the shell starts fine, runs `npx ...`, prints its own "not recognized" / "not found" error to stderr, and exits with a non-zero code. The `child.on("error")` event fires for OS-level spawn failures, not for the child process exiting with an error code. So the scenario the reviewer is worried about is already handled correctly by the existing `exit` handler — you'd just get a non-zero exit code and the shell's own error message on stderr.
  - Adds a handler for an event that is practically unreachable in normal use (`shell: true` + `npx` is a very stable surface).
  - Could mislead future readers into thinking this is the primary error path, when in practice errors surface through `exit`.

**Bottom line**

The suggestion is technically correct as defensive coding, but with `shell: true` it covers a near-impossible case — the shell binary itself missing. The real failure mode (npx not on PATH) already produces the right behavior.
</blockquote></details>

That is helpful to learn. Thanks!


<!-- Sent from Reviewable.io -->

PR #486, opened 2026-06-19
Polish paragraph-structure view: empty-verse placeholder, outline, gutter, RTL

katherinejensen00:
## Context

The platform editor's *Paragraph Structure* view (`markerMode: "hidden"` + `hasGutterParaMarkers: true` + `hasActiveTextFocusBox: true`) had a number of rough edges around the active-paragraph outline, the placeholder shown for empty verses, the rendering of paragraph-level markers in the gutter, and the bidi behavior of mixed-direction content. This pass focused on getting that view into a usable state for prose books.

Video Demo
https://github.com/user-attachments/assets/d7bc013e-1a98-42e6-be5d-5d18a16965a8

### Empty-verse placeholder
- **Per-verse detection.** Replaced `$isParaBodyEmpty` (paragraph-level) with `$getEmptyVerseStatus` (per-verse): walks each `\v` marker and inspects only the children between it and the next verse marker, so the `…` placeholder now shows next to **every** empty verse — not only verses that happen to occupy their own paragraph.
- **CSS selector fix.** The placeholder rule was originally written against `.verse.marker.psc-empty-text::after`, but the `marker` class is only present in `markerMode: "visible"`. In paragraph-structure view the verse number is rendered by `ImmutableVerseNode.decorate()`, so the selector was relaxed to `.verse.psc-empty-text::after` and the `content` was reduced to just `"…"` (the number isn't duplicated).
- **Color was effectively invisible.** `var(--muted-foreground, #6b7280)` produced an invalid color because the platform's `--muted-foreground` holds raw HSL component values (e.g. `215.4 16.3% 46.9%`), not a color. Wrapped in `hsl()` and used a lighter, opacity-reduced fallback so the ellipsis reads as a clear placeholder.
- **Click-to-position.** Clicking the ellipsis (a CSS `::after` whose target is the verse decorator) wasn't dropping the caret into the empty verse's section. Added a root-level click listener that maps the DOM hit back to the Lexical node and calls `para.select(idx, idx)` with the position immediately after the verse.

### Active-paragraph outline
- **Simplified to paragraph-level.** The earlier per-verse outline machinery (`updateTextOutlineBounds`, `ResizeObserver`, `$getActiveVerseSiblings`, `--active-verse-top` / `--active-verse-bottom`) was producing degenerate (flat) boxes whenever the active verse and next verse landed on the same line. Per-verse boxing is deferred to a follow-up work item; the outline now wraps the entire active paragraph with hardcoded `top: -2px; bottom: -2px;`.
- **Gutter marker no longer clipped.** The outline's `inset-inline-start` defaulted to the paragraph content edge, leaving the absolutely-positioned `\p` / `\s1` / `\q1` gutter markers outside. Added a `psc-gutter-markers`-scoped override that picks the further-inset of `--verse-text-start` (poetry hanging indent) and the gutter column itself.
- **Switched outline pseudo to `::after`.** `BookNode`'s data-code (e.g. `HAB`) is rendered via `.marker-hidden .book[data-code]::before`, which has higher specificity than the outline rule and was therefore winning the `content` cascade — so clicking into the `\id` line painted "HAB" inside the absolutely-positioned outline box and dragged it into the gutter. The outline now uses `::after` to avoid the collision.

### Paragraph-level markers in the gutter
- **`\id` now renders in the gutter.** `createBook` in the USJ→editor adaptor only inserted the marker prefix for `markerMode: "editable"` / `"visible"`; the corresponding `createPara` branch checks `hasGutterParaMarkers` too. Added the same branch to `createBook`, and extended the gutter CSS selectors (`.para`, `.marker:first-child`, and the RTL counterpart) to also target `.book`.
- **Section-heading styling normalized.** Removed `text-transform: uppercase`, `letter-spacing: 0.1em`, `font-size: 0.8rem`, `font-weight: 400`, and the muted color from the general section-heading rule so user-typed headings render in their natural case using the standard bold/body-size `.formatted-font .usfm_s1` styling. (No default placeholder text — that's a future work item.)

### Marker deletion resets paragraph type
- Added `ParaNode.updateDOM` so changes to `__marker` actually refresh the `data-marker` attribute and `usfm_*` class on the live element (`createDOM` was the only place these were set).
- Added a Lexical node transform on `ParaNode` (now its own `ParaMarkerPrefixGuardPlugin`) that resets `__marker` to `"p"` when the marker-prefix child has been deleted. The transform is gated on `markerMode in ("editable", "visible") || hasGutterParaMarkers`, since `ParaNode.__marker` is the sole source of truth in views that don't render a prefix child. `$isParaMarkerPrefix` recognises both the immutable prefix (gutter / visible) and the editable `MarkerNode` prefix.

### RTL bidi for paragraph-structure view
- **Empty-verse placeholder spacing.** The hardcoded NBSP inside the ellipsis `content` is a weak bidi character — in RTL it ended up between the ellipsis and the empty verse's own number, making the ellipsis look like it belonged to the next verse. Replaced with logical `margin-inline-start` / `margin-inline-end` so the gaps stay on the correct sides in both directions.
- **Verse atom isolation.** In RTL paragraphs with mixed Latin/English content, the verse number (Unicode EN, weak LTR) bidi-merged with adjacent Latin text into a single LTR run, rendering verse number → text in LTR order even within an RTL paragraph. The pre-existing `.verse.marker` rule handled the `markerMode: "visible"` shape; added the broader paragraph-structure shape under `.psc-gutter-markers[dir="rtl"] .verse`. `unicode-bidi: isolate` alone wasn't enough — an isolated element with no explicit direction is a *neutral* in the parent's bidi flow, and rule N1 promotes neutrals sandwiched between two strong-LTR runs back to LTR. Setting `direction: rtl` makes the isolated unit a *strong* RTL atom so every verse keeps level 1 regardless of neighbors. The digit inside still renders LTR because EN flips to level 2 inside the verse's RTL isolation.

  Beyond fixing the specific "1 3 2" misordering in mixed Arabic + English content, treating each verse as a bidi atom matters because:
  - **A verse marker is a semantic anchor, not a fragment of text.** Scripture verses are discrete editorial units — readers and editors expect each `\v N` to behave the same way regardless of what characters happen to sit next to it. Without isolation, the visual position of a verse number depends on whether its neighbors are Latin, Arabic, neutral punctuation, or another verse number, and the algorithm can shuffle them under bidi rules (W4/N1) that were designed for inline-text directionality, not for editorial markers.
  - **Adjacent empty verses no longer collide.** When two verses sit back-to-back with no text between (e.g. v2 followed immediately by v3 because v2's section is empty), their digits form a single contiguous EN run. At the level-2 reversal pass the run gets reversed in place — which is what produced the "1 3 2" ordering — *before* the level-1 pass reverses the whole line. Isolating each verse turns `[v2, v3]` into two separate atoms, so they keep their DOM order through every reversal pass.
  - **Editing stays stable.** Without isolation, typing or deleting a character next to a verse can cause the bidi algorithm to re-resolve and visually relocate the verse number. With each verse as a strong-direction atom, the verse's position in the flow is determined by its DOM position alone, so the caret and surrounding text don't jump when an unrelated edit changes the surrounding character mix.
  - **Real scripture is mixed content.** Even an RTL document inevitably contains footnote callers, USFM markers, book codes, embedded numbers, and English glosses. Each is an editorial element that should behave consistently in the bidi flow rather than letting its visual position depend on the directionality of the surrounding text run.

## Out of scope / follow-up

- **Per-verse outline boxes.** The active outline currently wraps the whole paragraph. A future pass will reintroduce per-verse section boxing (including Z-shape rendering for verses that begin mid-line) — the `psc-empty-text` per-verse class and `--verse-text-start` plumbing are still in place to support that.
- **Default placeholder text for empty section headings.** Reverted in this pass; a future work item will define the placeholder content and styling.

- - -
This change is [Reviewable](https://reviewable.io/reviews/eten-tech-foundation/scripture-editors/486)

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -220,7 +220,11 @@ When working with scripture data:
 # Code Style
 
 - Prefer `undefined` over `null` when representing missing values unless an API explicitly requires `null`.
-- Prefer chaining Lexical `append` calls inside the final `$getRoot().append(...)` when setting up document structure in tests so the hierarchy is readable at a glance.
+- When building Lexical structures in tests, chain `.append(...)` calls inside `$getRoot().append(...)` so the code shape mirrors the resulting tree. Even when you need a local reference to a node, hoist only the bare `$createXNode(...)` to the variable and put the children inside the root chain — `para = $createParaNode("p"); $getRoot().append(para.append(...children));` — rather than the half-flat `para = $createParaNode("p").append(...); $getRoot().append(para);` shape.
+- Construct Lexical nodes via `$create<X>Node` / `$create<X>Nodes` helpers (singular for one node, plural for an array), matching Lexical's own naming convention. Test helpers follow the same pattern.
+- When a named type alias exists for a union (e.g., `SomeVerseNode` for `VerseNode | ImmutableVerseNode`), use the alias rather than re-spelling the union inline. Aliases live alongside their `$is*` guards.
+- Don't call `editor.update(...)` from inside a listener (event handler, update listener, etc.) — nested updates risk infinite cascade loops. Lexical's command system (`editor.registerCommand`) is the canonical entry point for mutating state in response to user interactions.
+- Order `<*Plugin />` children in `packages/platform/src/editor/Editor.tsx` alphabetically by component name. The alphabetical block starts partway through — initial plugins in the setup section (`OnSelectionChangePlugin`, `DeltaOnChangePlugin`, `ActiveTextPlugin`, …) intentionally precede it.
 
 # Context 7 Library Documentation
 
--- a/libs/shared/src/nodes/usj/BookNode.ts
+++ b/libs/shared/src/nodes/usj/BookNode.ts
@@ -104,6 +104,9 @@ export class BookNode extends ElementNode {
   }
 
   override createDOM(): HTMLElement {
+    // The `.book` class is targeted by platform gutter-positioning rules in usj-nodes.css
+    // alongside `.para` (paragraph-like layout). Changes here that affect the first-child marker
+    // node, the data-code attribute, or the element tag may require corresponding CSS updates.
     const dom = document.createElement("p");
     dom.setAttribute("data-marker", this.__marker);
     dom.classList.add(this.__type, `usfm_${this.__marker}`);
--- a/libs/shared/src/nodes/usj/ParaNode.ts
+++ b/libs/shared/src/nodes/usj/ParaNode.ts
@@ -7,6 +7,7 @@ import {
   DOMConversionMap,
   DOMConversionOutput,
   DOMExportOutput,
+  EditorConfig,
   ElementFormatType,
   LexicalEditor,
   LexicalNode,
@@ -241,6 +242,16 @@ export class ParaNode extends ParagraphNode {
     dom.setAttribute("data-marker", this.__marker);
     dom.classList.add(this.__type, `usfm_${this.__marker}`);
     return dom;
+  }
+
+  override updateDOM(prevNode: this, dom: HTMLElement, config: EditorConfig): boolean {
+    const recreate = super.updateDOM(prevNode, dom, config);
+    if (!recreate && prevNode.__marker !== this.__marker) {
+      dom.setAttribute("data-marker", this.__marker);
+      dom.classList.remove(`usfm_${prevNode.__marker}`);
+      dom.classList.add(`usfm_${this.__marker}`);
+    }
+    return recreate;
   }
 
   override exportDOM(editor: LexicalEditor): DOMExportOutput {
--- a/libs/shared/src/nodes/usj/node.utils.ts
+++ b/libs/shared/src/nodes/usj/node.utils.ts
@@ -532,6 +532,19 @@ export function $isVisibleMarkerNode(
 }
 
 /**
+ * True for either flavor of paragraph-marker node: a `MarkerNode` (markerMode "editable") or an
+ * `ImmutableTypedTextNode` with `textType: "marker"` (markerMode "visible" or gutter views).
+ * These are the two node shapes used to render a paragraph's USFM marker (e.g. `\p`, `\s2`,
+ * `\q1`) as the visible first child of its paragraph.
+ *
+ * @param node - The node to check.
+ * @returns `true` if the node is a `MarkerNode` or a visible marker node.
+ */
+export function $isParaMarkerPrefix(node: LexicalNode | null | undefined): boolean {
+  return $isMarkerNode(node) || $isVisibleMarkerNode(node);
+}
+
+/**
  * Remove all known properties of the `markerObject`.
  * @param markerObject - Scripture marker and its contents.
  * @param markerObjectProps - List of known properties to remove. Defaults to `MARKER_OBJECT_PROPS`.
--- a/packages/platform/src/editor/ActiveTextPlugin.test.tsx
+++ b/packages/platform/src/editor/ActiveTextPlugin.test.tsx
@@ -12,11 +12,15 @@ import {
   ParaNode,
   VerseNode,
 } from "shared";
-import { $createImmutableVerseNode, ImmutableVerseNode } from "shared-react";
+import { $createImmutableVerseNode, ImmutableVerseNode, SomeVerseNode } from "shared-react";
 // Reaching inside only for tests.
 // eslint-disable-next-line @nx/enforce-module-boundaries
 import { createBasicTestEnvironment } from "../../../../libs/shared/src/nodes/usj/test.utils";
-import { $getEmptyVerseStatus, $getParaFromSelection } from "./ActiveTextPlugin";
+import {
+  $getActiveVerseKey,
+  $getEmptyVerseStatus,
+  $getParaFromSelection,
+} from "./ActiveTextPlugin";
 
 const nodes = [
   ParaNode,
@@ -43,7 +47,7 @@ const nodes = [
 type ParaShape = "editable" | "gutter-hidden" | "plain-hidden";
 
 /** Builds the leading paragraph-marker children for the given shape, or [] if the shape has none. */
-function $paraMarkerPrefix(shape: ParaShape, marker: string): LexicalNode[] {
+function $createParaMarkerPrefixNodes(shape: ParaShape, marker: string): LexicalNode[] {
   if (shape === "editable") return [$createMarkerNode(marker), $createTextNode(NBSP)];
   if (shape === "gutter-hidden")
     return [$createImmutableTypedTextNode("marker", `\\${marker}${NBSP}`)];
@@ -61,10 +65,10 @@ describe("$getEmptyVerseStatus", () => {
   describe.each(shapes)("%s shape", (shape) => {
     it("marks the verse empty when it has no body text", () => {
       let para: ParaNode;
-      let v1: VerseNode | ImmutableVerseNode;
+      let v1: SomeVerseNode;
       const { editor } = createBasicTestEnvironment(nodes, () => {
         v1 = $verseFor(shape, "1");
-        para = $createParaNode("p").append(...$paraMarkerPrefix(shape, "p"), v1);
+        para = $createParaNode("p").append(...$createParaMarkerPrefixNodes(shape, "p"), v1);
         $getRoot().append(para);
       });
 
@@ -78,11 +82,11 @@ describe("$getEmptyVerseStatus", () => {
 
     it("marks the verse non-empty when it has body text", () => {
       let para: ParaNode;
-      let v1: VerseNode | ImmutableVerseNode;
+      let v1: SomeVerseNode;
       const { editor } = createBasicTestEnvironment(nodes, () => {
         v1 = $verseFor(shape, "1");
         para = $createParaNode("p").append(
-          ...$paraMarkerPrefix(shape, "p"),
+          ...$createParaMarkerPrefixNodes(shape, "p"),
           v1,
           $createTextNode("In the beginning"),
         );
@@ -99,11 +103,11 @@ describe("$getEmptyVerseStatus", () => {
 
     it("marks the verse empty when it is followed by whitespace only", () => {
       let para: ParaNode;
-      let v1: VerseNode | ImmutableVerseNode;
+      let v1: SomeVerseNode;
       const { editor } = createBasicTestEnvironment(nodes, () => {
         v1 = $verseFor(shape, "1");
         para = $createParaNode("p").append(
-          ...$paraMarkerPrefix(shape, "p"),
+          ...$createParaMarkerPrefixNodes(shape, "p"),
           v1,
           $createTextNode("   "),
         );
@@ -120,13 +124,13 @@ describe("$getEmptyVerseStatus", () => {
 
     it("marks only the empty verse when one of two verses in the paragraph has no body", () => {
       let para: ParaNode;
-      let v1: VerseNode | ImmutableVerseNode;
-      let v2: VerseNode | ImmutableVerseNode;
+      let v1: SomeVerseNode;
+      let v2: SomeVerseNode;
       const { editor } = createBasicTestEnvironment(nodes, () => {
         v1 = $verseFor(shape, "1");
         v2 = $verseFor(shape, "2");
         para = $createParaNode("p").append(
-          ...$paraMarkerPrefix(shape, "p"),
+          ...$createParaMarkerPrefixNodes(shape, "p"),
           v1,
           v2,
           $createTextNode("In the beginning"),
@@ -144,13 +148,13 @@ describe("$getEmptyVerseStatus", () => {
 
     it("marks the trailing verse empty when it has no body after a non-empty preceding verse", () => {
       let para: ParaNode;
-      let v1: VerseNode | ImmutableVerseNode;
-      let v2: VerseNode | ImmutableVerseNode;
+      let v1: SomeVerseNode;
+      let v2: SomeVerseNode;
       const { editor } = createBasicTestEnvironment(nodes, () => {
         v1 = $verseFor(shape, "1");
         v2 = $verseFor(shape, "2");
         para = $createParaNode("p").append(
-          ...$paraMarkerPrefix(shape, "p"),
+          ...$createParaMarkerPrefixNodes(shape, "p"),
           v1,
           $createTextNode("text one"),
           v2,
@@ -170,7 +174,7 @@ describe("$getEmptyVerseStatus", () => {
       let para: ParaNode;
       const { editor } = createBasicTestEnvironment(nodes, () => {
         para = $createParaNode("q2").append(
-          ...$paraMarkerPrefix(shape, "q2"),
+          ...$createParaMarkerPrefixNodes(shape, "q2"),
           $createTextNode("continuation"),
         );
         $getRoot().append(para);
@@ -186,11 +190,11 @@ describe("$getEmptyVerseStatus", () => {
 
     it("ignores text before the first verse (intro text does not affect classification)", () => {
       let para: ParaNode;
-      let v1: VerseNode | ImmutableVerseNode;
+      let v1: SomeVerseNode;
       const { editor } = createBasicTestEnvironment(nodes, () => {
         v1 = $verseFor(shape, "1");
         para = $createParaNode("p").append(
-          ...$paraMarkerPrefix(shape, "p"),
+          ...$createParaMarkerPrefixNodes(shape, "p"),
           $createTextNode("pre-verse intro"),
           v1,
         );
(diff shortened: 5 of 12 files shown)


irahopkinson (changes requested):
Good job with all this clean-up

> @irahopkinson reviewed 10 files and all commit messages, and made 13 comments.
> *[Reviewable](https://reviewable.io/reviews/eten-tech-foundation/scripture-editors/486#-:-OvgwzBT3627C0-aCvXf:b421s09)* status: all files reviewed, 12 unresolved discussions (waiting on katherinejensen00).

___
*[`packages/platform/src/usj-nodes.css` line 2439 at r1](https://reviewable.io/reviews/eten-tech-foundation/scripture-editors/486#-OvgfdAS2y7Jf2aRRfbo:-OvgfdAS2y7Jf2aRRfbp:b-kk5y4a) ([raw file](https://github.com/eten-tech-foundation/scripture-editors/blob/eb2d8a69dd7ce1156089a339ef8aed7e6a12da8f/packages/platform/src/usj-nodes.css#L2439)):*
> ```CSS
>    we attach the ellipsis to `.verse` directly.) */
> .psc-gutter-markers .verse.psc-empty-text::after {
>   content: "…";
> ```

The ellipsis is a placeholder and should go away when the verse is selected.
___
*[`packages/platform/src/usj-nodes.css` line 2441 at r1](https://reviewable.io/reviews/eten-tech-foundation/scripture-editors/486#-OvgM5zqA-xwhVcs3VCY:-OvgM5zqA-xwhVcs3VCZ:b-mg7ha) ([raw file](https://github.com/eten-tech-foundation/scripture-editors/blob/eb2d8a69dd7ce1156089a339ef8aed7e6a12da8f/packages/platform/src/usj-nodes.css#L2441)):*
> ```CSS
>   content: "…";
>   /* Inline-logical margins so the spacing flips correctly in RTL: a hardcoded NBSP inside
>      `content` is a weak bidi character and ends up between the ellipsis and its own verse
> ```

BTW it seems odd to be talking about an NBSP here since it didn't have one in `content` before this change and still doesn't. Can we fix this comment?



_Code quote:_
```CSS
  /* Inline-logical margins so the spacing flips correctly in RTL: a hardcoded NBSP inside
     `content` is a weak bidi character and ends up between the ellipsis and its own verse
```
___
*[`packages/platform/src/usj-nodes.css` line 2451 at r1](https://reviewable.io/reviews/eten-tech-foundation/scripture-editors/486#-OvgfKwBAcaUHpCK9yjJ:-OvgfKwBAcaUHpCK9yjK:bheg8hj) ([raw file](https://github.com/eten-tech-foundation/scripture-editors/blob/eb2d8a69dd7ce1156089a339ef8aed7e6a12da8f/packages/platform/src/usj-nodes.css#L2451)):*
> ```CSS
>   color: hsl(var(--muted-foreground, 215 16% 65%) / 0.7);
>   font-style: italic;
>   font-weight: normal;
> ```

Add `cursor: text;` here so you at least get an I-beam cursor on the ellipsis. Ideally we'd also get the I-beam on the padding and margin between the ellipsis and the next inline item.
___
*[`packages/platform/src/editor/ActiveTextPlugin.tsx` line 59 at r1](https://reviewable.io/reviews/eten-tech-foundation/scripture-editors/486#-OvgbjoK7kyopDK4bT_B:-OvgbjoK7kyopDK4bT_C:b-x54bkc) ([raw file](https://github.com/eten-tech-foundation/scripture-editors/blob/eb2d8a69dd7ce1156089a339ef8aed7e6a12da8f/packages/platform/src/editor/ActiveTextPlugin.tsx#L59)):*
> ```XML
>         if (!$isElementNode(para)) return;
>         const after = node.getIndexWithinParent() + 1;
>         para.select(after, after);
> ```

This is one of the gotchas of Lexical - you shouldn't run `editor.update(...)` inside a listener because doing so can easily trigger an infinite loop or cause unexpected rendering behavior. See [Gemini's response](https://gemini.google.com/share/f465043130b1).

If you don't find this reasonably easy to fix then I'll take a TODO comment saying this needs fixing as long as it doesn't actually cause an update cascade already. To prove this, if you put a breakpoint on this line it should only fire once when you click the ellipsis.
___
*[`packages/platform/src/editor/ActiveTextPlugin.test.tsx` line 64 at r1](https://reviewable.io/reviews/eten-tech-foundation/scripture-editors/486#-OvgRWrgASWivWeXYRav:-OvgRWrgASWivWeXYRaw:befpexo) ([raw file](https://github.com/eten-tech-foundation/scripture-editors/blob/eb2d8a69dd7ce1156089a339ef8aed7e6a12da8f/packages/platform/src/editor/ActiveTextPlugin.test.tsx#L64)):*
> ```XML
>     it("marks the verse empty when it has no body text", () => {
>       let para: ParaNode;
>       let v1: VerseNode | ImmutableVerseNode;
> ```

NIT can use `SomeVerseNode`. Here and below.



_Code quote:_
```XML
VerseNode | ImmutableVerseNode
```
___
*[`packages/platform/src/editor/Editor.tsx` line 450 at r1](https://reviewable.io/reviews/eten-tech-foundation/scripture-editors/486#-OvgNPDg81u02fgPxRbX:-OvgNPDg81u02fgPxRbY:bg82uuw) ([raw file](https://github.com/eten-tech-foundation/scripture-editors/blob/eb2d8a69dd7ce1156089a339ef8aed7e6a12da8f/packages/platform/src/editor/Editor.tsx#L450)):*
> ```XML
>           />
>           <ActiveTextPlugin viewOptions={viewOptions} />
>           <ParaMarkerPrefixGuardPlugin viewOptions={viewOptions} logger={logger} />
> ```

NIT alphabetic order for this puts it above `<ParaNodePlugin />` and not here.
___
*[`packages/platform/src/editor/ParaMarkerPrefixGuardPlugin.tsx` line 14 at r1](https://reviewable.io/reviews/eten-tech-foundation/scripture-editors/486#-OvgV39T7mrG455mVmsh:-OvgV39T7mrG455mVmsi:by20jv6) ([raw file](https://github.com/eten-tech-foundation/scripture-editors/blob/eb2d8a69dd7ce1156089a339ef8aed7e6a12da8f/packages/platform/src/editor/ParaMarkerPrefixGuardPlugin.tsx#L14)):*
> ```XML
> 
> /**
>  * Resets a paragraph's marker to default when the user deletes its leading marker-prefix child.
> ```

I don't understand. What is a _"leading marker-prefix child"_ or _"user's handle for the paragraph's type"_ in this context? This TSDoc doesn't really explain what this does and why.
___
*[`packages/platform/src/editor/ParaMarkerPrefixGuardPlugin.tsx` line 62 at r1](https://reviewable.io/reviews/eten-tech-foundation/scripture-editors/486#-OvgY82X9DaXKG_lTuU7:-OvgY82X9DaXKG_lTuU8:b-skiic9) ([raw file](https://github.com/eten-tech-foundation/scripture-editors/blob/eb2d8a69dd7ce1156089a339ef8aed7e6a12da8f/packages/platform/src/editor/ParaMarkerPrefixGuardPlugin.tsx#L62)):*
> ```XML
>   if (para.isEmpty()) return;
>   const firstChild = para.getFirstChild();
>   if (firstChild && $isParaMarkerPrefix(firstChild)) return;
> ```

NIT I don't think you need this if the `$isParaMarkerPrefix` function allows `null` or `undefined` on the argument.



_Code quote:_
```XML
firstChild && 
```
___
*[`packages/platform/src/editor/ParaMarkerPrefixGuardPlugin.tsx` line 77 at r1](https://reviewable.io/reviews/eten-tech-foundation/scripture-editors/486#-OvgW2O6Br-arXquCY6J:-OvgW2O6Br-arXquCY6K:bf0dl6v) ([raw file](https://github.com/eten-tech-foundation/scripture-editors/blob/eb2d8a69dd7ce1156089a339ef8aed7e6a12da8f/packages/platform/src/editor/ParaMarkerPrefixGuardPlugin.tsx#L77)):*
> ```XML
>   if ($isMarkerNode(node)) return true;
>   if ($isImmutableTypedTextNode(node)) return node.getTextType() === "marker";
>   return false;
> ```

Replace with `return $isMarkerNode(node) || $isVisibleMarkerNode(node);`
This whole function likely belongs in `libs/shared/src/nodes/usj/node.utils.ts`.



_Code quote:_
```XML
  if ($isMarkerNode(node)) return true;
  if ($isImmutableTypedTextNode(node)) return node.getTextType() === "marker";
  return false;
```
___
*[`packages/platform/src/editor/ParaMarkerPrefixGuardPlugin.test.tsx` line 40 at r1](https://reviewable.io/reviews/eten-tech-foundation/scripture-editors/486#-OvgYgNf0QPmZkyg05Kq:-OvgYgNf0QPmZkyg05Kr:bxcfdit) ([raw file](https://github.com/eten-tech-foundation/scripture-editors/blob/eb2d8a69dd7ce1156089a339ef8aed7e6a12da8f/packages/platform/src/editor/ParaMarkerPrefixGuardPlugin.test.tsx#L40)):*
> ```XML
> type PrefixShape = "editable" | "gutter-hidden";
> 
> function $paraMarkerPrefix(shape: PrefixShape, marker: string): LexicalNode[] {
> ```

NIT following the existing pattern rename to `$createParaMarkerPrefixNodes`.



_Code quote:_
```XML
$paraMarkerPrefix
```
___
*[`packages/platform/src/editor/ParaMarkerPrefixGuardPlugin.test.tsx` line 52 at r1](https://reviewable.io/reviews/eten-tech-foundation/scripture-editors/486#-OvgZZpF7riVuMSUtJCH:-OvgZZpF7riVuMSUtJCI:b-qbeo3j) ([raw file](https://github.com/eten-tech-foundation/scripture-editors/blob/eb2d8a69dd7ce1156089a339ef8aed7e6a12da8f/packages/platform/src/editor/ParaMarkerPrefixGuardPlugin.test.tsx#L52)):*
> ```XML
>       let para: ParaNode;
>       const { editor } = createBasicTestEnvironment(nodes, () => {
>         para = $createParaNode("q1").append($createTextNode("a poetry line"));
> ```

NIT this append should be moved to the line below. Here and tests below this one.



_Code quote:_
```XML
.append($createTextNode("a poetry line"))
```
___
*[`packages/platform/src/editor/adaptors/usj-editor.adaptor.ts` line 238 at r1](https://reviewable.io/reviews/eten-tech-foundation/scripture-editors/486#-Ovg_BRr4IxF4QkuhNST:-Ovg_BRr4IxF4QkuhNSU:b3tppw7) ([raw file](https://github.com/eten-tech-foundation/scripture-editors/blob/eb2d8a69dd7ce1156089a339ef8aed7e6a12da8f/packages/platform/src/editor/adaptors/usj-editor.adaptor.ts#L238)):*
> ```XML
>     // Gutter mode hides inline markers, but the paragraph-structure view still wants the \id
>     // tag visible in the gutter alongside the other paragraph-level markers (\h, \s1, \p, ...).
>     children.push(createImmutableTypedText("marker", openingMarkerText(marker) + NBSP));
> ```

DRYer: don't duplicate this line. Just add the condition to the `if` above



_Code quote:_
```XML
children.push(createImmutableTypedText("marker", openingMarkerText(marker) + NBSP));
```

katherinejensen00 (commented):
Thank you for your helpful feedback, Ira! I learned some new things and I was able to add some new guidelines to help train the editor Claude AI. 

> @katherinejensen00 made 7 comments and resolved 6 discussions.
> *[Reviewable](https://reviewable.io/reviews/eten-tech-foundation/scripture-editors/486#-:-Ovknkd61tv2mjqiTyhk:b-487s31)* status: all files reviewed, 6 unresolved discussions (waiting on irahopkinson).

___
*[`packages/platform/src/usj-nodes.css` line 2439 at r1](https://reviewable.io/reviews/eten-tech-foundation/scripture-editors/486#-OvgfdAS2y7Jf2aRRfbo:-Ovk_E9j1AeZ76ZbqGAV:b-jhg1ir) ([raw file](https://github.com/eten-tech-foundation/scripture-editors/blob/eb2d8a69dd7ce1156089a339ef8aed7e6a12da8f/packages/platform/src/usj-nodes.css#L2439)):*

The ellipsis is a placeholder and should go away when the verse is selected.
</blockquote></details>

Good point. I fixed it so that when the user clicks in the verse, the ellipsis goes away even if they don't add text (but it will reappear if they click out of the verse without adding text).
___
*[`packages/platform/src/usj-nodes.css` line 2451 at r1](https://reviewable.io/reviews/eten-tech-foundation/scripture-editors/486#-OvgfKwBAcaUHpCK9yjJ:-OvkadyDAJwojjlylbSF:b-sm9nd7) ([raw file](https://github.com/eten-tech-foundation/scripture-editors/blob/eb2d8a69dd7ce1156089a339ef8aed7e6a12da8f/packages/platform/src/usj-nodes.css#L2451)):*

Add `cursor: text;` here so you at least get an I-beam cursor on the ellipsis. Ideally we'd also get the I-beam on the padding and margin between the ellipsis and the next inline item.
</blockquote></details>

Good idea. Added.
___
*[`packages/platform/src/editor/ActiveTextPlugin.tsx` line 59 at r1](https://reviewable.io/reviews/eten-tech-foundation/scripture-editors/486#-OvgbjoK7kyopDK4bT_B:-OvkctHJ7889BCaF1SQr:b843o41) ([raw file](https://github.com/eten-tech-foundation/scripture-editors/blob/eb2d8a69dd7ce1156089a339ef8aed7e6a12da8f/packages/platform/src/editor/ActiveTextPlugin.tsx#L59)):*

This is one of the gotchas of Lexical - you shouldn't run `editor.update(...)` inside a listener because doing so can easily trigger an infinite loop or cause unexpected rendering behavior. See [Gemini's response](https://gemini.google.com/share/f465043130b1).

If you don't find this reasonably easy to fix then I'll take a TODO comment saying this needs fixing as long as it doesn't actually cause an update cascade already. To prove this, if you put a breakpoint on this line it should only fire once when you click the ellipsis.
</blockquote></details>

That is helpful to know. Thanks! Adjusted my approach accordingly.
___
*[`packages/platform/src/editor/ParaMarkerPrefixGuardPlugin.tsx` line 14 at r1](https://reviewable.io/reviews/eten-tech-foundation/scripture-editors/486#-OvgV39T7mrG455mVmsh:-Ovkf2wT0Q7nUjES1wb9:b-mg809y) ([raw file](https://github.com/eten-tech-foundation/scripture-editors/blob/eb2d8a69dd7ce1156089a339ef8aed7e6a12da8f/packages/platform/src/editor/ParaMarkerPrefixGuardPlugin.tsx#L14)):*

I don't understand. What is a _"leading marker-prefix child"_ or _"user's handle for the paragraph's type"_ in this context? This TSDoc doesn't really explain what this does and why.
</blockquote></details>

Fair, clarified.
___
*[`packages/platform/src/editor/ParaMarkerPrefixGuardPlugin.tsx` line 77 at r1](https://reviewable.io/reviews/eten-tech-foundation/scripture-editors/486#-OvgW2O6Br-arXquCY6J:-OvkhW8GD5o8sWDnclNH:b-mrcyce) ([raw file](https://github.com/eten-tech-foundation/scripture-editors/blob/eb2d8a69dd7ce1156089a339ef8aed7e6a12da8f/packages/platform/src/editor/ParaMarkerPrefixGuardPlugin.tsx#L77)):*

Replace with `return $isMarkerNode(node) || $isVisibleMarkerNode(node);`
This whole function likely belongs in `libs/shared/src/nodes/usj/node.utils.ts`.
</blockquote></details>

Good idea. Moved.
___
*[`packages/platform/src/editor/adaptors/usj-editor.adaptor.ts` line 238 at r1](https://reviewable.io/reviews/eten-tech-foundation/scripture-editors/486#-Ovg_BRr4IxF4QkuhNST:-Ovkn7hXC3JmWeCmYYQh:bgjtzhe) ([raw file](https://github.com/eten-tech-foundation/scripture-editors/blob/eb2d8a69dd7ce1156089a339ef8aed7e6a12da8f/packages/platform/src/editor/adaptors/usj-editor.adaptor.ts#L238)):*

DRYer: don't duplicate this line. Just add the condition to the `if` above
</blockquote></details>

There is a subtle difference in behavior with having the code duplicated here. This makes it so in the paragraph level marker gutter, only the "\id" shows in the gutter while the book name, "GEN" shows in the text after it (which I think is expected because I don't think the book title is part of the marker). If I combine the lines, then "\id GEN" will display in the paragraph level marker gutter. 

My other option is to put a ternary condition instead of the if/else, but that is harder to read at a glance. I am happy to go with that option though to work with the DRY principle. What do you think?

irahopkinson (approved):
:lgtm: thanks so much

> @irahopkinson reviewed 8 files and all commit messages, made 2 comments, and resolved 6 discussions.
> *[Reviewable](https://reviewable.io/reviews/eten-tech-foundation/scripture-editors/486#-:-OvleX9FAx0JNqhS1bQU:bty9m1)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved.

___
*[`CLAUDE.md` at r3](https://reviewable.io/reviews/eten-tech-foundation/scripture-editors/486#-OvlcigR07L37BGYt3x2:-OvlcigR07L37BGYt3x3:bhismt) ([raw file](https://github.com/eten-tech-foundation/scripture-editors/blob/0c39415419426875b1049a6a1e99f796839614b9/CLAUDE.md)):*
FYI thanks for doing this. We can tweak it if it's not doing what we want.

PR #492, opened 2026-06-30
PT-4020 Backspace confirmation

jolierabideau:
# Code Review Summary

**Branch**: pt-4020-backspace-confirmation

**Base**: origin/main

**Date**: 2026-07-01

**Review model**: Claude Opus 4.8

**Files changed**: 10

## Overview

This branch makes deleting structural verse/paragraph markers a deliberate two-step
gesture, so they can no longer be removed with a single accidental keystroke. When the
editor is *not* in structure-protected mode, the first Backspace/Delete at a marker
boundary "arms" the target — a verse marker is put into a `NodeSelection` and rendered
with destructive styling (`verse-selected`), while a paragraph is selected as a range.
A second matching keypress fires the delete (removing the verse marker, or merging the
paragraph into the previous one); a mismatched-direction key, arrow, click, or typing
clears the armed latch instead. The former `StructureProtectionPlugin` was renamed to
`StructureKeyboardPlugin` since it now governs both PT-4013 structure protection and the
new two-step delete. Supporting work includes a `VerseDecorator` that reflects node
selection, a `VERSE_SELECTED_CLASS_NAME` export, a themeable `.verse-selected` demo
stylesheet rule, and an `ImmutableVerseNode.isSelected` override that guards against an
expected Lexical element-point `getNodes()` throw.

## API Changes

- `shared-react` (`libs/shared-react/src/plugins/usj`): Removed `StructureProtectionPlugin` export (file deleted, no longer re-exported from `plugins/usj/index.ts`).
- `shared-react` (`libs/shared-react/src/plugins/usj`): Added `StructureKeyboardPlugin` export (new file `StructureKeyboardPlugin.tsx`, re-exported from `plugins/usj/index.ts`). Same props shape as the removed plugin: `{ isStructureProtected: boolean }`, returns `null`.
- `shared-react` (`libs/shared-react/src/nodes/usj/ImmutableVerseNode.tsx`): Added `VERSE_SELECTED_CLASS_NAME` export (`"verse-selected"`).
- `shared-react` (`libs/shared-react/src/nodes/usj/ImmutableVerseNode.tsx`): Added `isSelected(selection?: BaseSelection | null): boolean` override on `ImmutableVerseNode` (matches Lexical base signature; non-breaking refinement).
- Internal only (not on any public barrel): `structureKeyboard.utils.ts` gained new helpers (`$structuralDeleteTarget`, `$isArmedSelection`, `$placeCaretAtEnd`, `$mergeParaIntoPrevious`, `DeleteTarget`, `ArmedDelete`, etc.). Not part of the public API surface.

- [ ] ~~Public export `StructureProtectionPlugin` renamed to `StructureKeyboardPlugin` in `libs/shared-react/src/plugins/usj/index.ts` — a breaking change for any external consumer importing the old name.~~ _(Author: paranext-core does not import this plugin by name; it enables it through an editor option, so no external consumer depends on the export name. Not a real breaking change.)_

[Author response: Dismissed — the plugin is not consumed by name externally; consumers enable it via an editor option, so the rename is behavior- and consumer-safe.]

- [ ] ~~Stale build artifacts (`libs/shared-react/dist/`, `out-tsc/`) still reference the removed `StructureProtectionPlugin`; a stale `.d.ts` could mask the rename if `dist/` is published.~~ _(Verified `dist/` and `out-tsc/` are gitignored and untracked — not in the diff or committed anywhere; a clean rebuild regenerates them.)_
- [ ] The armed "pending delete" state is communicated by color only (red text on light-red background plus `border-radius: 2px`), with no non-color affordance — an accessibility gap for color-vision-deficient users and high-contrast themes. _(Deferred: author will address armed-state affordances holistically in a follow-up.)_
- [ ] Inconsistent visual feedback between armed states: an armed verse marker gets destructive `.verse-selected` styling, but an armed paragraph is shown via a plain native range selection that looks identical to ordinary text selection — no "next press is destructive" signal. _(Deferred: making the paragraph case look destructive is non-trivial — paragraphs aren't decorator nodes, so it would require switching paragraph arming to a `NodeSelection` and teaching `ParaNode` to render a destructive class, with some risk to the existing range-based merge/caret logic. Author chose to defer to the follow-up.)_
- [ ] No screen-reader announcement when a marker becomes armed (no `aria-live` / ARIA state), so non-sighted users get no signal that the next keypress will delete. _(Deferred: there is no existing `aria-live`/announcement pattern anywhere in the codebase — all 13 aria usages are static attributes — so adding one is net-new accessibility infrastructure, folded into the follow-up.)_

[Author response: The stale-artifact finding was dismissed after verifying the paths are gitignored/untracked. The three UX/accessibility findings (color-only cue, inconsistent verse-vs-paragraph feedback, no SR announcement) are acknowledged as legitimate but intentionally deferred to a dedicated follow-up that will handle armed-state affordances holistically; #4 is non-trivial and #5 has no existing pattern to extend.]

### Minor — Consider

- [x] **File-name inconsistency**: the plugin/test were renamed to `StructureKeyboard*`, but the support trio kept the old `structureProtection.*` stem. _(fixed during review: `git mv` renamed `structureProtection.utils.ts`/`.utils.test.ts`/`.baseline.test.tsx` to `structureKeyboard.*`; updated the import in `StructureKeyboardPlugin.tsx`, the import + `describe` label in `structureKeyboard.utils.test.ts`, and a comment reference in `StructureKeyboardPlugin.test.tsx`.)_
- [x] **DRY duplication**: `$caretAdjacentToVerseMarker` and `$adjacentVerseMarker` in `structureKeyboard.utils.ts` reimplemented the same boundary-resolution logic line-for-line. _(fixed during review: collapsed `$caretAdjacentToVerseMarker` to `return !!$adjacentVerseMarker(selection, direction);`, making `$adjacentVerseMarker` the single source of truth.)_
- [ ] ~~`.verse-selected` in `packages/platform/src/usj-nodes.css:2269-2279` uses literal hex reds (`#dc2626`) rather than a theme token.~~ _(Intentional and well-commented: the demo theme stores `--destructive` as HSL components that can't be referenced directly; the platform app rebinds `.verse-selected` to the live `--destructive` token in paranext-core's `_usj-nodes.scss`. Consistent with other hardcoded colors in the demo stylesheet.)_
- [ ] New public string contract `VERSE_SELECTED_CLASS_NAME = "verse-selected"` (`ImmutableVerseNode.tsx:36`) — a clean, documented addition; noted only that host apps may now depend on the string value, so treat it as a stable contract going forward. (FYI, no action.)

[Author response: Requested and received fixes for the two style items (file rename + DRY collapse). The demo-CSS hardcoded-color note was dismissed as intentional/documented; the new public class-name constant is an accepted FYI.]

#### Shared Regions Modified

None. (No `#region shared with <url>` markers exist in any changed file.)

#### Extension Config Changes

None. (This is the `scripture-editors` Nx monorepo — no `extensions/` tree.)

#### Other Propagation Notes

- [ ] `packages/platform/src/usj-nodes.css:2269-2279` — The demo `.verse-selected` rule is the counterpart of styling the platform app maintains in paranext-core's `_usj-nodes.scss` (bound to the live `--destructive` token, per the in-file comment). If the armed/selected verse-marker visual treatment changes here, the paranext-core rule may need a matching update. This is a code-comment-documented cross-repo coupling, not a formal shared-region marker.

## Positive Observations

- The two-step delete is a strong UX safeguard: it converts an easy-to-trigger destructive keystroke into a deliberate, reversible gesture, with the latch cleared on any arrow/click/typing (`clearStaleLatch`).
- Thoughtful mismatched-direction handling: arming with Backspace then pressing Delete cancels rather than deletes (`if (intent !== armed.intent) return true;`).
- Paragraph deletion uses merge-into-previous semantics that never lose text and place the caret at the junction, preserving edit position.
- The new `StructureKeyboardPlugin` preserves the exact props contract (`{ isStructureProtected: boolean }`) and `null` return of the plugin it replaces — the migration is a one-line rename for consumers.
- `ImmutableVerseNode.isSelected` matches Lexical's base signature exactly and reuses the existing public `isSelectionStartNodeExpectedError` helper rather than duplicating error-detection logic; the guard only swallows the one known-benign element-point throw and re-throws anything else.
- The `decorate()` refactor into `VerseDecorator` is purely internal rendering; `SerializedImmutableVerseNode` and all top-level exports are unchanged.
- New internal helpers were correctly kept off the public barrel, avoiding accidental public-API expansion.
- Strong test coverage: `structureKeyboard.utils.test.ts` (46 cases) exercises boundary detection and target resolution; `StructureKeyboardPlugin.test.tsx` (21 cases) covers the protected-blocking path and the arm/fire/cancel flow incl. mismatched-direction cancel and stale-latch guards; `ImmutableVerseNode.test.tsx` (3 new cases) covers selected-class rendering and the element-point edge case.
- `<StructureKeyboardPlugin />` is correctly placed in the alphabetical plugin block in `Editor.tsx`, honoring the CLAUDE.md ordering rule.
- No user-visible strings added (styling + keyboard behavior only) — no localization concerns.

## Interview Notes

- **Stated purpose**: Add an intentional two-step deletion for verse/paragraph markers to prevent too-easily deleting them via keyboard. First delete selects the marker and gives it destructive styling; second delete removes it.
- **Public export rename** (Critical): Author clarified paranext-core does not import the plugin by name — it enables the behavior through an editor option — so the `StructureProtectionPlugin` → `StructureKeyboardPlugin` rename is not a breaking change for real consumers. No deprecation alias needed.
- **UX/accessibility findings**: Author agreed the color-only armed cue, the verse-vs-paragraph feedback inconsistency, and the missing screen-reader announcement are legitimate but chose to defer them to a dedicated follow-up that handles armed-state affordances holistically. Rationale recorded: the paragraph-styling change is non-trivial (paragraphs aren't decorator nodes), and there is no existing `aria-live` pattern in the codebase to extend.
- **Design understanding — `isSelected` override**: Author asked for a walkthrough of the override rather than explaining it unprompted, then confirmed the explanation matched their mental model. The reasoning: `ImmutableVerseNode` is a `DecoratorNode` whose `VerseDecorator` subscribes via `useLexicalNodeSelection`, which calls `isSelected` → base `getNodes()`, which throws on a `RangeSelection` with an element-type point on the decorator (caret on the verse number). The override catches only that expected throw (matched via `isSelectionStartNodeExpectedError`) and returns `false`, re-throwing anything else; a `try/catch` was chosen because the node must remain a `DecoratorNode`. Author confirmed this matches their understanding — no unresolved gap, though a reviewer may still want to spot-check this subtle area.
- **Nothing else flagged** at wrap-up.

## In-Review Quality Check

Two minor findings were fixed during the review (file-trio rename to `structureKeyboard.*`, and collapsing `$caretAdjacentToVerseMarker` to delegate to `$adjacentVerseMarker`). Quality checks on `shared-react` after the changes:

- `nx typecheck shared-react` — passed clean. (The editor's transient `describe`/`it`/`expect`/`vi` "cannot find name" diagnostics on the renamed test files did not appear in the real typecheck — confirmed spurious tsserver artifacts.)
- `nx lint shared-react` — 0 errors. 3 pre-existing warnings in unrelated files (`filterAndRankItems.ts`, `selection.utils.data-driven.test.ts`) left untouched.
- Format (changed files only) — all 10 files already formatted; no changes.
- `nx test shared-react` — 1076 passed / 3 skipped across 19 test files; the renamed `structureKeyboard.*` tests ran and passed.

No fixes beyond the requested changes were required.

## Suggested Review Focus

Prioritized areas for the author-reviewer meeting:

- [ ] Confirm the `StructureProtectionPlugin` → `StructureKeyboardPlugin` rename is safe: verify paranext-core (and any other consumer) truly enables the plugin via an editor option and never imports it by name.
- [ ] Armed-state affordances follow-up: agree on scope for the deferred UX/accessibility work — non-color cue for verses, consistent destructive styling for armed paragraphs, and a screen-reader announcement (net-new `aria-live` infrastructure).
- [ ] Spot-check the `ImmutableVerseNode.isSelected` element-point guard: confirm `isSelectionStartNodeExpectedError` reliably identifies only the benign throw and that no other selection shape is silently swallowed.
- [ ] Review the two-step latch state machine for edge cases (arm → clear on arrow/click/typing; mismatched-direction cancel; paragraph merge caret placement).

---

This change is [Reviewable](https://reviewable.io/reviews/eten-tech-foundation/scripture-editors/492)


captaincrazybro (commented):
> @captaincrazybro partially reviewed 17 files and all commit messages, and made 7 comments.
> *[Reviewable](https://reviewable.io/reviews/eten-tech-foundation/scripture-editors/492)* status: all files reviewed, 7 unresolved discussions (waiting on jolierabideau).

___
*[a discussion](https://reviewable.io/reviews/eten-tech-foundation/scripture-editors/492#-OwV7SK43O2WVdi-OUyY:-OwV7SK43O2WVdi-OUyZ:b-4bhp71) (no related file):*
Just had a few mostly documentation and style comments with one potential bug reported by Devin.
Probably should have Ira look at this as I don't have access to actually approve.
___
*[`libs/shared-react/src/plugins/usj/armedDeleteMessage.ts` line 10 at r1](https://reviewable.io/reviews/eten-tech-foundation/scripture-editors/492#-OwV4ojNAn-FSQDHK32R:-OwV4ojNAn-FSQDHK32S:b-1dhwe2) ([raw file](https://github.com/eten-tech-foundation/scripture-editors/blob/706ce60160ecc20637a5403ee601a1196dfd7c92/libs/shared-react/src/plugins/usj/armedDeleteMessage.ts#L10)):*
> ```XML
>   kind: ArmedDelete["kind"],
>   intent: ArmedDelete["intent"],
> ): { key: string; text: string } {
> ```

Add JSDoc comments for parameters? And maybe return JSDoc?
___
*[`libs/shared-react/src/plugins/usj/StructureKeyboardPlugin.tsx` line 48 at r1](https://reviewable.io/reviews/eten-tech-foundation/scripture-editors/492#-OwV56tOCeXwc8akE7Zk:-OwV56tOCeXwc8akE7Zl:bdpkxa0) ([raw file](https://github.com/eten-tech-foundation/scripture-editors/blob/706ce60160ecc20637a5403ee601a1196dfd7c92/libs/shared-react/src/plugins/usj/StructureKeyboardPlugin.tsx#L48)):*
> ```XML
>  * @param isStructureProtected - When true, structural keystrokes are blocked; when false, the
>  *   two-step intentional-delete behavior is active.
>  * @returns Always `null`; this component has no UI.
> ```

This JSDoc comment doesn't seem to be correct. Seems like it does sometimes return JSX.
___
*[`libs/shared-react/src/plugins/usj/structureKeyboard.utils.ts` at r1](https://reviewable.io/reviews/eten-tech-foundation/scripture-editors/492#-OwV6MZx5uNNeq4NjUyP:-OwV6MZx5uNNeq4NjUyQ:b3y6z2a) ([raw file](https://github.com/eten-tech-foundation/scripture-editors/blob/706ce60160ecc20637a5403ee601a1196dfd7c92/libs/shared-react/src/plugins/usj/structureKeyboard.utils.ts)):*
Should the functions in this file have JSDoc comments for the parameters and return values?
___
*[`libs/shared-react/src/plugins/usj/structureKeyboard.utils.ts` line 288 at r1](https://reviewable.io/reviews/eten-tech-foundation/scripture-editors/492#-OwV4gNK71J1npxjx8GB:-OwV4gNLD91hKmm2z4os:b-cdtnan) ([raw file](https://github.com/eten-tech-foundation/scripture-editors/blob/706ce60160ecc20637a5403ee601a1196dfd7c92/libs/shared-react/src/plugins/usj/structureKeyboard.utils.ts#L288)):*
> ```XML
>   para.remove();
>   if (junction) $placeCaretAtEnd(junction);
>   else if (moved.length > 0) $placeCaretAtEnd(prev);
> ```

From Devin:

Cursor lands at end instead of merge junction when merging into an empty paragraph

The caret is placed at the end of the merged content (`$placeCaretAtEnd(prev)` at structureKeyboard.utils.ts:288) when the previous paragraph was empty, so the cursor jumps to the wrong position after a two-step paragraph merge.

Impact: After merging a paragraph into an empty preceding paragraph via the two-step delete, the user's cursor unexpectedly appears at the end of the text instead of at the beginning.

Detailed mechanism: $mergeParaIntoPrevious caret placement logic

Suggested fix

```
  else if (moved.length > 0) prev.selectStart();
```
___
*[`libs/shared-react/src/plugins/usj/VerseDeleteTooltip.tsx` line 15 at r1](https://reviewable.io/reviews/eten-tech-foundation/scripture-editors/492#-OwV6hMQ21oDILd3ifmF:-OwV6hMQ21oDILd3ifmG:bum1rn9) ([raw file](https://github.com/eten-tech-foundation/scripture-editors/blob/706ce60160ecc20637a5403ee601a1196dfd7c92/libs/shared-react/src/plugins/usj/VerseDeleteTooltip.tsx#L15)):*
> ```XML
>  * that the jsdom test environment lacks; we re-position on window scroll/resize instead.
>  */
> export function VerseDeleteTooltip({
> ```

JSDoc?
___
*[`libs/shared-react/src/plugins/usj/VerseDeleteTooltip.tsx` line 37 at r1](https://reviewable.io/reviews/eten-tech-foundation/scripture-editors/492#-OwV6kI16Bt7-lr4N_Rd:-OwV6kI16Bt7-lr4N_Re:bhj0y6k) ([raw file](https://github.com/eten-tech-foundation/scripture-editors/blob/706ce60160ecc20637a5403ee601a1196dfd7c92/libs/shared-react/src/plugins/usj/VerseDeleteTooltip.tsx#L37)):*
> ```XML
>         middleware: [flip(), shift()],
>       })
>         .then((pos) => {
> ```

I don't know what the style guide for this repo is but should we try to use `try/catch` notation here instead of `.then()` and `.catch()`.

katherinejensen00 (changes requested):
Jolie, thanks for making it clearer what might get deleted when a verse marker could be impacted! I just added a few nitpicky changes to consider. Levi did a great job with documentation related reviews that are also worth looking at. I made one change blocking just to make sure you have a chance to look at all of these before it merges.

> @katherinejensen00 partially reviewed 17 files and all commit messages, and made 4 comments.
> *[Reviewable](https://reviewable.io/reviews/eten-tech-foundation/scripture-editors/492#-:-OwWTcwO5Ox1uUX-suw9:b-d2ft6)* status: all files reviewed, 10 unresolved discussions (waiting on jolierabideau).

___
*[`libs/shared-react/src/nodes/usj/ImmutableVerseNode.tsx` line 44 at r1](https://reviewable.io/reviews/eten-tech-foundation/scripture-editors/492#-OwWSk2wDua519rriJ7Q:-OwWSk2wDua519rriJ7R:b6tzua7) ([raw file](https://github.com/eten-tech-foundation/scripture-editors/blob/706ce60160ecc20637a5403ee601a1196dfd7c92/libs/shared-react/src/nodes/usj/ImmutableVerseNode.tsx#L44)):*
> ```XML
>  */
> function VerseDecorator({ nodeKey, text }: { nodeKey: NodeKey; text: string }): ReactElement {
>   const [isSelected] = useLexicalNodeSelection(nodeKey);
> ```

NIT Claude Finding (test hygeine)

**This new selection subscription makes the test run print a bunch of `not wrapped in act(...)` warnings.** They're harmless (tests still pass), but they're new with this PR and worth silencing so the suite stays clean.

What's happening: `VerseDecorator` now subscribes to selection here via `useLexicalNodeSelection`, so changing the selection re-renders it — which the old plain-`<span>` `decorate()` never did. The tests change selection via `updateSelection`, and that helper runs its update without an `act()` wrapper (unlike `pressKey`, which wraps), so React complains.

Easiest fix: wrap the update inside `updateSelection` (in `libs/shared/src/nodes/usj/test.utils.ts`) — or its call sites in the plugin tests, e.g. `StructureKeyboardPlugin.test.tsx:432` — in `act()`. You've already got the right pattern in your new `ImmutableVerseNode.test.tsx`, which wraps its selection updates in `act()`, so it's just applying that here too.
___
*[`libs/shared-react/src/plugins/usj/VerseDeleteTooltip.tsx` line 60 at r1](https://reviewable.io/reviews/eten-tech-foundation/scripture-editors/492#-OwWRJeFEuAbzEoWQSo2:-OwWRJeFEuAbzEoWQSo3:b-faep5g) ([raw file](https://github.com/eten-tech-foundation/scripture-editors/blob/706ce60160ecc20637a5403ee601a1196dfd7c92/libs/shared-react/src/plugins/usj/VerseDeleteTooltip.tsx#L60)):*
> ```XML
>       coords={coords}
>       className="floating-box verse-delete-tooltip"
>       role="status"
> ```

NIT Claude Finding (Minor accessibility improvement)

Nice addition with the tooltip! Quick heads-up: **a screen reader probably won't reliably read this tooltip aloud when a marker gets armed**, so its accessibility benefit may not actually land yet.

The reason: the tooltip renders through `FloatingBox`, which sets `aria-hidden` on the element until it has finished positioning itself. But positioning happens asynchronously, so for the first moment the tooltip exists it's both a live region (`aria-live="polite"`) _and_ `aria-hidden="true"` — and anything `aria-hidden` is invisible to screen readers. By the time it un-hides, the text is already there, which is the case screen readers are least reliable about announcing.

Small fix if you want it: don't put `aria-hidden` on the live region (gate just the `visibility`/`opacity` on position instead), or keep one always-present region and drop the message into it when arming. Totally fine to roll into the deferred armed-state follow-up.
___
*[`libs/shared-react/src/plugins/usj/structureKeyboard.baseline.test.tsx` line 76 at r1](https://reviewable.io/reviews/eten-tech-foundation/scripture-editors/492#-OwWTG-L5qofpe93S65_:-OwWTG-L5qofpe93S65a:b-87lu0n) ([raw file](https://github.com/eten-tech-foundation/scripture-editors/blob/706ce60160ecc20637a5403ee601a1196dfd7c92/libs/shared-react/src/plugins/usj/structureKeyboard.baseline.test.tsx#L76)):*
> ```XML
> 
> // With the plugin present but not in protected mode (isStructureProtected={false}), the plugin
> // now implements the two-step intentional-delete behavior (PT-4020). Structural Backspace is
> ```

NIT This is super nitpicky and potentially subjective, but it can be nice to include the url to the jira work item so it is easy to navigate to.

irahopkinson (changes requested):
Looking pretty good however there is the problem of localized messages and tooltips. I wrestled for some time with how to deal with all this between this repo and core and in the end realized that we want to keep the editor repo doing editor only things and that means all user-facing things including messages, localizations, and the look-and-feel (including the decision to ShadCN or something else) belongs in core so that all those decisions and their business logic are there also.

> @irahopkinson partially reviewed 15 files and all commit messages, and made 14 comments.
> *[Reviewable](https://reviewable.io/reviews/eten-tech-foundation/scripture-editors/492#-:-Ow_iW0k8Wy9-zmXL43Y:b-2jj9gj)* status: all files reviewed, 21 unresolved discussions (waiting on jolierabideau and katherinejensen00).

___
*[`libs/shared-react/src/plugins/usj/structureKeyboard.baseline.test.tsx` line 76 at r1](https://reviewable.io/reviews/eten-tech-foundation/scripture-editors/492#-OwWTG-L5qofpe93S65_:-Ow_Z4kUA80NSbDqyUTW:bn566v4) ([raw file](https://github.com/eten-tech-foundation/scripture-editors/blob/706ce60160ecc20637a5403ee601a1196dfd7c92/libs/shared-react/src/plugins/usj/structureKeyboard.baseline.test.tsx#L76)):*

NIT This is super nitpicky and potentially subjective, but it can be nice to include the url to the jira work item so it is easy to navigate to.
</blockquote></details>

This is super nitpicky on my part but I don't even like to refer to Jira issues in code since this is open source but Jira is not. Describe what is fixed without reference to Jira (you already do here). We could also change to a different issue tracker in the future and our code would still be littered with references devs can't see.

On the other hand it's fine to reference them in PRs and commit messages which is the point it makes most sense to reference them since a PR usually completes an issue.
___
*[`libs/shared-react/src/plugins/usj/StructureProtectionPlugin.tsx` at r1](https://reviewable.io/reviews/eten-tech-foundation/scripture-editors/492#-OwZvDvECTrxzaE_04CB:-OwZvDvECTrxzaE_04CC:bw2h92u) ([raw file](https://github.com/eten-tech-foundation/scripture-editors/blob/706ce60160ecc20637a5403ee601a1196dfd7c92/libs/shared-react/src/plugins/usj/StructureProtectionPlugin.tsx)):*
FYI it's really hard to see what is changed if you rename and modify in one commit. If the modification isn't small then git will show this (like here) as a delete and add rather than a change. It also means you loose all the git history (including git blame) for the file. That doesn't really matter in this case since the file was only recently added. In the future it would be better to do all the renames in one commit and the modification in subsequent commits. You also have to retain both commits when you merge rather than squashing.
___
*[`libs/shared-react/src/nodes/usj/ImmutableVerseNode.tsx` line 43 at r1](https://reviewable.io/reviews/eten-tech-foundation/scripture-editors/492#-OwZsFkq8cCp--aTIe8e:-OwZsFkq8cCp--aTIe8f:bnq80mj) ([raw file](https://github.com/eten-tech-foundation/scripture-editors/blob/706ce60160ecc20637a5403ee601a1196dfd7c92/libs/shared-react/src/nodes/usj/ImmutableVerseNode.tsx#L43)):*
> ```XML
>  * {@link VERSE_SELECTED_CLASS_NAME}.
>  */
> function VerseDecorator({ nodeKey, text }: { nodeKey: NodeKey; text: string }): ReactElement {
> ```

NIT move this React component definition to below the `ImmutableVerseNode` class definition. That class matches the filename so we want to see that first in the file before the detail of the new component used in the class.
___
*[`libs/shared-react/src/nodes/usj/ImmutableVerseNode.tsx` line 46 at r1](https://reviewable.io/reviews/eten-tech-foundation/scripture-editors/492#-OwZso-W51eIlNyeUjFW:-OwZso-W51eIlNyeUjFX:b-8k03zm) ([raw file](https://github.com/eten-tech-foundation/scripture-editors/blob/706ce60160ecc20637a5403ee601a1196dfd7c92/libs/shared-react/src/nodes/usj/ImmutableVerseNode.tsx#L46)):*
> ```XML
>   const [isSelected] = useLexicalNodeSelection(nodeKey);
>   // ZWSPs stay inside this span so double-click word selection still excludes the number.
>   return <span className={isSelected ? VERSE_SELECTED_CLASS_NAME : undefined}>{text}</span>;
> ```

It will get this CSS class when it's selected from this delete new trigger mechanism but won't it also get this class if it's selected for any other reason? I.e. is it possible to do node selection with the cursor or is that always a `RangeSelection`? Will that cause any issues?

Since I first wrote this I see you now intentionally trigger on a range selection anyway. Doesn't this CSS class only show on a NodeSelection? Is that right?



_Code quote:_
```XML
VERSE_SELECTED_CLASS_NAME
```
___
*[`libs/shared-react/src/nodes/usj/ImmutableVerseNode.test.tsx` line 21 at r1](https://reviewable.io/reviews/eten-tech-foundation/scripture-editors/492#-Ow_9IpG5kCLBzgR-CLT:-Ow_9IpHEPCytV3G8xgN:bboalzp) ([raw file](https://github.com/eten-tech-foundation/scripture-editors/blob/706ce60160ecc20637a5403ee601a1196dfd7c92/libs/shared-react/src/nodes/usj/ImmutableVerseNode.test.tsx#L21)):*
> ```XML
> /** Query the rendered verse marker that carries the "selected" class, if any. */
> function selectedVerseEl(editor: LexicalEditor): Element | null {
>   return editor.getRootElement()?.querySelector(`.${VERSE_SELECTED_CLASS_NAME}`) ?? null;
> ```

NIT this seems redundant since it's a null coalescing operator and `querySelector` returns `Element | null` anyway. Am I missing something? If instead you did `?? undefined` that would allow you to favor undefined over null but isn't necessary in tests.

Also move helper functions like this after the tests since it is detail we don't need to read first.



_Code quote:_
```XML
?? null
```
___
*[`libs/shared-react/src/plugins/usj/armedDeleteMessage.ts` line 7 at r1](https://reviewable.io/reviews/eten-tech-foundation/scripture-editors/492#-Ow_SjXJ6TJa9tdUjjzV:-Ow_SjXJ6TJa9tdUjjzW:b-pgbnty) ([raw file](https://github.com/eten-tech-foundation/scripture-editors/blob/706ce60160ecc20637a5403ee601a1196dfd7c92/libs/shared-react/src/plugins/usj/armedDeleteMessage.ts#L7)):*
> ```XML
>  * of the sentence. `para` is never shown a tooltip, so it maps to the verse-marker wording.
>  */
> export function armedDeleteMessage(
> ```

We can't do messages like this from the editor repo because we only do localization in core. Find a way to trigger this in core.
___
*[`libs/shared-react/src/plugins/usj/StructureKeyboardPlugin.tsx` line 42 at r1](https://reviewable.io/reviews/eten-tech-foundation/scripture-editors/492#-Ow_TlB4C8I0JxIq7w4b:-Ow_TlB4C8I0JxIq7w4c:b-7v5hvb) ([raw file](https://github.com/eten-tech-foundation/scripture-editors/blob/706ce60160ecc20637a5403ee601a1196dfd7c92/libs/shared-react/src/plugins/usj/StructureKeyboardPlugin.tsx#L42)):*
> ```XML
>  * boundaries. When `isStructureProtected`, blocks them (PT-4013) and sanitizes paste/drop payloads
>  * so structural markers (paragraph breaks, verse markers, chapter markers) cannot be inserted —
>  * text, inline character formatting, and notes are kept. When not protected, makes deletion a
> ```

BTW the unprotected state means the 2-step removal will be enabled for Power. Have you thought if that is appropriate? I think it's probably fine but you might want to check with UX (if you haven't already).



_Code quote:_
```XML
When not protected
```
___
*[`libs/shared-react/src/plugins/usj/StructureKeyboardPlugin.tsx` line 105 at r1](https://reviewable.io/reviews/eten-tech-foundation/scripture-editors/492#-Ow_aLCH2KsqinPzjzET:-Ow_aLCH2KsqinPzjzEU:bopwoez) ([raw file](https://github.com/eten-tech-foundation/scripture-editors/blob/706ce60160ecc20637a5403ee601a1196dfd7c92/libs/shared-react/src/plugins/usj/StructureKeyboardPlugin.tsx#L105)):*
> ```XML
>         } else if (armed.kind === "selection") {
>           if ($isRangeSelection(selection)) selection.removeText(); // deletes the whole range, verse included
>         } else if (node && $isElementNode(node)) {
> ```

Should this be `$isSomeParaNode`? We shouldn't merge some other `ElementNode` with either of `SomeParaNode`.

For that matter we should think if it's ok to merge a `ParaNode` with an `ImpliedParaNode` (and vice versa). Now that I have written that I think merging either to either is fine (but not with some other `ElementNode`).



_Code quote:_
```XML
$isElementNode
```
___
*[`libs/shared-react/src/plugins/usj/structureKeyboard.utils.ts` at r1](https://reviewable.io/reviews/eten-tech-foundation/scripture-editors/492#-Ow_XA5R-c3vZYl8vv4x:-Ow_XA5R-c3vZYl8vv4y:btvu84j) ([raw file](https://github.com/eten-tech-foundation/scripture-editors/blob/706ce60160ecc20637a5403ee601a1196dfd7c92/libs/shared-react/src/plugins/usj/structureKeyboard.utils.ts)):*
BTW consider if any of these utils are generic enough to be moved to `node.utils.ts` or `node-react.utils.ts`. Also consider if there is already a utility there that would do what you need here. The `/simplify` command might help with this.
___
*[`libs/shared-react/src/plugins/usj/structureKeyboard.utils.ts` line 172 at r1](https://reviewable.io/reviews/eten-tech-foundation/scripture-editors/492#-Ow_Y7ox5lQKaz6GzyXe:-Ow_Y7oy8YUpEjIMrhga:b-517k78) ([raw file](https://github.com/eten-tech-foundation/scripture-editors/blob/706ce60160ecc20637a5403ee601a1196dfd7c92/libs/shared-react/src/plugins/usj/structureKeyboard.utils.ts#L172)):*
> ```XML
> 
> /** What a structural delete keystroke would act on. */
> export type DeleteTarget =
> ```

BTW it's good practice to declare types (including interfaces) at the top of the file (after imports).

Here and below also.
___
*[`libs/shared-react/src/plugins/usj/structureKeyboard.utils.ts` line 174 at r1](https://reviewable.io/reviews/eten-tech-foundation/scripture-editors/492#-Ow_WDhZ95ppUmfwy9Mc:-Ow_WDhZ95ppUmfwy9Md:bwi5an6) ([raw file](https://github.com/eten-tech-foundation/scripture-editors/blob/706ce60160ecc20637a5403ee601a1196dfd7c92/libs/shared-react/src/plugins/usj/structureKeyboard.utils.ts#L174)):*
> ```XML
> export type DeleteTarget =
>   | { kind: "verse"; node: SomeVerseNode }
>   | { kind: "para"; node: ElementNode };
> ```

NIT seems like this should be more specific like `SomeParaNode`. If you really mean `ElementNode` then the `kind` probably shouldn't be _"para"_.



_Code quote:_
```XML
ElementNode
```
___
*[`libs/shared-react/src/plugins/usj/VerseDeleteTooltip.tsx` line 15 at r1](https://reviewable.io/reviews/eten-tech-foundation/scripture-editors/492#-OwZtwGe8L1Zl7oBVuMe:-OwZtwGf2ThY8XhAnFu0:b-6l5o6q) ([raw file](https://github.com/eten-tech-foundation/scripture-editors/blob/706ce60160ecc20637a5403ee601a1196dfd7c92/libs/shared-react/src/plugins/usj/VerseDeleteTooltip.tsx#L15)):*
> ```XML
>  * that the jsdom test environment lacks; we re-position on window scroll/resize instead.
>  */
> export function VerseDeleteTooltip({
> ```

We shouldn't do tooltips in this repo because we want tooltips (and any other component) to conform to ShadCN that we use in core (or any other library we might change to in the future). By rolling our own here (which AI has a bad propensity to do) we also would have to fix all the edge cases that are already functional in ShadCN. Instead can we add something here that would allow us to trigger a tooltip in core?
___
*[`packages/platform/src/usj-nodes.css` line 2269 at r1](https://reviewable.io/reviews/eten-tech-foundation/scripture-editors/492#-OwZwjpNBNKzWGaw7tYT:-OwZwjpNBNKzWGaw7tYU:blz1lq9) ([raw file](https://github.com/eten-tech-foundation/scripture-editors/blob/706ce60160ecc20637a5403ee601a1196dfd7c92/packages/platform/src/usj-nodes.css#L2269)):*
> ```CSS
> }
> 
> /* Verse marker armed for the two-step intentional delete (first Backspace/Delete selects it).
> ```

These changes belong elsewhere like in `editor.css`. In the future this file for USJ nodes will be generated for each project/resource by core so this change would get lost but we still want these changes to remain so move them. This also means we can use `editor-overrides.scss` in core if we want to change the colors there.

jolierabideau (commented):
Agreed! The editor repo no longer renders user-facing copy/tooltip:

-   Deleted `armedDeleteMessage.ts` and `VerseDeleteTooltip.tsx` (and their tests) from shared-react.
-   `StructureKeyboardPlugin` now publishes the armed state to the editor root as `data-verse-delete-intent` (`deleteBackward`/`deleteForward`) and `data-verse-delete-kind` (`verse`/`selection`); the armed marker already carries `verse-selected`. It renders nothing.
-   In **paranext-core**, new `verse-delete-tooltip/VerseDeleteTooltipOverlay` observes those DOM signals and renders a **ShadCN `Tooltip`** with copy from **`useLocalizedStrings`**, anchored to the marker — built to match the existing `ParagraphMarkerTooltipOverlay` precedent. Wired into `platform-scripture-editor.web-view.tsx` around the editor. New localized keys `..._verseDelete_removeVerseMarker` / `..._verseDelete_deleteSelection` added to `localizedStrings.json` (en + es).

> @jolierabideau made 23 comments and resolved 7 discussions.
> *[Reviewable](https://reviewable.io/reviews/eten-tech-foundation/scripture-editors/492#-:-OwxChMf-T-gPjenAMa1:bmnd53v)* status: all files reviewed, 14 unresolved discussions (waiting on captaincrazybro, irahopkinson, and katherinejensen00).

___
*[a discussion](https://reviewable.io/reviews/eten-tech-foundation/scripture-editors/492#-OwV7SK43O2WVdi-OUyY:-OwxCjnjCGmjgXuNSg9j:b-896fix) (no related file):*
<details><summary><i>Previously, captaincrazybro (Cqptain) wrote…</i></summary><blockquote>

Just had a few mostly documentation and style comments with one potential bug reported by Devin.
Probably should have Ira look at this as I don't have access to actually approve.
</blockquote></details>

Done.
___
*[`libs/shared-react/src/nodes/usj/ImmutableVerseNode.tsx` line 43 at r1](https://reviewable.io/reviews/eten-tech-foundation/scripture-editors/492#-OwZsFkq8cCp--aTIe8e:-OwxEz1j0nYlWjbf5vvG:bjphk79) ([raw file](https://github.com/eten-tech-foundation/scripture-editors/blob/706ce60160ecc20637a5403ee601a1196dfd7c92/libs/shared-react/src/nodes/usj/ImmutableVerseNode.tsx#L43)):*

NIT move this React component definition to below the `ImmutableVerseNode` class definition. That class matches the filename so we want to see that first in the file before the detail of the new component used in the class.
</blockquote></details>

Done. `VerseDecorator` now sits after the `ImmutableVerseNode` class definition.
___
*[`libs/shared-react/src/nodes/usj/ImmutableVerseNode.tsx` line 44 at r1](https://reviewable.io/reviews/eten-tech-foundation/scripture-editors/492#-OwWSk2wDua519rriJ7Q:-OwxENwm-AUdEQ7V5RWL:bg6x3i7) ([raw file](https://github.com/eten-tech-foundation/scripture-editors/blob/706ce60160ecc20637a5403ee601a1196dfd7c92/libs/shared-react/src/nodes/usj/ImmutableVerseNode.tsx#L44)):*

NIT Claude Finding (test hygeine)

**This new selection subscription makes the test run print a bunch of `not wrapped in act(...)` warnings.** They're harmless (tests still pass), but they're new with this PR and worth silencing so the suite stays clean.

What's happening: `VerseDecorator` now subscribes to selection here via `useLexicalNodeSelection`, so changing the selection re-renders it — which the old plain-`<span>` `decorate()` never did. The tests change selection via `updateSelection`, and that helper runs its update without an `act()` wrapper (unlike `pressKey`, which wraps), so React complains.

Easiest fix: wrap the update inside `updateSelection` (in `libs/shared/src/nodes/usj/test.utils.ts`) — or its call sites in the plugin tests, e.g. `StructureKeyboardPlugin.test.tsx:432` — in `act()`. You've already got the right pattern in your new `ImmutableVerseNode.test.tsx`, which wraps its selection updates in `act()`, so it's just applying that here too.
</blockquote></details>

Done Added a React-aware `updateSelection` wrapper in `react-test.utils.tsx` that runs the selection change inside `act(...)`, and repointed the component test files to it (the shared `updateSelection` stays framework-agnostic so `libs/shared` keeps no React-testing dependency). Two raw `editor.update` selection calls in `structureKeyboard.utils.test.ts` were routed through the existing `sutUpdate` helper.
___
*[`libs/shared-react/src/nodes/usj/ImmutableVerseNode.tsx` line 46 at r1](https://reviewable.io/reviews/eten-tech-foundation/scripture-editors/492#-OwZso-W51eIlNyeUjFW:-OwxFFx9CWVnmMySD7Uf:b-wge1xn) ([raw file](https://github.com/eten-tech-foundation/scripture-editors/blob/706ce60160ecc20637a5403ee601a1196dfd7c92/libs/shared-react/src/nodes/usj/ImmutableVerseNode.tsx#L46)):*

It will get this CSS class when it's selected from this delete new trigger mechanism but won't it also get this class if it's selected for any other reason? I.e. is it possible to do node selection with the cursor or is that always a `RangeSelection`? Will that cause any issues?

Since I first wrote this I see you now intentionally trigger on a range selection anyway. Doesn't this CSS class only show on a NodeSelection? Is that right?
</blockquote></details>

`verse-selected` reflects whether the node is selected, so it shows both when armed (a `NodeSelection` on the verse) and when the verse merely falls inside an ordinary `RangeSelection`. A verse inside a selection gets the subtle destructive tint, but the caret-like **blink** is gated on the `verse-delete-armed` root class, so only an actually-armed marker blinks. So: tint = "this verse is in the selection", blink = "armed, press again to delete".
___
*[`libs/shared-react/src/nodes/usj/ImmutableVerseNode.test.tsx` line 21 at r1](https://reviewable.io/reviews/eten-tech-foundation/scripture-editors/492#-Ow_9IpG5kCLBzgR-CLT:-OwxFQFaEvG4kz2szJkA:bukfd4s) ([raw file](https://github.com/eten-tech-foundation/scripture-editors/blob/706ce60160ecc20637a5403ee601a1196dfd7c92/libs/shared-react/src/nodes/usj/ImmutableVerseNode.test.tsx#L21)):*

NIT this seems redundant since it's a null coalescing operator and `querySelector` returns `Element | null` anyway. Am I missing something? If instead you did `?? undefined` that would allow you to favor undefined over null but isn't necessary in tests.

Also move helper functions like this after the tests since it is detail we don't need to read first.
</blockquote></details>

Done. Dopped the `?? null` and moved the helper function.
___
*[`libs/shared-react/src/plugins/usj/armedDeleteMessage.ts` line 7 at r1](https://reviewable.io/reviews/eten-tech-foundation/scripture-editors/492#-Ow_SjXJ6TJa9tdUjjzV:-OwxFYRZ75BW6mwixk_p:b-896fix) ([raw file](https://github.com/eten-tech-foundation/scripture-editors/blob/706ce60160ecc20637a5403ee601a1196dfd7c92/libs/shared-react/src/plugins/usj/armedDeleteMessage.ts#L7)):*

We can't do messages like this from the editor repo because we only do localization in core. Find a way to trigger this in core.
</blockquote></details>

Done.
___
*[`libs/shared-react/src/plugins/usj/armedDeleteMessage.ts` line 10 at r1](https://reviewable.io/reviews/eten-tech-foundation/scripture-editors/492#-OwV4ojNAn-FSQDHK32R:-OwxDYhu0XjW_smb2KJd:b1d3tvl) ([raw file](https://github.com/eten-tech-foundation/scripture-editors/blob/706ce60160ecc20637a5403ee601a1196dfd7c92/libs/shared-react/src/plugins/usj/armedDeleteMessage.ts#L10)):*
<details><summary><i>Previously, captaincrazybro (Cqptain) wrote…</i></summary><blockquote>

Add JSDoc comments for parameters? And maybe return JSDoc?
</blockquote></details>

Done Removed this to handle it in core
___
*[`libs/shared-react/src/plugins/usj/StructureKeyboardPlugin.tsx` line 42 at r1](https://reviewable.io/reviews/eten-tech-foundation/scripture-editors/492#-Ow_TlB4C8I0JxIq7w4b:-OwxFcfN8tgkjdgJObVs:b-1bbtl9) ([raw file](https://github.com/eten-tech-foundation/scripture-editors/blob/706ce60160ecc20637a5403ee601a1196dfd7c92/libs/shared-react/src/plugins/usj/StructureKeyboardPlugin.tsx#L42)):*

BTW the unprotected state means the 2-step removal will be enabled for Power. Have you thought if that is appropriate? I think it's probably fine but you might want to check with UX (if you haven't already).
</blockquote></details>

Done. Scoped it to simple only
___
*[`libs/shared-react/src/plugins/usj/StructureKeyboardPlugin.tsx` line 48 at r1](https://reviewable.io/reviews/eten-tech-foundation/scripture-editors/492#-OwV56tOCeXwc8akE7Zk:-OwxDgFX0_YMgGY2OEx3:bh9ssf0) ([raw file](https://github.com/eten-tech-foundation/scripture-editors/blob/706ce60160ecc20637a5403ee601a1196dfd7c92/libs/shared-react/src/plugins/usj/StructureKeyboardPlugin.tsx#L48)):*
<details><summary><i>Previously, captaincrazybro (Cqptain) wrote…</i></summary><blockquote>

This JSDoc comment doesn't seem to be correct. Seems like it does sometimes return JSX.
</blockquote></details>

Done Good catch, it did return `<VerseDeleteTooltip>`. With the tooltip moved to core the component now genuinely always returns `null`, so the return type is `null` and the JSDoc is accurate. I also expanded the JSDoc to document the DOM signals it publishes.
___
*[`libs/shared-react/src/plugins/usj/StructureKeyboardPlugin.tsx` line 105 at r1](https://reviewable.io/reviews/eten-tech-foundation/scripture-editors/492#-Ow_aLCH2KsqinPzjzET:-OwxFnSC3tw0aQbZ7I2H:b-5kr2nt) ([raw file](https://github.com/eten-tech-foundation/scripture-editors/blob/706ce60160ecc20637a5403ee601a1196dfd7c92/libs/shared-react/src/plugins/usj/StructureKeyboardPlugin.tsx#L105)):*

Should this be `$isSomeParaNode`? We shouldn't merge some other `ElementNode` with either of `SomeParaNode`.

For that matter we should think if it's ok to merge a `ParaNode` with an `ImpliedParaNode` (and vice versa). Now that I have written that I think merging either to either is fine (but not with some other `ElementNode`).
</blockquote></details>

Done. The guard is now `$isSomeParaNode(node)`, and `$mergeParaIntoPrevious` takes a `SomeParaNode` and also checks its previous sibling with `$isSomeParaNode` — so only paragraphs merge into paragraphs.
___
*[`libs/shared-react/src/plugins/usj/StructureProtectionPlugin.tsx` at r1](https://reviewable.io/reviews/eten-tech-foundation/scripture-editors/492#-OwZvDvECTrxzaE_04CB:-OwxEsUL0h3VGMThu3cI:b-hb5bdk) ([raw file](https://github.com/eten-tech-foundation/scripture-editors/blob/706ce60160ecc20637a5403ee601a1196dfd7c92/libs/shared-react/src/plugins/usj/StructureProtectionPlugin.tsx)):*

FYI it's really hard to see what is changed if you rename and modify in one commit. If the modification isn't small then git will show this (like here) as a delete and add rather than a change. It also means you loose all the git history (including git blame) for the file. That doesn't really matter in this case since the file was only recently added. In the future it would be better to do all the renames in one commit and the modification in subsequent commits. You also have to retain both commits when you merge rather than squashing.
</blockquote></details>

That is a great point, thanks!
___
*[`libs/shared-react/src/plugins/usj/structureKeyboard.utils.ts` line 172 at r1](https://reviewable.io/reviews/eten-tech-foundation/scripture-editors/492#-Ow_Y7ox5lQKaz6GzyXe:-OwxG8mz0E8a7eAIAAYh:bs7wumx) ([raw file](https://github.com/eten-tech-foundation/scripture-editors/blob/706ce60160ecc20637a5403ee601a1196dfd7c92/libs/shared-react/src/plugins/usj/structureKeyboard.utils.ts#L172)):*

BTW it's good practice to declare types (including interfaces) at the top of the file (after imports).

Here and below also.
</blockquote></details>

Done. 
___
*[`libs/shared-react/src/plugins/usj/structureKeyboard.utils.ts` line 174 at r1](https://reviewable.io/reviews/eten-tech-foundation/scripture-editors/492#-Ow_WDhZ95ppUmfwy9Mc:-OwxFu9lBsOrP5mfkAY6:b-k3936v) ([raw file](https://github.com/eten-tech-foundation/scripture-editors/blob/706ce60160ecc20637a5403ee601a1196dfd7c92/libs/shared-react/src/plugins/usj/structureKeyboard.utils.ts#L174)):*

NIT seems like this should be more specific like `SomeParaNode`. If you really mean `ElementNode` then the `kind` probably shouldn't be _"para"_.
</blockquote></details>

Done. `DeleteTarget`'s para variant is now `{ kind: "para"; node: SomeParaNode }`, and `$structuralDeleteTarget` resolves it with `$isSomeParaNode`.
___
*[`libs/shared-react/src/plugins/usj/structureKeyboard.utils.ts` line 288 at r1](https://reviewable.io/reviews/eten-tech-foundation/scripture-editors/492#-OwV4gNK71J1npxjx8GB:-OwxE7etAjX92ZxuE1kl:b-2vv1nh) ([raw file](https://github.com/eten-tech-foundation/scripture-editors/blob/706ce60160ecc20637a5403ee601a1196dfd7c92/libs/shared-react/src/plugins/usj/structureKeyboard.utils.ts#L288)):*
<details><summary><i>Previously, captaincrazybro (Cqptain) wrote…</i></summary><blockquote>

From Devin:

Cursor lands at end instead of merge junction when merging into an empty paragraph

The caret is placed at the end of the merged content (`$placeCaretAtEnd(prev)` at structureKeyboard.utils.ts:288) when the previous paragraph was empty, so the cursor jumps to the wrong position after a two-step paragraph merge.

Impact: After merging a paragraph into an empty preceding paragraph via the two-step delete, the user's cursor unexpectedly appears at the end of the text instead of at the beginning.

Detailed mechanism: $mergeParaIntoPrevious caret placement logic

Suggested fix

```
  else if (moved.length > 0) prev.selectStart();
```
</blockquote></details>

Done Confirmed the bug: when the previous paragraph was empty (`junction` undefined), `$placeCaretAtEnd(prev)` put the caret at the end of the moved content. The logic now collapses to `if (junction) $placeCaretAtEnd(junction); else prev.selectStart();`, so the caret lands at the junction in both cases.
___
*[`libs/shared-react/src/plugins/usj/structureKeyboard.utils.ts` at r1](https://reviewable.io/reviews/eten-tech-foundation/scripture-editors/492#-OwV6MZx5uNNeq4NjUyP:-OwxE0fb-tIFj4opk1xY:b-896fix) ([raw file](https://github.com/eten-tech-foundation/scripture-editors/blob/706ce60160ecc20637a5403ee601a1196dfd7c92/libs/shared-react/src/plugins/usj/structureKeyboard.utils.ts)):*
<details><summary><i>Previously, captaincrazybro (Cqptain) wrote…</i></summary><blockquote>

Should the functions in this file have JSDoc comments for the parameters and return values?
</blockquote></details>

Done.
___
*[`libs/shared-react/src/plugins/usj/structureKeyboard.utils.ts` at r1](https://reviewable.io/reviews/eten-tech-foundation/scripture-editors/492#-Ow_XA5R-c3vZYl8vv4x:-OwxHtiZ0Ffe-yv-YwNA:b-rsqj7z) ([raw file](https://github.com/eten-tech-foundation/scripture-editors/blob/706ce60160ecc20637a5403ee601a1196dfd7c92/libs/shared-react/src/plugins/usj/structureKeyboard.utils.ts)):*

BTW consider if any of these utils are generic enough to be moved to `node.utils.ts` or `node-react.utils.ts`. Also consider if there is already a utility there that would do what you need here. The `/simplify` command might help with this.
</blockquote></details>

I do not believe they are generic enough to move, and I could not find any in the other files that would be useful. `/simplify` confirmed that but did:
1.  **Removed dead `verseKeys` field** from `ArmedDelete` — it was only ever read by the now-deleted `VerseDeleteTooltip`. Dropped the field from the type, stopped populating it in [StructureKeyboardPlugin.tsx](vscode-webview://0oqmghua87getfiijg7t4honkp5tbk6k01u8478fm49o1l6q4ig9/libs/shared-react/src/plugins/usj/StructureKeyboardPlugin.tsx#L149), and removed it from two test fixtures. (`anchor`/`focus` are _not_ dead — `$isArmedSelection` still reads them — so they stayed.)
2.  **Fixed stale JSDoc** on `ArmedDelete.key` that still referenced "the tooltip anchor."
3.  **Restored the tighter `Element | null` return type** on the `selectedVerseEl` test helper (the diff had widened it to include `undefined`).
___
*[`libs/shared-react/src/plugins/usj/VerseDeleteTooltip.tsx` line 15 at r1](https://reviewable.io/reviews/eten-tech-foundation/scripture-editors/492#-OwV6hMQ21oDILd3ifmF:-OwxE9uF67wi2BTtg99a:b-5boy2z) ([raw file](https://github.com/eten-tech-foundation/scripture-editors/blob/706ce60160ecc20637a5403ee601a1196dfd7c92/libs/shared-react/src/plugins/usj/VerseDeleteTooltip.tsx#L15)):*
<details><summary><i>Previously, captaincrazybro (Cqptain) wrote…</i></summary><blockquote>

JSDoc?
</blockquote></details>

Done. File removed
___
*[`libs/shared-react/src/plugins/usj/VerseDeleteTooltip.tsx` line 15 at r1](https://reviewable.io/reviews/eten-tech-foundation/scripture-editors/492#-OwZtwGe8L1Zl7oBVuMe:-OwxGisa8AKX85syqwZT:b-896fix) ([raw file](https://github.com/eten-tech-foundation/scripture-editors/blob/706ce60160ecc20637a5403ee601a1196dfd7c92/libs/shared-react/src/plugins/usj/VerseDeleteTooltip.tsx#L15)):*

We shouldn't do tooltips in this repo because we want tooltips (and any other component) to conform to ShadCN that we use in core (or any other library we might change to in the future). By rolling our own here (which AI has a bad propensity to do) we also would have to fix all the edge cases that are already functional in ShadCN. Instead can we add something here that would allow us to trigger a tooltip in core?
</blockquote></details>

Done.
___
*[`libs/shared-react/src/plugins/usj/VerseDeleteTooltip.tsx` line 37 at r1](https://reviewable.io/reviews/eten-tech-foundation/scripture-editors/492#-OwV6kI16Bt7-lr4N_Rd:-OwxECUhA2mUWDK18yCp:b-5boy2z) ([raw file](https://github.com/eten-tech-foundation/scripture-editors/blob/706ce60160ecc20637a5403ee601a1196dfd7c92/libs/shared-react/src/plugins/usj/VerseDeleteTooltip.tsx#L37)):*
<details><summary><i>Previously, captaincrazybro (Cqptain) wrote…</i></summary><blockquote>

I don't know what the style guide for this repo is but should we try to use `try/catch` notation here instead of `.then()` and `.catch()`.
</blockquote></details>

Done. File removed
___
*[`libs/shared-react/src/plugins/usj/VerseDeleteTooltip.tsx` line 60 at r1](https://reviewable.io/reviews/eten-tech-foundation/scripture-editors/492#-OwWRJeFEuAbzEoWQSo2:-OwxESkv3oOQ8tSPGdBo:b-9d8is5) ([raw file](https://github.com/eten-tech-foundation/scripture-editors/blob/706ce60160ecc20637a5403ee601a1196dfd7c92/libs/shared-react/src/plugins/usj/VerseDeleteTooltip.tsx#L60)):*

NIT Claude Finding (Minor accessibility improvement)

Nice addition with the tooltip! Quick heads-up: **a screen reader probably won't reliably read this tooltip aloud when a marker gets armed**, so its accessibility benefit may not actually land yet.

The reason: the tooltip renders through `FloatingBox`, which sets `aria-hidden` on the element until it has finished positioning itself. But positioning happens asynchronously, so for the first moment the tooltip exists it's both a live region (`aria-live="polite"`) _and_ `aria-hidden="true"` — and anything `aria-hidden` is invisible to screen readers. By the time it un-hides, the text is already there, which is the case screen readers are least reliable about announcing.

Small fix if you want it: don't put `aria-hidden` on the live region (gate just the `visibility`/`opacity` on position instead), or keep one always-present region and drop the message into it when arming. Totally fine to roll into the deferred armed-state follow-up.
</blockquote></details>

Done. Fixed by migrating this to core. The hint is now a vetted ShadCN `Tooltip` in core rather than our own `FloatingBox` live region, so we no longer own the `aria-hidden`\-while-positioning edge case. If we want guaranteed announcement of the armed state we can layer a dedicated live region in the core overlay as a follow-up — flagging for the deferred armed-state work you mentioned.
___
*[`packages/platform/src/usj-nodes.css` line 2269 at r1](https://reviewable.io/reviews/eten-tech-foundation/scripture-editors/492#-OwZwjpNBNKzWGaw7tYT:-OwxGgez7DPSsQVgDDJZ:bq863cw) ([raw file](https://github.com/eten-tech-foundation/scripture-editors/blob/706ce60160ecc20637a5403ee601a1196dfd7c92/packages/platform/src/usj-nodes.css#L2269)):*

These changes belong elsewhere like in `editor.css`. In the future this file for USJ nodes will be generated for each project/resource by core so this change would get lost but we still want these changes to remain so move them. This also means we can use `editor-overrides.scss` in core if we want to change the colors there.
</blockquote></details>

Done. Moved the marker styling (`verse-selected`, the armed blink, reduced-motion) to `packages/platform/src/editor/editor.css`.
___
*[`libs/shared-react/src/plugins/usj/structureKeyboard.baseline.test.tsx` line 76 at r1](https://reviewable.io/reviews/eten-tech-foundation/scripture-editors/492#-OwWTG-L5qofpe93S65_:-OwxEaG58cdDQGzJmVN_:b-s8smyz) ([raw file](https://github.com/eten-tech-foundation/scripture-editors/blob/706ce60160ecc20637a5403ee601a1196dfd7c92/libs/shared-react/src/plugins/usj/structureKeyboard.baseline.test.tsx#L76)):*

This is super nitpicky on my part but I don't even like to refer to Jira issues in code since this is open source but Jira is not. Describe what is fixed without reference to Jira (you already do here). We could also change to a different issue tracker in the future and our code would still be littered with references devs can't see.

On the other hand it's fine to reference them in PRs and commit messages which is the point it makes most sense to reference them since a PR usually completes an issue.
</blockquote></details>

Done. Removed Jira reference from code

irahopkinson (changes requested):
Thanks for fixing this all up. Just one blocker (notified in 2 places).

> @irahopkinson partially reviewed 19 files and all commit messages, made 5 comments, and resolved 6 discussions.
> *[Reviewable](https://reviewable.io/reviews/eten-tech-foundation/scripture-editors/492#-:-Ox8uYtz7EvV4PHHrk9B:bhwb0cq)* status: all files reviewed, 11 unresolved discussions (waiting on captaincrazybro, jolierabideau, and katherinejensen00).

___
*[`libs/shared-react/src/nodes/usj/ImmutableVerseNode.test.tsx` line 21 at r1](https://reviewable.io/reviews/eten-tech-foundation/scripture-editors/492#-Ow_9IpG5kCLBzgR-CLT:-Ox8mWmq70SdGzJ-EKQe:bf85i77) ([raw file](https://github.com/eten-tech-foundation/scripture-editors/blob/706ce60160ecc20637a5403ee601a1196dfd7c92/libs/shared-react/src/nodes/usj/ImmutableVerseNode.test.tsx#L21)):*

Done. Dopped the `?? null` and moved the helper function.
</blockquote></details>

err, it still has the `?? null`.
___
*[`libs/shared-react/src/plugins/usj/react-test.utils.tsx` line 9 at r2](https://reviewable.io/reviews/eten-tech-foundation/scripture-editors/492#-Ox8nPxc0QZ79ZSTsaOn:-Ox8nPxc0QZ79ZSTsaOo:bwzwlj2) ([raw file](https://github.com/eten-tech-foundation/scripture-editors/blob/ef3b44cdcffa2dc0b9eda355173911c4cfdfbedb/libs/shared-react/src/plugins/usj/react-test.utils.tsx#L9)):*
> ```XML
> } from "../../../../../packages/utilities/src/converters/usj/converter-test.data";
> // eslint-disable-next-line @nx/enforce-module-boundaries
> import { updateSelection as $updateSelection } from "../../../../../libs/shared/src/nodes/usj/test.utils";
> ```

NIT `$` is special in Lexical signifying that this function must be called in an `editor.update()` or `editor.read()` so use something else here for the local name. A common pattern we use in core would be to rename `$updateSelection` to `updateSelectionInternal` or `internalUpdateSelection`.

Consider if the original function in `test.utils.ts` could be internally wrapped in `act()` and would that work every where it's already called.



_Code quote:_
```XML
$updateSelection
```
___
*[`packages/platform/src/editor/Editor.tsx` line 145 at r2](https://reviewable.io/reviews/eten-tech-foundation/scripture-editors/492#-Ox8r_mFEoHZ_0tbZ7wX:-Ox8r_mFEoHZ_0tbZ7wY:bio16wi) ([raw file](https://github.com/eten-tech-foundation/scripture-editors/blob/ef3b44cdcffa2dc0b9eda355173911c4cfdfbedb/packages/platform/src/editor/Editor.tsx#L145)):*
> ```XML
>     isReadonly = false,
>     isStructureProtected = false,
>     isStructureProtectionActive = true,
> ```

Shouldn't this default to false so it works for Power and Simple explicitly sets it true?
___
*[`packages/platform/src/editor/editor.model.ts` line 197 at r2](https://reviewable.io/reviews/eten-tech-foundation/scripture-editors/492#-Ox8ryFt4TlVsq-RUFYl:-Ox8ryFt4TlVsq-RUFYm:b-58vkem) ([raw file](https://github.com/eten-tech-foundation/scripture-editors/blob/ef3b44cdcffa2dc0b9eda355173911c4cfdfbedb/packages/platform/src/editor/editor.model.ts#L197)):*
> ```XML
>    * Whether the structure-protection feature applies at all. When false (e.g. Power interface
>    * mode), the structural keyboard handling — both protection-blocking and the two-step
>    * intentional delete — is disabled and editing is fully native. Defaults to active (true).
> ```

I think this should default to false (i.e. Power).

irahopkinson (changes requested):
> @irahopkinson made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/eten-tech-foundation/scripture-editors/492)* status: all files reviewed, 12 unresolved discussions (waiting on captaincrazybro, jolierabideau, and katherinejensen00).

___
*[`packages/platform/src/editor/editor.model.ts` line 193 at r2](https://reviewable.io/reviews/eten-tech-foundation/scripture-editors/492#-OxB-rlm39mekd7jnnnN:-OxB-rlm39mekd7jnnnO:b-sfwn0q) ([raw file](https://github.com/eten-tech-foundation/scripture-editors/blob/ef3b44cdcffa2dc0b9eda355173911c4cfdfbedb/packages/platform/src/editor/editor.model.ts#L193)):*
> ```XML
>   isReadonly?: boolean;
>   /** When true, paragraph and verse markers cannot be changed via keyboard input. */
>   isStructureProtected?: boolean;
> ```

Instead of having 2 booleans how about one tri-state property, i.e. `structureProtectionMode: 'off' | 'protected' | 'protectedAnd2StepRemove';` - you can probably think of better state names.



_Code quote:_
```XML
isStructureProtected
```

captaincrazybro (commented):
:lgtm:

> @captaincrazybro made 1 comment and resolved 6 discussions.
> *[Reviewable](https://reviewable.io/reviews/eten-tech-foundation/scripture-editors/492#-:-OxcsM-P0wVXQHqy4FIJ:bnfp4nl)* status: all files reviewed, 6 unresolved discussions (waiting on jolierabideau and katherinejensen00).

captaincrazybro (commented):
:lgtm:

> @captaincrazybro partially reviewed 19 files and all commit messages, and made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/eten-tech-foundation/scripture-editors/492#-:-OxcsQEe2uM5RRePN_im:bnfp4nl)* status: all files reviewed, 6 unresolved discussions (waiting on jolierabideau and katherinejensen00).

captaincrazybro (commented):
:lgtm:

> @captaincrazybro made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/eten-tech-foundation/scripture-editors/492#-:-OxctJJ5942u_mjrDmW2:bnfp4nl)* status: all files reviewed, 6 unresolved discussions (waiting on jolierabideau and katherinejensen00).

captaincrazybro (commented):
> @captaincrazybro resolved 1 discussion.
> *[Reviewable](https://reviewable.io/reviews/eten-tech-foundation/scripture-editors/492)* status: all files reviewed, 5 unresolved discussions (waiting on jolierabideau and katherinejensen00).

tombogle:
*[`packages/platform/src/editor/editor.model.ts` line 193 at r2](https://reviewable.io/reviews/eten-tech-foundation/scripture-editors/492#-OxB-rlm39mekd7jnnnN:-OxgkWdfAvsJ_MCVU5mj:b-3eah8d) ([raw file](https://github.com/eten-tech-foundation/scripture-editors/blob/ef3b44cdcffa2dc0b9eda355173911c4cfdfbedb/packages/platform/src/editor/editor.model.ts#L193)):*

Instead of having 2 booleans how about one tri-state property, i.e. `structureProtectionMode: 'off' | 'protected' | 'protectedAnd2StepRemove';` - you can probably think of better state names.
</blockquote></details>

Agreed. I propose: 'off' | 'guarded' | 'protected'

katherinejensen00 (dismissed):
> @katherinejensen00 partially reviewed 19 files and all commit messages, and resolved 1 discussion.
> *[Reviewable](https://reviewable.io/reviews/eten-tech-foundation/scripture-editors/492)* status: all files reviewed, 4 unresolved discussions (waiting on irahopkinson and jolierabideau).

tombogle (commented):
> @tombogle made 2 comments.
> *[Reviewable](https://reviewable.io/reviews/eten-tech-foundation/scripture-editors/492)* status: 14 of 24 files reviewed, 4 unresolved discussions (waiting on captaincrazybro, irahopkinson, jolierabideau, and katherinejensen00).

___
*[`libs/shared-react/src/plugins/usj/react-test.utils.tsx` line 9 at r2](https://reviewable.io/reviews/eten-tech-foundation/scripture-editors/492#-Ox8nPxc0QZ79ZSTsaOn:-OxgIXAj8ikDOkEeZjf5:b-bjutkk) ([raw file](https://github.com/eten-tech-foundation/scripture-editors/blob/ef3b44cdcffa2dc0b9eda355173911c4cfdfbedb/libs/shared-react/src/plugins/usj/react-test.utils.tsx#L9)):*

NIT `$` is special in Lexical signifying that this function must be called in an `editor.update()` or `editor.read()` so use something else here for the local name. A common pattern we use in core would be to rename `$updateSelection` to `updateSelectionInternal` or `internalUpdateSelection`.

Consider if the original function in `test.utils.ts` could be internally wrapped in `act()` and would that work every where it's already called.
</blockquote></details>

Fixed: renamed the import alias from `$updateSelection` to `updateSelectionInternal` (react-test.utils.tsx:9,90,92) since it's not itself a `$`\-function — it takes `editor` and calls `editor.update()` internally, so the `$` prefix was misleading.

On moving `act()` into the shared `test.utils.ts` `updateSelection`: that wouldn't be a good idea. `annotation/selection.utils.test.ts` calls the shared `updateSelection` directly via `createBasicTestEnvironment` with no React rendering at all (no `LexicalComposer`, no `render()`). Baking `act()` (from `@testing-library/react`) into `libs/shared` would pull a React-specific testing dependency into the framework-agnostic package, breaking the `shared` → `shared-react` layering this repo relies on. Keeping the thin act()-wrapping shim in `react-test.utils.tsx` for React-rendering tests, while non-React tests call the shared version directly, is the correct split — so I just fixed the naming nit rather than merging the two.
___
*[`packages/platform/src/editor/Editor.tsx` line 145 at r2](https://reviewable.io/reviews/eten-tech-foundation/scripture-editors/492#-Ox8r_mFEoHZ_0tbZ7wX:-OxgSelOBzuWxZuD6XJM:bbpfdzo) ([raw file](https://github.com/eten-tech-foundation/scripture-editors/blob/ef3b44cdcffa2dc0b9eda355173911c4cfdfbedb/packages/platform/src/editor/Editor.tsx#L145)):*

Shouldn't this default to false so it works for Power and Simple explicitly sets it true?
</blockquote></details>

Agreed (as we discussed in standup meeting). keeping the existing default behavior (for Power) is probably the best way. Note: Corresponding PR in paranext-core will need to set this for Simple and leave the default alone for Power.

katherinejensen00 (dismissed):
:lgtm:

> @katherinejensen00 partially reviewed 10 files and all commit messages, and made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/eten-tech-foundation/scripture-editors/492#-:-Oxl2jdk6DIzhLVkiRd-:bnfp4nl)* status: all files reviewed, 4 unresolved discussions (waiting on irahopkinson and jolierabideau).

irahopkinson (dismissed):
:lgtm: with a couple of NITs marked as FYIs so they really aren't blocking.

> @irahopkinson partially reviewed 10 files and all commit messages, made 3 comments, and resolved 4 discussions.
> *[Reviewable](https://reviewable.io/reviews/eten-tech-foundation/scripture-editors/492#-:-OxlifDg80wo4svUR-ge:b-z7gc51)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved.

___
*[`libs/shared-react/src/plugins/usj/StructureKeyboardPlugin.test.tsx` line 702 at r4](https://reviewable.io/reviews/eten-tech-foundation/scripture-editors/492#-Oxli92QBcacEEcrwk0e:-Oxli92R0QsH75VQjU-W:ba74dn2) ([raw file](https://github.com/eten-tech-foundation/scripture-editors/blob/e020122d41005f263fd9eee47c7fc3450d0d2323/libs/shared-react/src/plugins/usj/StructureKeyboardPlugin.test.tsx#L702)):*
> ```XML
> });
> 
> async function testEnvironment($initialEditorState: () => void) {
> ```

FYI rename `protectedEnvironment`



_Code quote:_
```XML
testEnvironment
```
___
*[`libs/shared-react/src/plugins/usj/StructureKeyboardPlugin.test.tsx` line 709 at r4](https://reviewable.io/reviews/eten-tech-foundation/scripture-editors/492#-OxliHw10TRfstSheQGo:-OxliHw10TRfstSheQGp:byv2i3o) ([raw file](https://github.com/eten-tech-foundation/scripture-editors/blob/e020122d41005f263fd9eee47c7fc3450d0d2323/libs/shared-react/src/plugins/usj/StructureKeyboardPlugin.test.tsx#L709)):*
> ```XML
> }
> 
> async function unprotectedEnvironment($initialEditorState: () => void) {
> ```

FYI rename `guardedEnvironment`



_Code quote:_
```XML
unprotectedEnvironment
```

irahopkinson (approved):
:lgtm:

> @irahopkinson reviewed 1 file and all commit messages, and made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/eten-tech-foundation/scripture-editors/492#-:-Oxm1XusAejmVK92wJQg:bnfp4nl)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved.

PR #494, opened 2026-07-01
PT-3086: Render USJ tables in the platform editor

katherinejensen00:
## Why

USJ table elements (`table` / `table:row` / `table:cell`) fell through the platform forward adaptor's `default` case → `UnknownNode` → `display:none`, so they rendered **invisibly**. This hid handbook single-cell section headings (HBKENG/TNN/TND) — e.g. TND's boxed *"Section 3:1–12…"* headings — and 2 Samuel's multi-column genealogy tables.

## What

- **`libs/shared`**: three new `ElementNode` subclasses — `TableNode` / `TableRowNode` / `TableCellNode` — rendering real `<table>`/`<tr>`/`<td>|<th>` with `data-marker` + `usfm_<marker>` classes (following the `ParaNode` pattern). Cells map `align`→`text-align`, `colspan`→`colSpan`, and pick `th` vs `td` by marker prefix. Registered in `usjBaseNodes`.
- **`packages/platform`**: forward adaptor (`usj-editor`) builds the new nodes; reverse adaptor (`editor-usj`) serializes them back to USJ. Round-trip is **lossless** — `marker`, `align`, `colspan`, and arbitrary `unknownAttributes` (e.g. `category`) all survive.
- **`packages/utilities`**: `colspan` added to the `MarkerObject` model (+ API report).
- **`libs/test-data`**: regenerated 2 Samuel Lexical fixtures (tables were `UnknownNode`s, now real table nodes).

## Testing

- `shared` 105/105, `platform-editor` 119 pass (+3 generator skips), `utilities` 19/19; typecheck + lint clean across all affected packages.
- New unit tests per node (DOM shape, th/td, align/colspan, JSON + `unknownAttributes` round-trip) and a reverse-adaptor round-trip test.
- **Manually verified** in the `platform` demo: the editor emits correct real-table DOM (`<table class="table"><tr class="table-row usfm_tr"><td class="table-cell usfm_tc1">…`) with proper `table`/`table-row`/`table-cell` display.

## Follow-up required in paranext-core (styling)

Table **styling** is intentionally a consumer concern (no CSS ships here). Manual testing surfaced that the existing handbook stylesheet styles table *markers* as indented block paragraphs (`.text-spacing .usfm_tr { text-indent: -5vw; margin-left: 10vw }`); because `text-indent` is inherited, it collapses real table columns on top of each other. The paranext-core follow-up must author real-`<table>` CSS **and** neutralize those paragraph rules (plus capture the PT9 box styling for single-cell headings). Details + a ready-to-run kickoff prompt are in the design spec.

## Out of scope (separate follow-ups)

- paranext-core table/box CSS (above)
- Scribe adaptor wiring (Scribe has its own adaptors with the same gap)
- `StructureProtectionPlugin` paste protection for tables

- - -
This change is [Reviewable](https://reviewable.io/reviews/eten-tech-foundation/scripture-editors/494)


irahopkinson (changes requested):
This isn't a code review - I was just curious how you were adding table support so I had a quick look.

> @irahopkinson made 2 comments.
> *[Reviewable](https://reviewable.io/reviews/eten-tech-foundation/scripture-editors/494#-:-OwpQrv-ErZz3r-8pCTO:b1ohimo)* status: 0 of 19 files reviewed, 1 unresolved discussion (waiting on katherinejensen00).

___
*[`libs/shared/src/nodes/usj/TableCellNode.ts` line 46 at r1](https://reviewable.io/reviews/eten-tech-foundation/scripture-editors/494#-OwpPepVCJXFtqDhkzT7:-OwpPepVCJXFtqDhkzT8:b-c8zsqq) ([raw file](https://github.com/eten-tech-foundation/scripture-editors/blob/bea9b0f975d3b8f2190193c15641c3c5b652505e/libs/shared/src/nodes/usj/TableCellNode.ts#L46)):*
> ```XML
> }
> 
> export class TableCellNode extends ElementNode {
> ```

For these 3 new nodes why not use the existing nodes, i.e.:
```ts
import {
  TableCellNode,
  TableNode,
  TableRowNode,
} from '@lexical/table';
```
For read-only you may not need it but there is also `import {TablePlugin} from '@lexical/react/LexicalTablePlugin';`.

Given all that there is also new way to do custom nodes with extensions instead so you can also do tables using the Extension architecture which is likely the better path forward. Look for demo code comparison in https://lexical.dev/gallery.

irahopkinson (changes requested):
> @irahopkinson made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/eten-tech-foundation/scripture-editors/494)* status: 0 of 19 files reviewed, 2 unresolved discussions (waiting on katherinejensen00).

___
*[`packages/utilities/src/converters/usj/usj.model.ts` at r1](https://reviewable.io/reviews/eten-tech-foundation/scripture-editors/494#-OwpRDl-DSC01kcNtSmJ:-OwpRDl-DSC01kcNtSmK:b9bzbon) ([raw file](https://github.com/eten-tech-foundation/scripture-editors/blob/bea9b0f975d3b8f2190193c15641c3c5b652505e/packages/utilities/src/converters/usj/usj.model.ts)):*
Revert changes to this file as we don't want to allow this node-specific property for all nodes. You can deal with node-specific properties in the node without this change.

katherinejensen00 (commented):
Thanks for looking, Ira! Two things:

Fixed **`usj.model.ts` / `colspan`** — reverted. `colspan` is no longer on the shared `MarkerObject`. It's now a node-local `TableCellMarker = MarkerObject & { colspan?: string }` defined in `TableCellNode`, and the forward adaptor reads it through that type. Nothing table-specific leaks into the shared model anymore.

**Using `@lexical/table` / the Extension architecture** — I looked into it. Here is what Claude said:

 For an _editable_ table it'd be the right call, but these are read-only render nodes and the fit isn't there:
-   `@lexical/table`'s `TableCellNode` has no slot for the USJ data we round-trip — no `marker`, no logical `align` (it only has `verticalAlign`), and `colSpan` is a `number` (USJ carries it as a string) — and nowhere for `unknownAttributes`.
-   Its `createDOM` emits theme classes + width/background/vertical-align styling and derives `th`/`td` from a numeric `headerState`; we need `data-marker`, `usfm_${marker}`, logical `text-align`, and `th`/`td` from the marker prefix — so we'd override all of it.
-   Its `SerializedTableCellNode` requires `headerState` (plus `colSpan`/`rowSpan`/`width`/…), which would either push those editing fields into our USJ editor state or need casts to work around.

So subclassing would mean overriding essentially everything while inheriting unused grid/editing machinery — more code, no reuse, for the read-only path. I've kept the purpose-built nodes for now. If these tables become editable later, migrating onto `@lexical/table` (or the Extension architecture) makes sense then — happy to file a follow-up to track it.

> @katherinejensen00 made 3 comments.
> *[Reviewable](https://reviewable.io/reviews/eten-tech-foundation/scripture-editors/494#-:-OwulHxf8Zuc9Tz-TPLp:b8x4z7v)* status: 0 of 19 files reviewed, 2 unresolved discussions (waiting on irahopkinson).

___
*[`libs/shared/src/nodes/usj/TableCellNode.ts` line 46 at r1](https://reviewable.io/reviews/eten-tech-foundation/scripture-editors/494#-OwpPepVCJXFtqDhkzT7:-OwulDtH5zcag5c70Tiq:b-152u7m) ([raw file](https://github.com/eten-tech-foundation/scripture-editors/blob/7d53567ba3163d295d73cc88f78218c9636ad345/libs/shared/src/nodes/usj/TableCellNode.ts#L46)):*

For these 3 new nodes why not use the existing nodes, i.e.:
```ts
import {
  TableCellNode,
  TableNode,
  TableRowNode,
} from '@lexical/table';
```
For read-only you may not need it but there is also `import {TablePlugin} from '@lexical/react/LexicalTablePlugin';`.

Given all that there is also new way to do custom nodes with extensions instead so you can also do tables using the Extension architecture which is likely the better path forward. Look for demo code comparison in https://lexical.dev/gallery.
</blockquote></details>

Researched, but it didn't work out. See summary message for more details.
___
*[`packages/utilities/src/converters/usj/usj.model.ts` at r1](https://reviewable.io/reviews/eten-tech-foundation/scripture-editors/494#-OwpRDl-DSC01kcNtSmJ:-OwulDOn6GuXGLzpHvFV:b-896fix) ([raw file](https://github.com/eten-tech-foundation/scripture-editors/blob/7d53567ba3163d295d73cc88f78218c9636ad345/packages/utilities/src/converters/usj/usj.model.ts)):*

Revert changes to this file as we don't want to allow this node-specific property for all nodes. You can deal with node-specific properties in the node without this change.
</blockquote></details>

Done.

irahopkinson (changes requested):
> @irahopkinson reviewed 12 files and all commit messages, made 1 comment, and resolved 1 discussion.
> *[Reviewable](https://reviewable.io/reviews/eten-tech-foundation/scripture-editors/494)* status: 12 of 19 files reviewed, 1 unresolved discussion (waiting on katherinejensen00).

___
*[`libs/shared/src/nodes/usj/TableCellNode.ts` line 46 at r1](https://reviewable.io/reviews/eten-tech-foundation/scripture-editors/494#-OwpPepVCJXFtqDhkzT7:-OxcdFG21R1P00-g5HPD:bl32vtv) ([raw file](https://github.com/eten-tech-foundation/scripture-editors/blob/7d53567ba3163d295d73cc88f78218c9636ad345/libs/shared/src/nodes/usj/TableCellNode.ts#L46)):*

Researched, but it didn't work out. See summary message for more details.
</blockquote></details>

In that case follow the existing pattern and add `Immutable` to front of each of their names. That also means the current names are free for the editable version of them and then we can switch between them if we need or just deprecate the immutable ones if they are no longer needed.

katherinejensen00 (commented):
> @katherinejensen00 made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/eten-tech-foundation/scripture-editors/494)* status: 5 of 20 files reviewed, 1 unresolved discussion (waiting on irahopkinson).

___
*[`libs/shared/src/nodes/usj/TableCellNode.ts` line 46 at r1](https://reviewable.io/reviews/eten-tech-foundation/scripture-editors/494#-OwpPepVCJXFtqDhkzT7:-OxlAklP7EymTc524hDi:bws8ezy) ([raw file](https://github.com/eten-tech-foundation/scripture-editors/blob/7d53567ba3163d295d73cc88f78218c9636ad345/libs/shared/src/nodes/usj/TableCellNode.ts#L46)):*

In that case follow the existing pattern and add `Immutable` to front of each of their names. That also means the current names are free for the editable version of them and then we can switch between them if we need or just deprecate the immutable ones if they are no longer needed.
</blockquote></details>

Fixed— renamed all three to `ImmutableTableNode` / `ImmutableTableRowNode` / `ImmutableTableCellNode`, following the `ImmutableChapterNode`/`ImmutableVerseNode` pattern. Their `getType()` now serializes as `immutable-table` / `immutable-table-row` / `immutable-table-cell` too, so they're consistent with the other immutable render nodes (they'd been the only ones still using the bare USJ type).

The USJ marker types (`"table"`/`"table:row"`/`"table:cell"`) are kept as separate constants: the forward adaptor matches USJ input against them, and the reverse adaptor emits them explicitly — so `table` is now free for an editable variant to claim later, and we can switch between the two exactly like the verse pair does.

irahopkinson (dismissed):
> @irahopkinson partially reviewed 15 files and all commit messages, made 3 comments, and resolved 1 discussion.
> *[Reviewable](https://reviewable.io/reviews/eten-tech-foundation/scripture-editors/494)* status: all files reviewed, 3 unresolved discussions (waiting on katherinejensen00).

___
*[`libs/shared/src/nodes/usj/ImmutableTableCellNode.test.ts` line 94 at r4](https://reviewable.io/reviews/eten-tech-foundation/scripture-editors/494#-Oy1DEGI6ZIUa4B_saHN:-Oy1DEGI6ZIUa4B_saHO:bmz08hb) ([raw file](https://github.com/eten-tech-foundation/scripture-editors/blob/44eb01c7294a7c75db59d25daf3660b1da4c3dc7/libs/shared/src/nodes/usj/ImmutableTableCellNode.test.ts#L94)):*
> ```XML
>     });
>     const container = document.createElement("div");
>     editor.setRootElement(container);
> ```

NIT should use `createBasicTestEnvironment` or some other test utility rather than roll-your-own.



_Code quote:_
```XML
    const editor = createEditor({
      nodes: [ImmutableTableCellNode],
      onError: (e) => {
        throw e;
      },
    });
    const container = document.createElement("div");
    editor.setRootElement(container);
```
___
*[`libs/shared/src/nodes/usj/ImmutableTableCellNode.test.ts` line 107 at r4](https://reviewable.io/reviews/eten-tech-foundation/scripture-editors/494#-Oy1DlbOAW7JXasUAnzA:-Oy1DlbOAW7JXasUAnzB:b-8po5jv) ([raw file](https://github.com/eten-tech-foundation/scripture-editors/blob/44eb01c7294a7c75db59d25daf3660b1da4c3dc7/libs/shared/src/nodes/usj/ImmutableTableCellNode.test.ts#L107)):*
> ```XML
>     );
> 
>     editor.update(
> ```

NIT can you tell me why you need 2 separate `editor.update`s. Either join them or comment why they are needed.



_Code quote:_
```XML
editor.update(
```
___
*[`libs/test-data/src/data/2sa.lexical.editable.ts` line 8845 at r4](https://reviewable.io/reviews/eten-tech-foundation/scripture-editors/494#-Oy16KlJ1ci_ETjQtc0g:-Oy16KlJ1ci_ETjQtc0h:bh5h511) ([raw file](https://github.com/eten-tech-foundation/scripture-editors/blob/d97560eebf720f55d5eed965498b9b4daae39a45/libs/test-data/src/data/2sa.lexical.editable.ts#L8845)):*
> ```XML
>                     detail: 0,
>                     format: 0,
>                     mode: "token",
> ```

BTW we might want to retain this token mode as it stops the text from being editable since the table is immutable. It would need to be created that way in the usj-editor adaptor. Is there some other way provided that the text in an immutable table remains un-editable? Is there some reason it should be editable?



_Code quote:_
```XML
mode: "token",
```

katherinejensen00:
Re: `mode: "token"` in `2sa.lexical.editable.ts`

The "Immutable" prefix on these nodes means the *table structure* is immutable (users can't add or remove rows/cells) — not that text content inside cells is locked. `isShadowRoot()` on `ImmutableTableCellNode` provides that structural boundary (selection stays within a cell, like a real table).

The text nodes inside cells should remain editable: this is the *editable* editor path (`2sa.lexical.editable.ts`) and scripture table cells need to be authorable. `mode: "normal"` is correct there; `mode: "token"` was present in an earlier revision before the adaptor fully wired up the children and was removed when the content was made properly editable.

No change needed.

katherinejensen00:
Re: `mode: "token"` in `2sa.lexical.editable.ts`

The "Immutable" prefix on these nodes means the *table structure* is immutable (users can't add or remove rows/cells) — not that text content inside cells is locked. `isShadowRoot()` on `ImmutableTableCellNode` provides that structural boundary (selection stays within a cell, like a real table).

The text nodes inside cells should remain editable: this is the *editable* editor path (`2sa.lexical.editable.ts`) and scripture table cells need to be authorable. `mode: "normal"` is correct there; `mode: "token"` was present in an earlier revision before the adaptor fully wired up the children and was removed when the content was made properly editable.

No change needed.

katherinejensen00 (commented):
> @katherinejensen00 made 3 comments.
> *[Reviewable](https://reviewable.io/reviews/eten-tech-foundation/scripture-editors/494)* status: all files reviewed, 3 unresolved discussions (waiting on irahopkinson).

___
*[`libs/test-data/src/data/2sa.lexical.editable.ts` line 8845 at r4](https://reviewable.io/reviews/eten-tech-foundation/scripture-editors/494#-Oy16KlJ1ci_ETjQtc0g:-Oy9kG8B2cZ4W8BiekkC:b1pqwrt) ([raw file](https://github.com/eten-tech-foundation/scripture-editors/blob/e7113f4d65d46f9c75527832725933c61d691143/libs/test-data/src/data/2sa.lexical.editable.ts#L8845)):*

BTW we might want to retain this token mode as it stops the text from being editable since the table is immutable. It would need to be created that way in the usj-editor adaptor. Is there some other way provided that the text in an immutable table remains un-editable? Is there some reason it should be editable?
</blockquote></details>

The "Immutable" prefix on these nodes means the _table structure_ is immutable (users can't add or remove rows/cells) — not that text content inside cells is locked. `isShadowRoot()` on `ImmutableTableCellNode` provides that structural boundary (selection stays within a cell, like a real table).

The text nodes inside cells should remain editable: this is the _editable_ editor path (`2sa.lexical.editable.ts`) and scripture table cells need to be authorable. `mode: "normal"` is correct there; `mode: "token"` was present in an earlier revision before the adaptor fully wired up the children and was removed when the content was made properly editable.

Based on that,  I didn't make any changes, but feel free to push back on that idea.
___
*[`libs/shared/src/nodes/usj/ImmutableTableCellNode.test.ts` line 94 at r4](https://reviewable.io/reviews/eten-tech-foundation/scripture-editors/494#-Oy1DEGI6ZIUa4B_saHN:-Oy9k6e03rTkXuEMLdQv:b-q5xsov) ([raw file](https://github.com/eten-tech-foundation/scripture-editors/blob/4f82f538a958f495a3822b38c309dcd9f547511d/libs/shared/src/nodes/usj/ImmutableTableCellNode.test.ts#L94)):*

NIT should use `createBasicTestEnvironment` or some other test utility rather than roll-your-own.
</blockquote></details>

Good idea. Fixed.
___
*[`libs/shared/src/nodes/usj/ImmutableTableCellNode.test.ts` line 107 at r4](https://reviewable.io/reviews/eten-tech-foundation/scripture-editors/494#-Oy1DlbOAW7JXasUAnzA:-Oy9k8o39_mnVv7nTo4t:b3dohs9) ([raw file](https://github.com/eten-tech-foundation/scripture-editors/blob/4f82f538a958f495a3822b38c309dcd9f547511d/libs/shared/src/nodes/usj/ImmutableTableCellNode.test.ts#L107)):*

NIT can you tell me why you need 2 separate `editor.update`s. Either join them or comment why they are needed.
</blockquote></details>

Yes, added comments to make that more clear for everyone reading this code.

irahopkinson (approved):
:lgtm: thanks for finishing this off.

> @irahopkinson partially reviewed 1 file and all commit messages, made 1 comment, and resolved 3 discussions.
> *[Reviewable](https://reviewable.io/reviews/eten-tech-foundation/scripture-editors/494#-:-OyAMAYG95Sd4HwSCcZP:b-n36d1r)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved.

PR #496, opened 2026-07-06
PT-3835: Make USJ selection locations annotation-transparent

irahopkinson:
# Code Review Summary

**Branch**: pt-3835-annotation-selection

**Base**: origin/main

**Date**: 2026-07-06

**Review model**: Claude Fable 5

**Files changed**: 8 (7 in the committed squash; in-review fixes touched 1 additional file, `packages/platform/src/editor/adaptors/editor-usj.adaptor.ts`. Step 1.3 reported 12, which included 5 uncommitted local workstation config files excluded from this review.)

## Overview

Fixes Jira PT-3835: selection locations (`jsonPath` + `offset`) reported by `onSelectionChange` and consumed by `setAnnotation` previously counted `TypedMarkNode` annotation wrappers as content items, even though annotations do not exist in exported USJ (the editor→USJ adaptor strips them and coalesces split text back into single strings). Once one annotation existed in a paragraph, locations at or after it were wrong against the real USJ; hosts like paranext-core that resolve reported locations against actual USJ failed (e.g. `No result found for JSONPath query: $.content[20].content[2]`) when inserting annotations front-to-back. Inside this repo the bug was invisible because both conversion directions shared the same wrong convention and the errors cancelled.

The fix adds a logical content model to `libs/shared/src/nodes/usj/node.utils.ts` (`$getLogicalContentItems` and lookup helpers) that mirrors what `recurseNodes` in the platform `editor-usj.adaptor.ts` exports — presentation-only nodes skipped without breaking text runs, TypedMarkNodes transparent, contiguous plain text coalesced with cumulative segment offsets — and rebuilds both conversion directions in `libs/shared-react/.../selection.utils.ts` on that single model. Published package APIs are unchanged. Test coverage: unit tests for the model, a 204-case data-driven annotation-invariance suite (3 marker modes × 2 insertion orders over the 2SA corpus), `AnnotationPlugin` integration tests reproducing the repro flow in both orders, and an "Insert at selection" button in the platform demo for manual verification.

## API Changes

- `packages/*`: no files changed — the published public API surface (exports, types, signatures) is unchanged.
- `libs/shared` (`node.utils.ts`, re-exported via `export *` barrels): added exports — types `LogicalTextSegment`, `LogicalTextItem`, `LogicalContentItem`, `LogicalPoint`; functions `$shouldIgnoreNodeForContentIndexes`, `$getLogicalContentItems`, `$getLogicalParent`, `$getLogicalIndexOfChild`, `$getLogicalTextLocation`, `$getTextNodeAtLogicalOffset`, `$getLogicalPointFromElementPoint`, `$getElementOffsetFromLogicalIndex`. All additive; no exports removed or modified.
- `libs/shared-react` (`selection.utils.ts`): exported surface unchanged (`$getRangeFromUsjSelection`, `$getUsjSelectionFromEditor`); only private helpers were replaced.
- Runtime behavior of the unchanged APIs changed by design: `UsjTextContentLocation` values emitted by `onSelectionChange` and consumed by `setAnnotation` are now annotation-transparent (coalesced-USJ coordinates) in documents containing annotations.

(Analysis caveat: the API checklist targets paranext-core surfaces; the analyzer mapped it onto this repo's structure — `packages/*` = published API, `libs/*` = internal.)

None.

- [ ] ~~Deliberate behavior change on unchanged API signatures (`selection.utils.ts`): hosts that shipped workarounds compensating for the old annotation-inclusive coordinates would double-adjust; recommend release-notes callout and coordinated paranext-core rollout~~ _(author: the previous behavior was not intentional — it was the bug; this fix was surfaced by paranext-core itself and there are no known workarounds)_
- [x] `$getLogicalContentItems` JSDoc overstated exporter parity: comment-type TypedMarkNodes are serialized as milestone items by the exporter, so logical indexes diverge for paragraphs containing comments _(fixed during review: known-exclusion note added to the JSDoc — comment marks are treated as transparent like all marks; the milestone serialization is deprecated and pending removal, so the model intentionally ignores it)_

[Author response: Finding 1 dismissed as designed — bug fix restoring the documented contract. Finding 2 fixed with the documentation note the author requested.]

### Minor — Consider

- [x] Exporter-parity gap: exporter drops NBSP-only/empty text nodes, model counted their characters _(fixed during review: `$shouldIgnoreNodeForContentIndexes` now skips empty and NBSP-only spacer text, with a revert-tested unit test using the real-world shape — a standalone NBSP spacer between element items, since Lexical merges adjacent plain text nodes. The deeper fix — creating NBSP spacers as presentation-typed text nodes in the USJ→editor direction — is noted in the code as follow-up work beyond this PR.)_
- [ ] ~~Mixed-children-inside-mark boundary: a logical boundary between a text child and a non-text child inside a mark resolves past the whole mark (`$getElementOffsetFromLogicalIndex` / `$getLocationFromNode`)~~ _(author dismissed as the documented known edge: rare scenario, output is a valid nearby point in the correct run — strictly better than both prior behaviors; an exact fix requires a resolution-side change able to return points inside a mark (~30–50 lines), reviewed and declined for this PR)_
- [x] The model mirrors `recurseNodes` but neither exporter copy referenced it _(fixed during review: "keep in sync" comment added above the platform `recurseNodes`; the scribe copy was dismissed — scribe is no longer being kept updated)_
- [x] Code comment cited "(see PT-3835 final review)" — a reference unreachable from the code _(fixed during review: citation trimmed; the surrounding comment carries the reasoning)_
- [x] `$getTextNodeAtLogicalOffset` accepted the full `LogicalContentItem` union and rejected element items at runtime _(fixed during review: parameter tightened to a new exported `LogicalTextItem` interface — misuse is now a compile error; tests use a narrowing `$getFirstParaTextItem()` helper)_
- [x] Demo "Insert at selection" id `sel-${Date.now()}` can collide on same-millisecond clicks _(fixed during review: comment added noting the collision is acceptable for demo-only scaffolding)_
- [ ] ~~`$shouldIgnoreNodeForContentIndexes` has no direct unit tests~~ _(author dismissed as adequately covered: moved-verbatim logic exercised indirectly by the `$getLogicalContentItems` tests, and the new NBSP spacer test directly covers the newest branch)_

[Author response: author requested fixes for five of the seven; the two dismissals are reasoned above.]

#### Shared Regions Modified

None. (No `#region shared with` markers exist in this repository.)

#### Extension Config Changes

None. (No config/build files changed; this repo has no `extensions/` directory.)

## Positive Observations

- Strong consolidation: five ad-hoc one-direction content-index helpers were replaced by a single well-documented logical content model in `libs/shared`, and both conversion directions now derive from it — eliminating the asymmetry-by-duplication that caused PT-3835.
- `$shouldIgnoreNodeForContentIndexes` was moved verbatim to the correct layer and consumed through existing barrels; new code reuses the existing `$isDescendantOf` helper rather than re-implementing ancestry checks.
- TSDoc on the new model is unusually thorough about edge cases (VerseNode/MarkerNode extending TextNode; nested-mark recursion as defense-in-depth; segment-boundary preference at annotation edges).
- Test investment matches the change: every new exported utility has direct unit tests; the 204-case invariance suite derives its annotation ranges from the corpus itself and hard-fails (rather than silently skipping) if the corpus stops qualifying; integration tests replicate the actual host "insert at reported selection" flow in both insertion orders.
- The changed expectation in the pre-existing test suite is the intended behavior fix and is annotated with an explanatory comment; tests are behavior-focused (round-trip assertions against USJ locations, not implementation internals).
- The demo button guards correctly (`isUsjTextContentLocation` on both endpoints, null-checked ref) and follows the demo app's existing conventions.

## Interview Notes

- **Stated purpose**: bug fix for PT-3835 (from the branch commit message) — make selection locations annotation-transparent so they are interpreted against the actual USJ.
- **Root cause, in the author's words**: what was hard to spot was that adding annotations forward or backward through the text worked in this repo (even though forward failed in core) because the Lexical↔USJ selection conversion had the same error in both directions — the errors cancelled. You had to inspect the actual selection values and ask whether they were correct, rather than observing annotations appearing in the expected places. This demonstrates clear author understanding of the failure mode and the fix.
- **Behavior-change finding**: dismissed as designed — the old coordinates were unintentional; no known host workarounds exist to break.
- **Comment milestones**: deprecated code pending removal; the model intentionally ignores milestone serialization (now documented in the JSDoc).
- **Scribe**: no longer being kept updated — its copy of `recurseNodes` was deliberately left without the sync comment.
- **Verification status**: paranext-core has been retested against this fix and the original repro is resolved. Playwright e2e tests in paranext-core remain follow-up work (per the ticket's Definition of Done).
- **New bug surfaced in core**: with this fix in place, a separate pre-existing paranext-core bug is now easier to see — it reports a notification that the selection can't include markers even though the selection doesn't include any. To be tracked in core.
- No areas of author uncertainty or AI-deferral were observed; the author explained the design and each dismissal with specific technical reasoning.

## In-Review Quality Check

- `pnpm nx run-many -t typecheck`: passed for all 10 projects.
- `pnpm nx run-many -t lint`: one error auto-fixed — the new `LogicalTextItem` type alias violated `@typescript-eslint/consistent-type-definitions` and was changed to an `interface`; 0 errors remain (pre-existing warnings only).
- `pnpm nx format:check`: passed.
- Tests: `shared` 107 passed; `shared-react` 1265 passed (3 pre-existing skips). No unfixable failures.

## Suggested Review Focus

- [ ] The logical content model (`$getLogicalContentItems` in `libs/shared/src/nodes/usj/node.utils.ts`): confirm the skip/splice/coalesce rules match the platform exporter's `recurseNodes` case-for-case, including the plain-`"text"`-type-only run membership (VerseNode extends TextNode) and the new NBSP/empty spacer skip.
- [ ] The self-cancelling-error root cause: worth walking through how both conversion directions shared the wrong convention, since it explains why this repo's tests couldn't catch the bug before the invariance suite existed.
- [ ] Consumer-visible nuances for the paranext-core rollout: annotation-edge element points now emit deeper `textContent` locations; stale mark-counting locations persisted by hosts resolve to `undefined` (the bug being fixed).
- [ ] Dismissed known edge: interior element point between mixed children inside a mark approximates to a nearby point in the correct run — confirm the team is comfortable shipping this documented limitation.
- [ ] Follow-up items to track: USJ→editor adaptor creating NBSP spacers as presentation-typed text nodes; Playwright e2e in paranext-core; the newly visible core bug (spurious "selection can't include markers" notification).

- - -
This change is [Reviewable](https://reviewable.io/reviews/eten-tech-foundation/scripture-editors/496)

### DIFF
--- a/demos/platform/src/app/app.tsx
+++ b/demos/platform/src/app/app.tsx
@@ -27,6 +27,7 @@ import {
   MarginalRef,
   MarkerMode,
   NoteMode,
+  SelectionRange,
   TextDirection,
   UsjNodeOptions,
   ViewOptions,
@@ -116,6 +117,7 @@ export default function App() {
   const marginalRef = useRef<MarginalRef | null>(null);
   const noteEditorRef = useRef<EditorRef | null>(null);
   const noteNodeKeyRef = useRef<string | undefined>();
+  const selectionRef = useRef<SelectionRange | undefined>(undefined);
   const [isNoteEditorVisible, setIsNoteEditorVisible] = useState(false);
   const [isOptionsDefined, setIsOptionsDefined] = useState(false);
   const [isReadonly, setIsReadonly] = useState(false);
@@ -409,6 +411,34 @@ export default function App() {
               >
                 stand
               </button>
+              <button
+                id="insertAtSelection"
+                onClick={() => {
+                  const selection = selectionRef.current;
+                  if (
+                    selection?.start &&
+                    selection?.end &&
+                    marginalRef.current &&
+                    isUsjTextContentLocation(selection.start) &&
+                    isUsjTextContentLocation(selection.end)
+                  ) {
+                    // Demo-only id: could collide on same-millisecond clicks, which is fine here.
+                    const id = `sel-${Date.now()}`;
+                    marginalRef.current.setAnnotation(
+                      { start: selection.start, end: selection.end },
+                      annotationType,
+                      id,
+                      handleAnnotationOnClick,
+                      handleAnnotationOnRemove,
+                    );
+                    console.log("Inserted annotation at selection", { selection, id });
+                  } else {
+                    console.warn("No valid selection for annotation");
+                  }
+                }}
+              >
+                Insert at selection
+              </button>
             </div>
           </span>
           <pre title="contextMarker" style={{ color: "black" }}>
@@ -525,7 +555,10 @@ export default function App() {
           defaultUsj={EMPTY_USJ}
           scrRef={scrRef}
           onScrRefChange={setScrRef}
-          onSelectionChange={(selection) => console.log({ selection })}
+          onSelectionChange={(selection) => {
+            selectionRef.current = selection;
+            console.log({ selection });
+          }}
           onCommentChange={(comments) => console.log({ comments })}
           onUsjChange={handleUsjChange}
           onStateChange={({
--- a/libs/shared-react/src/plugins/usj/annotation/AnnotationPlugin.test.tsx
+++ b/libs/shared-react/src/plugins/usj/annotation/AnnotationPlugin.test.tsx
@@ -1,8 +1,21 @@
+// Reaching inside only for tests.
+// eslint-disable-next-line @nx/enforce-module-boundaries
+import { updateSelection } from "../../../../../../libs/shared/src/nodes/usj/test.utils";
 import { baseTestEnvironment } from "../react-test.utils";
 import { AnnotationPlugin, AnnotationRef } from "./AnnotationPlugin";
 import { AnnotationRange } from "./selection.model";
+import { $getUsjSelectionFromEditor } from "./selection.utils";
+import { isUsjTextContentLocation } from "@eten-tech-foundation/scripture-utilities";
 import { act } from "@testing-library/react";
-import { $createTextNode, $getRoot, $isElementNode, $isTextNode, LexicalNode } from "lexical";
+import {
+  $createTextNode,
+  $getRoot,
+  $isElementNode,
+  $isTextNode,
+  LexicalEditor,
+  LexicalNode,
+  TextNode,
+} from "lexical";
 import { createRef } from "react";
 import {
   $createParaNode,
@@ -207,6 +220,88 @@ describe("AnnotationPlugin", () => {
       });
     });
   });
+});
+
+describe("PT-3835: inserting annotations from reported selections", () => {
+  const text = "the neva word and the sleep word";
+  // indexOf: "neva" = 4, "sleep" = 22
+
+  it("annotates front-to-back (the PT-3835 failure direction)", async () => {
+    const { annotationPlugin, editor } = await testEnvironment(() => {
+      $getRoot().append($createParaNode().append($createTextNode(text)));
+    });
+
+    await annotateWordViaReportedSelection(annotationPlugin, editor, "neva", "c-1");
+    await annotateWordViaReportedSelection(annotationPlugin, editor, "sleep", "c-2");
+
+    editor.getEditorState().read(() => $expectAnnotatedWords(["neva", "sleep"]));
+  });
+
+  it("annotates back-to-front (regression guard)", async () => {
+    const { annotationPlugin, editor } = await testEnvironment(() => {
+      $getRoot().append($createParaNode().append($createTextNode(text)));
+    });
+
+    await annotateWordViaReportedSelection(annotationPlugin, editor, "sleep", "c-2");
+    await annotateWordViaReportedSelection(annotationPlugin, editor, "neva", "c-1");
+
+    editor.getEditorState().read(() => $expectAnnotatedWords(["neva", "sleep"]));
+  });
+
+  /**
+   * Select a word in the editor, verify the reported USJ selection is in coalesced-USJ
+   * coordinates (annotation-independent), then insert an annotation with exactly that
+   * reported selection — the paranext-core "Insert Comment" flow.
+   */
+  async function annotateWordViaReportedSelection(
+    annotationPlugin: AnnotationRef,
+    editor: LexicalEditor,
+    word: string,
+    id: string,
+  ) {
+    let target: TextNode | undefined;
+    let localOffset = 0;
+    editor.getEditorState().read(() => {
+      const para = $getRoot().getFirstChild();
+      if (!$isElementNode(para)) throw new Error("Expected a para node");
+      target = para.getAllTextNodes().find((node) => node.getTextContent().includes(word));
+      if (!target) throw new Error(`No text node contains '${word}'`);
+      localOffset = target.getTextContent().indexOf(word);
+    });
+    if (!target) throw new Error(`No text node contains '${word}'`);
+    updateSelection(editor, target, localOffset, target, localOffset + word.length);
+
+    const reported = editor.getEditorState().read($getUsjSelectionFromEditor);
+    const start = reported?.start;
+    const end = reported?.end;
+    if (!start || !end || !isUsjTextContentLocation(start) || !isUsjTextContentLocation(end)) {
+      throw new Error(`Expected a reported text selection for '${word}'`);
+    }
+    // The reported location must be in actual-USJ coordinates regardless of annotations.
+    expect(start).toEqual({
+      jsonPath: "$.content[0].content[0]",
+      offset: text.indexOf(word),
+    });
+    expect(end).toEqual({
+      jsonPath: "$.content[0].content[0]",
+      offset: text.indexOf(word) + word.length,
+    });
+
+    await act(async () => {
+      annotationPlugin.setAnnotation({ start, end }, "review", id);
+    });
+  }
+
+  function $expectAnnotatedWords(words: string[]) {
+    const para = $getRoot().getFirstChild();
+    if (!$isElementNode(para)) throw new Error("Expected a para node");
+    const markTexts = para
+      .getChildren()
+      .filter($isTypedMarkNode)
+      .map((node) => node.getTextContent());
+    expect(markTexts).toEqual(words);
+    expect(para.getTextContent()).toBe(text);
+  }
 });
 
 /**
--- a/libs/shared-react/src/plugins/usj/annotation/selection.utils.annotated.data-driven.test.ts
+++ b/libs/shared-react/src/plugins/usj/annotation/selection.utils.annotated.data-driven.test.ts
@@ -1,0 +1,154 @@
+/**
+ * Annotation-invariance tests for USJ ↔ Lexical selection conversion (PT-3835).
+ *
+ * Loads the usj2Sa serialized states (all 3 marker modes), wraps several USJ ranges in
+ * TypedMarkNodes (in both application orders), then asserts every textContent location still
+ * resolves and round-trips to exactly the same UsjDocumentLocation as in the unannotated
+ * document. Locations are defined against the actual USJ, which never contains annotations,
+ * so annotations must not shift them.
+ *
+ * @see selection.utils.data-driven.test.ts for the unannotated baseline suite.
+ */
+// Reaching inside only for tests.
+// eslint-disable-next-line @nx/enforce-module-boundaries
+import {
+  createBasicTestEnvironment,
+  updateSelection,
+} from "../../../../../../libs/shared/src/nodes/usj/test.utils";
+import { usjReactNodes } from "../../../nodes/usj";
+import type { SelectionRange } from "./selection.model";
+import { $getRangeFromUsjSelection, $getUsjSelectionFromEditor } from "./selection.utils";
+import { isUsjTextContentLocation } from "@eten-tech-foundation/scripture-utilities";
+import type { LexicalEditor, LexicalNode, SerializedEditorState } from "lexical";
+import { $wrapSelectionInTypedMarkNode, TypedMarkNode } from "shared";
+import type { LocationEntry2Sa } from "test-data";
+import {
+  usjLocations2Sa,
+  lexicalEditable2Sa,
+  lexicalVisible2Sa,
+  lexicalHidden2Sa,
+} from "test-data";
+
+const MARKER_MODES = [
+  { name: "editable", state: lexicalEditable2Sa },
+  { name: "visible", state: lexicalVisible2Sa },
+  { name: "hidden", state: lexicalHidden2Sa },
+] as const;
+
+const APPLICATION_ORDERS = ["front-to-back", "back-to-front"] as const;
+
+/**
+ * Matches plain text-run paths such as "$.content[20].content[0]" (paragraph-level) or
+ * "$.content[18].content[1].content[0]" (nested inside a char/note node), i.e. any dotted
+ * chain of numeric `content[N]` segments. Excludes property paths like "['marker']".
+ *
+ * Widened from a paragraph-level-only match: the usj2Sa corpus has only 3 distinct
+ * paragraph-level textContent paths, and only 2 of them have an entry at offset >= 4
+ * (the third, `$.content[16].content[2]`, is only sampled at offset 0). Including
+ * char/note-nested text runs is needed to reach 3 qualifying paths.
+ */
+const TEXT_RUN_PATH = /^\$\.content\[\d+\](\.content\[\d+\])+$/;
+
+const textEntries = usjLocations2Sa.filter(
+  (entry) =>
+    entry.locationType === "textContent" && isUsjTextContentLocation(entry.documentLocation),
+);
+
+describe("data-driven: usj2Sa textContent locations are annotation-invariant", () => {
+  describe.each(MARKER_MODES)("'$name' mode", ({ state: serializedState }) => {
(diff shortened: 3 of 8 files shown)


katherinejensen00 (changes requested):
Thanks for working on this bug! Just a couple of findings from Claude to think about. Overall though I think the code looks good and you have great test coverage. It was also cool learning more about how a bug like this was introduced and now fixed.

> @katherinejensen00 reviewed 8 files and all commit messages, and made 5 comments.
> *[Reviewable](https://reviewable.io/reviews/eten-tech-foundation/scripture-editors/496#-:-OwxHjXiECVOCZW17P7-:bofngik)* status: all files reviewed, 4 unresolved discussions (waiting on irahopkinson).

___
*[`libs/shared/src/nodes/usj/node.utils.ts` line 762 at r1](https://reviewable.io/reviews/eten-tech-foundation/scripture-editors/496#-OwxEpCv6L15IKcTvrrQ:-OwxEpCv6L15IKcTvrrR:b-o9wnyy) ([raw file](https://github.com/eten-tech-foundation/scripture-editors/blob/25207dd1aa03a8675e7420c13c13636eb999e70d/libs/shared/src/nodes/usj/node.utils.ts#L762)):*
> ```XML
>     if (textType === "marker-trailing-space" || textType === "attribute") return true;
> 
>     const text = node.getTextContent();
> ```

NIT Claude Finding

This is an undocumented parity exclusion — narrow, but worth a line since the JSDoc now claims the model mirrors the exporter.

In editable/UNFORMATTED mode the note caller is a bare `$createTextNode(getEditableCallerText(...))` (`node-react.utils.ts:261`), ~3 chars and not NBSP-only, so this function doesn't skip it and it gets coalesced into the note's text run. The exporter explicitly drops it (`editor-usj.adaptor.ts:425`: `text !== getEditableCallerText(noteCaller)`), so selections inside an editable-mode note would report offsets shifted by the caller length. The 2SA corpus samples no note-interior text (only the note's `['caller']`), so the data-driven suites can't catch this.

Suggestion: add the caller-text case here (or document it alongside the comment exclusion), and consider a unit/corpus case for a selection inside an expanded note body.
___
*[`libs/shared/src/nodes/usj/node.utils.ts` line 773 at r1](https://reviewable.io/reviews/eten-tech-foundation/scripture-editors/496#-OwxAq7v9SZI8qIGSyER:-OwxAq7v9SZI8qIGSyES:bcb2mn1) ([raw file](https://github.com/eten-tech-foundation/scripture-editors/blob/25207dd1aa03a8675e7420c13c13636eb999e70d/libs/shared/src/nodes/usj/node.utils.ts#L773)):*
> ```XML
>  * transparent (children spliced in, recursively), contiguous text coalesced into single items.
>  *
>  * Known exclusion: comment-type TypedMarkNodes are treated as transparent like every other
> ```

Claude Finding that sounded important, but you are closer to the code so feel free to push back on this.

"Comment-type marks look like they still round-trip through milestone content items on export (`recurseNodes` → `replaceMarkWithMilestones`) while the new model treats them transparently — so exported USJ and reported selection coordinates seem to disagree for paragraphs that contain a comment, which is the PT-3835 shape again. The JSDoc says the milestone path is deprecated/pending removal — is there a tracking issue for that? If it's still live for now, could we either emit comment marks as content items in the model or add a comment-mark case to the invariance suite so the gap is proven closed?"
___
*[`libs/shared-react/src/plugins/usj/annotation/selection.utils.ts` line 410 at r1](https://reviewable.io/reviews/eten-tech-foundation/scripture-editors/496#-OwxH4HeAPqKazKJt-qq:-OwxH4HeAPqKazKJt-qr:b7pvvxf) ([raw file](https://github.com/eten-tech-foundation/scripture-editors/blob/25207dd1aa03a8675e7420c13c13636eb999e70d/libs/shared-react/src/plugins/usj/annotation/selection.utils.ts#L410)):*
> ```XML
>     // to treat the mark itself as the logical parent, which would drop the mark's content index.
>     // The mark contributes no content of its own, so the boundary before/after it is the
>     // boundary before/after its own position in the logical parent.
> ```

This is a helpful comment
___
*[`libs/shared-react/src/plugins/usj/annotation/selection.utils.ts` line 469 at r1](https://reviewable.io/reviews/eten-tech-foundation/scripture-editors/496#-OwxHNPj8VPtxN6iRsUT:-OwxHNPj8VPtxN6iRsUU:b4sbyuo) ([raw file](https://github.com/eten-tech-foundation/scripture-editors/blob/25207dd1aa03a8675e7420c13c13636eb999e70d/libs/shared-react/src/plugins/usj/annotation/selection.utils.ts#L469)):*
> ```XML
>     !$isParaLikeNode(previousContentSibling) &&
>     !$isTextNode(previousContentSibling) &&
>     !$isTypedMarkNode(previousContentSibling)
> ```

Good to know that all three of these node types need to be checked

jolierabideau (dismissed):
Thank you for taking this over and fixing it! This implementation is definitely more thorough architecturally. I will close #463 

> @jolierabideau reviewed all commit messages and made 4 comments.
> *[Reviewable](https://reviewable.io/reviews/eten-tech-foundation/scripture-editors/496#-:-OwxTElcAFEfZ24taaWw:bttfby)* status: all files reviewed, 5 unresolved discussions (waiting on irahopkinson and katherinejensen00).

___
*[`libs/shared-react/src/plugins/usj/annotation/selection.utils.ts` line 410 at r1](https://reviewable.io/reviews/eten-tech-foundation/scripture-editors/496#-OwxH4HeAPqKazKJt-qq:-OwxVdA22b0yanqWwRtb:b24cige) ([raw file](https://github.com/eten-tech-foundation/scripture-editors/blob/25207dd1aa03a8675e7420c13c13636eb999e70d/libs/shared-react/src/plugins/usj/annotation/selection.utils.ts#L410)):*

This is a helpful comment
</blockquote></details>

Agreed, this comment is helpful. Claude pointed out that it doesn't mention the case of when a cursor lands inside a mark that has several children of different kinds. `ts
const elementOffset = offset >= childrenSize ? markIndex + 1 : markIndex;` This doesn't place the point between the exact children it snaps to the front edge of the whole mark. So the reported position would be a little off. Recommending another sentence in the comment to describe.
___
*[`libs/shared/src/nodes/usj/node.utils.ts` line 869 at r1](https://reviewable.io/reviews/eten-tech-foundation/scripture-editors/496#-OwxVDgI35JDRYfDAT1q:-OwxVDgI35JDRYfDAT1r:b35l1e7) ([raw file](https://github.com/eten-tech-foundation/scripture-editors/blob/25207dd1aa03a8675e7420c13c13636eb999e70d/libs/shared/src/nodes/usj/node.utils.ts#L869)):*
> ```XML
> 
> /** A point in logical USJ content: between items, or inside a coalesced text item. */
> export type LogicalPoint =
> ```

NIT the other three `Logical*` types are grouped just above `$getLogicalContentItems`, but `LogicalPoint` is defined mid-file here. Near-first-use is defensible; grouping all four together would be marginally more consistent. Non-blocking.
___
*[`packages/platform/src/editor/adaptors/editor-usj.adaptor.ts` line 339 at r1](https://reviewable.io/reviews/eten-tech-foundation/scripture-editors/496#-OwxW4nu4lYBbzUjYBqz:-OwxW4nu4lYBbzUjYBr-:b-j1xl9m) ([raw file](https://github.com/eten-tech-foundation/scripture-editors/blob/25207dd1aa03a8675e7420c13c13636eb999e70d/packages/platform/src/editor/adaptors/editor-usj.adaptor.ts#L339)):*
> ```XML
> // Keep this function's content semantics in sync with `$getLogicalContentItems` in
> // `libs/shared/src/nodes/usj/node.utils.ts` — the logical content model mirrors which nodes
> // this export skips, splices (TypedMarkNodes), and coalesces into single text strings.
> ```

Nice to have this comment. Katherine already suggested adding a test case for a selection inside a note body, which would catch the specific mismatch she found. A broader safety net would be a test that runs the same node tree through both functions and confirms they produce the same item boundaries that way any future drift between them gets caught automatically, not just the one case. Might be worth doing both.

irahopkinson (commented):
I've already used more than the time I have for coding in this sprint. Either approve as is and add your own follow-up PR or take over the branch and fix yourself. However I have fixed a couple of things you each suggested.

> @irahopkinson made 5 comments and resolved 3 discussions.
> *[Reviewable](https://reviewable.io/reviews/eten-tech-foundation/scripture-editors/496#-:-Ox-F4St3tsb4XSyK7vi:bk7s39x)* status: 6 of 8 files reviewed, 2 unresolved discussions (waiting on katherinejensen00).

___
*[`libs/shared/src/nodes/usj/node.utils.ts` line 762 at r1](https://reviewable.io/reviews/eten-tech-foundation/scripture-editors/496#-OwxEpCv6L15IKcTvrrQ:-Ox-Dw_l1N5r-VSmZ6z3:b-lkmkzy) ([raw file](https://github.com/eten-tech-foundation/scripture-editors/blob/25207dd1aa03a8675e7420c13c13636eb999e70d/libs/shared/src/nodes/usj/node.utils.ts#L762)):*

NIT Claude Finding

This is an undocumented parity exclusion — narrow, but worth a line since the JSDoc now claims the model mirrors the exporter.

In editable/UNFORMATTED mode the note caller is a bare `$createTextNode(getEditableCallerText(...))` (`node-react.utils.ts:261`), ~3 chars and not NBSP-only, so this function doesn't skip it and it gets coalesced into the note's text run. The exporter explicitly drops it (`editor-usj.adaptor.ts:425`: `text !== getEditableCallerText(noteCaller)`), so selections inside an editable-mode note would report offsets shifted by the caller length. The 2SA corpus samples no note-interior text (only the note's `['caller']`), so the data-driven suites can't catch this.

Suggestion: add the caller-text case here (or document it alongside the comment exclusion), and consider a unit/corpus case for a selection inside an expanded note body.
</blockquote></details>

Nah, out-of-scope for this fix - unformatted mode doesn't claim to be functional yet.

We likely need to refactor to have structural spaces (as compared with content spaces) having specific nodes rather than plain text nodes. Those nodes now exist but we haven't refactored to use them.
___
*[`libs/shared/src/nodes/usj/node.utils.ts` line 773 at r1](https://reviewable.io/reviews/eten-tech-foundation/scripture-editors/496#-OwxAq7v9SZI8qIGSyER:-Ox-Ei4HD4I5bhkB2TMo:b-fv5tjj) ([raw file](https://github.com/eten-tech-foundation/scripture-editors/blob/25207dd1aa03a8675e7420c13c13636eb999e70d/libs/shared/src/nodes/usj/node.utils.ts#L773)):*

Claude Finding that sounded important, but you are closer to the code so feel free to push back on this.

"Comment-type marks look like they still round-trip through milestone content items on export (`recurseNodes` → `replaceMarkWithMilestones`) while the new model treats them transparently — so exported USJ and reported selection coordinates seem to disagree for paragraphs that contain a comment, which is the PT-3835 shape again. The JSDoc says the milestone path is deprecated/pending removal — is there a tracking issue for that? If it's still live for now, could we either emit comment marks as content items in the model or add a comment-mark case to the invariance suite so the gap is proven closed?"
</blockquote></details>

Out-of-scope for this fix. We don't give the editor any comment milestones in the USJ from core so we shouldn't hit this code path (it was an earlier idea for how to handle comments). Feel free to create a Jira work item if you like.
___
*[`libs/shared/src/nodes/usj/node.utils.ts` line 869 at r1](https://reviewable.io/reviews/eten-tech-foundation/scripture-editors/496#-OwxVDgI35JDRYfDAT1q:-Ox-GJ5L-0rv0WrD5cor:ba0fsyr) ([raw file](https://github.com/eten-tech-foundation/scripture-editors/blob/25207dd1aa03a8675e7420c13c13636eb999e70d/libs/shared/src/nodes/usj/node.utils.ts#L869)):*

NIT the other three `Logical*` types are grouped just above `$getLogicalContentItems`, but `LogicalPoint` is defined mid-file here. Near-first-use is defensible; grouping all four together would be marginally more consistent. Non-blocking.
</blockquote></details>

Done. Moved all to the top of file.
___
*[`libs/shared-react/src/plugins/usj/annotation/selection.utils.ts` line 410 at r1](https://reviewable.io/reviews/eten-tech-foundation/scripture-editors/496#-OwxH4HeAPqKazKJt-qq:-Ox-IxYJARxUrz75VclT:b-896fix) ([raw file](https://github.com/eten-tech-foundation/scripture-editors/blob/25207dd1aa03a8675e7420c13c13636eb999e70d/libs/shared-react/src/plugins/usj/annotation/selection.utils.ts#L410)):*

Agreed, this comment is helpful. Claude pointed out that it doesn't mention the case of when a cursor lands inside a mark that has several children of different kinds. `ts
const elementOffset = offset >= childrenSize ? markIndex + 1 : markIndex;` This doesn't place the point between the exact children it snaps to the front edge of the whole mark. So the reported position would be a little off. Recommending another sentence in the comment to describe.
</blockquote></details>

Done.

katherinejensen00 (approved):
Thanks for your improvements and responses! It looks good to me.

> @katherinejensen00 reviewed 2 files and all commit messages, made 1 comment, and resolved 2 discussions.
> *[Reviewable](https://reviewable.io/reviews/eten-tech-foundation/scripture-editors/496#-:-Ox1LpZG0gJTGEEEx4wE:b-xp0ht0)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved.

PR #497, opened 2026-07-07
PT-3835: Don't insert verse spacing after gutter para marker prefixes

irahopkinson:
# Code Review Summary

**Branch**: pt-3835-part2

**Base**: pt-3835-annotation-selection (this branch is stacked on the PT-3835 branch / PR #496; diffing against `origin/main` would re-include the parent PR's already-reviewed changes)

**Date**: 2026-07-07

**Review model**: Claude Fable 5

**Files changed**: 3

## Overview

Follow-up fix for PT-3835, found while verifying the annotation-transparency fix in paranext-core. With gutter para markers enabled (formatted view), a paragraph's first child is a `\p` marker prefix node that `TextSpacingPlugin`'s `$verseNodeTransform` did not recognize, so the first editor update inserted a spurious single-space TextNode before verse 1. That space is real content to both the logical content model and the editor→USJ export, shifting every content index in the paragraph by one. In paranext-core's Insert Comment flow this made the thread annotation fail to re-apply after the pending annotation was removed ("Failed to find start or end node of the annotation.") and left every subsequent reported selection off by one, tripping the host's marker-validation guard. The same latent bug existed in plain visible-marker mode.

The fix is a one-line guard: exclude para marker prefixes (`$isParaMarkerPrefix`) from the transform's insert-space condition — marker prefixes carry their own trailing spacing. The transform's remaining insertion paths are intentional: the author verified in Paratext 9 that a space between a note/char and a following verse marker is *structural* (USJ→USFM conversion needs it; P9 re-inserts it when removed), so only the marker-prefix case was spurious. Coverage: a unit regression in `TextSpacingPlugin.test.tsx`, plus a lifecycle suite (`annotation-pending-rewrap.test.tsx`) mounting the real platform Editor and running the exact host flow under both failing-mode view-option variants plus an all-off passing control, asserting post-rewrap selections still report coalesced-USJ coordinates. The author retested in paranext-core: working in both Genesis 1 and Genesis 2 scenarios.

## API Changes

None. No public API surfaces changed:

- `packages/platform` (`@eten-tech-foundation/platform-editor`, published): only a new test file was added — no source or export changes.
- `libs/shared-react` (internal lib): `TextSpacingPlugin.tsx` changed internally (one added condition inside the non-exported `$verseNodeTransform`); the exported `TextSpacingPlugin` component's signature and exports are unchanged.
- `libs/shared`: not modified.

None.

None.

### Minor — Consider

- [x] Missing in-code rationale for the `!$isParaMarkerPrefix` exclusion in `TextSpacingPlugin.tsx` (the "why" lived only in tests and the commit message) _(fixed during review: comment added above the condition — marker prefixes carry their own trailing NBSP, and an inserted plain space becomes exporter-visible USJ content that shifts content indices; folded into the amended commit)_
- [ ] ~~Test-helper duplication between `annotation-pending-rewrap.test.tsx` (`createRealEditor`, DOM-mark helpers) and `Editor.test.tsx` (`createEditorRefForTesting`, `getMarkElement`)~~ _(author dismissed: same pattern, not verbatim duplication — the new helpers' parameterization (view options, injected logger error-spy, all-marks text assertions) is load-bearing; extract a shared helper when a third editor-lifecycle test file appears)_
- [x] The commit message's "known follow-up" (transform still inserts `" "` when a verse follows a NoteNode/CharNode) was untracked in code _(fixed during review — and reclassified: the author tested in Paratext 9 that this space is structural, not content — USJ→USFM conversion requires it and P9 re-inserts it when removed. No follow-up ticket needed; the commit message was amended to document that the remaining insertion paths are intentional and only the marker-prefix case was spurious)_
- [x] Process references in the lifecycle test's comments ("found by working the contingency ladder", "spec-mandated failing-mode variants") lacked context for future readers _(fixed during review: references trimmed; the boundary knowledge — minimal failing configuration and what does NOT reproduce — kept intact)_

[Author response: author requested fixes for three; the helper-duplication finding was dismissed with the extract-at-third-file rationale. The follow-up finding produced a genuine domain insight (structural vs content spaces) that reshaped the commit message rather than a ticket.]

#### Shared Regions Modified

None.

#### Extension Config Changes

None.

## Positive Observations

- The fix reuses the existing shared predicate `$isParaMarkerPrefix` rather than inlining node-type checks, covers both prefix flavors (editable `MarkerNode` and gutter/visible `ImmutableTypedTextNode`), and matches the transform's existing exclusion-list style exactly.
- The unit regression constructs the exact gutter-mode node shape the adaptor produces (`$createImmutableTypedTextNode("marker", openingMarkerText("p") + NBSP)`) rather than a stand-in, and is verifiably falsifiable (without the guard, the child count assertion fails).
- The lifecycle suite is exemplary regression engineering: it mounts the real `Editor` through the same `EditorRef` API paranext-core drives, includes a fixture-sanity test that distinguishes fixture bugs from code bugs, covers both failing-mode variants plus an all-off control, covers same-tick and separate-act sequencing (matching the app's single event handler), and pins the original defect's downstream symptom (post-rewrap selections elsewhere still report correct coalesced-USJ coordinates). The header documents the minimal failing configuration and what does NOT reproduce it.
- Both suites verified passing by multiple independent analyzers (18/18 unit, 7/7 lifecycle); RTL auto-cleanup prevents cross-test DOM contamination.
- Test placement and naming follow repo conventions; no localization, build-config, or shadcn surfaces touched.

## Interview Notes

- **Stated purpose**: from the branch commit message — stop `TextSpacingPlugin` inserting a spurious, index-shifting space after gutter para marker prefixes (PT-3835 follow-up found during core verification).
- **Key domain insight from the author**: when asked about tracking the "remaining insertion paths" as a follow-up bug, the author tested the behavior in Paratext 9 — deleting the space between a footnote/char and a following verse marker causes P9 to re-insert it, because that space is *structural* (required when converting USJ to USFM), not content. So the transform's remaining insertions are correct by design; only the marker-prefix case was spurious. The commit message was amended to record this.
- **Verification status**: the author retested in paranext-core — working in both Genesis 1 and Genesis 2 scenarios (the original 100%-reproducible failure is gone).
- **Open follow-up**: the paranext-core Playwright e2e tests (PT-3835 Definition of Done item) remain follow-up work in that repo.
- The author demonstrated clear understanding throughout — including independently testing conversion behavior in Paratext 9 to settle a finding. No uncertainty or AI-deferral observed.

## In-Review Quality Check

- `pnpm nx run-many -t typecheck`: all 10 projects pass.
- `pnpm nx lint shared-react` / `pnpm nx lint @eten-tech-foundation/platform-editor`: 0 errors.
- `pnpm nx format:check`: clean.
- Focused tests: `TextSpacingPlugin` 18/18, `annotation-pending-rewrap` 7/7.
- No auto-fixes needed; all in-review changes were comment-only. (Note: an earlier commit attempt surfaced that the repo still tracks `.claude/`/`CLAUDE.md` paths now replaced by ai-prompts symlinks on this machine, which broke lint-staged's backup — the author fixed the local state; a repo-level untrack+gitignore may be worth a separate change.)

## Suggested Review Focus

- [ ] The guard's blast radius: `$verseNodeTransform` no longer inserts a space after ANY para marker prefix in any mode — this also silently fixes the same latent bug in plain visible-marker mode. Confirm no visible-mode rendering depended on the old space (analysis found none: prefixes carry their own trailing NBSP).
- [ ] The structural-vs-content space distinction (author's Paratext 9 finding): spaces between note/char and a following verse are required for USJ→USFM conversion. Worth aligning on this as the documented rule for future TextSpacingPlugin work.
- [ ] The lifecycle suite (`annotation-pending-rewrap.test.tsx`) as a pattern: real-Editor-mount regression tests for host-flow bugs — consider whether future editor-level defects should get the same treatment.
- [ ] Stacking: this branch is based on `pt-3835-annotation-selection` (PR #496) — decide merge order / whether to fold into #496.
- [ ] Follow-up work: paranext-core Playwright e2e tests for the PT-3835 DoD.

- - -
This change is [Reviewable](https://reviewable.io/reviews/eten-tech-foundation/scripture-editors/497)

### DIFF
--- a/libs/shared-react/src/plugins/usj/TextSpacingPlugin.test.tsx
+++ b/libs/shared-react/src/plugins/usj/TextSpacingPlugin.test.tsx
@@ -16,6 +16,7 @@ import { $createTextNode, $getRoot, $isTextNode, TextNode, $setSelection } from 
 import {
   $createCharNode,
   $createImmutableChapterNode,
+  $createImmutableTypedTextNode,
   $createNoteNode,
   $createParaNode,
   $createTypedMarkNode,
@@ -24,6 +25,9 @@ import {
   $isParaNode,
   $isTypedMarkNode,
   $isUnknownNode,
+  $isVisibleMarkerNode,
+  NBSP,
+  openingMarkerText,
   ParaNode,
   UnknownNode,
 } from "shared";
@@ -360,6 +364,34 @@ describe("TextSpacingPlugin", () => {
       // Should remain [UnknownNode, VerseNode]
       expect(para.getChildren()).toHaveLength(2);
       expect($isUnknownNode(para.getChildAtIndex(0))).toBe(true);
+      expect($isSomeVerseNode(para.getChildAtIndex(1))).toBe(true);
+    });
+  });
+
+  it("should not insert a space before a verse if preceded by a gutter paragraph marker prefix", async () => {
+    // Regression test for PT-3835 Gen 2: gutter mode (`hasGutterParaMarkers: true`) renders the
+    // paragraph's `\p` marker as a visible-marker ImmutableTypedTextNode (textType "marker") that is
+    // verse 1's previous sibling. That node is a DecoratorNode, not a TextNode/UnknownNode, so
+    // $verseNodeTransform used to treat it like arbitrary unrecognized content and insert a spurious
+    // leading space, shifting every logical content index in the paragraph.
+    const { editor } = await testEnvironment(() => {
+      $getRoot().append(
+        $createParaNode().append(
+          $createImmutableTypedTextNode("marker", openingMarkerText("p") + NBSP),
+          $createImmutableVerseNode("1"),
+        ),
+      );
+    });
+
+    // Trigger an update (no change expected).
+    await act(async () => editor.update(() => undefined));
+
+    editor.getEditorState().read(() => {
+      const para = $getRoot().getFirstChild();
+      if (!$isParaNode(para)) throw new Error("Expected a ParaNode");
+      // Should remain [marker prefix, VerseNode] -- no spurious space inserted.
+      expect(para.getChildren()).toHaveLength(2);
+      expect($isVisibleMarkerNode(para.getChildAtIndex(0))).toBe(true);
       expect($isSomeVerseNode(para.getChildAtIndex(1))).toBe(true);
     });
   });
--- a/libs/shared-react/src/plugins/usj/TextSpacingPlugin.test.tsx
+++ b/libs/shared-react/src/plugins/usj/TextSpacingPlugin.test.tsx
@@ -21,6 +21,7 @@ import {
   $createParaNode,
   $createTypedMarkNode,
   $createUnknownNode,
+  $getLogicalContentItems,
   $isCharNode,
   $isParaNode,
   $isTypedMarkNode,
@@ -393,6 +394,91 @@ describe("TextSpacingPlugin", () => {
       expect(para.getChildren()).toHaveLength(2);
       expect($isVisibleMarkerNode(para.getChildAtIndex(0))).toBe(true);
       expect($isSomeVerseNode(para.getChildAtIndex(1))).toBe(true);
+    });
+  });
+
+  it("should space an annotation over plain text before a verse without shifting its logical index", async () => {
+    // The annotation wrapper is transparent to USJ content: the inserted space coalesces onto
+    // the wrapped text's run (the trailing-space transform can't reach text inside a
+    // TypedMarkNode), so the verse's logical content index is unchanged.
+    const { editor } = await testEnvironment(() => {
+      $getRoot().append(
+        $createParaNode().append(
+          $createImmutableVerseNode("1"),
+          $createTextNode("the "),
+          $createTypedMarkNode({ t: ["1"] }).append($createTextNode("beginning")),
+          $createImmutableVerseNode("2"),
+        ),
+      );
+    });
+
+    // Trigger an update (transforms run).
+    await act(async () => editor.update(() => undefined));
+
+    editor.getEditorState().read(() => {
+      const para = $getRoot().getFirstChild();
+      if (!$isParaNode(para)) throw new Error("Expected a ParaNode");
+      // [verse 1, "the beginning ", verse 2] — space coalesced into the run, no index shift.
+      expect($getLogicalContentItems(para)).toHaveLength(3);
+      expect(para.getTextContent()).toBe("the beginning ");
+    });
+  });
+
+  it("should insert the structural space when an annotation ending on a CharNode precedes a verse", async () => {
+    // A space between a char and a following verse marker is structural, not content: USJ→USFM
+    // conversion needs it and Paratext 9 re-inserts it when removed. Canonical USJ therefore has
+    // a standalone " " item here, so inserting it matches the exported shape — the annotation
+    // wrapper must not suppress it.
+    const { editor } = await testEnvironment(() => {
+      $getRoot().append(
+        $createParaNode().append(
+          $createImmutableVerseNode("1"),
+          $createTextNode("text "),
+          $createTypedMarkNode({ t: ["1"] }).append(
+            $createCharNode("nd").append($createTextNode("LORD")),
+          ),
+          $createImmutableVerseNode("2"),
+        ),
+      );
+    });
+
+    // Trigger an update (transforms run).
+    await act(async () => editor.update(() => undefined));
+
+    editor.getEditorState().read(() => {
+      const para = $getRoot().getFirstChild();
+      if (!$isParaNode(para)) throw new Error("Expected a ParaNode");
+      // [verse 1, "text ", char, " ", verse 2] — the structural space is its own content item,
+      // exactly as canonical USJ from Paratext has it.
+      const items = $getLogicalContentItems(para);
+      expect(items).toHaveLength(5);
+      expect(items[3].type).toBe("text");
+    });
+  });
+
+  it("should not insert a space before a verse for an empty annotation wrapper", async () => {
+    // An empty TypedMarkNode resolves to no content, so no structural space belongs after it —
+    // inserting one would add exporter-visible USJ content because of a presentation-only node.
+    const { editor } = await testEnvironment(() => {
+      $getRoot().append(
+        $createParaNode().append(
+          $createImmutableVerseNode("1"),
+          $createTextNode("a "),
+          $createTypedMarkNode({ t: ["1"] }),
+          $createImmutableVerseNode("2"),
+        ),
+      );
+    });
+
+    // Trigger an update (transforms run).
+    await act(async () => editor.update(() => undefined));
+
+    editor.getEditorState().read(() => {
+      const para = $getRoot().getFirstChild();
+      if (!$isParaNode(para)) throw new Error("Expected a ParaNode");
+      // [verse 1, "a ", verse 2] — no double space, no extra content item.
+      expect($getLogicalContentItems(para)).toHaveLength(3);
+      expect(para.getTextContent()).toBe("a ");
     });
   });
 
--- a/libs/shared-react/src/plugins/usj/TextSpacingPlugin.tsx
+++ b/libs/shared-react/src/plugins/usj/TextSpacingPlugin.tsx
@@ -13,6 +13,7 @@ import {
   $isCharNode,
   $isNoteNode,
   $isParaLikeNode,
+  $isParaMarkerPrefix,
   $isTypedMarkNode,
   $isUnknownNode,
   CharNode,
@@ -116,7 +117,11 @@ function $verseNodeTransform(node: SomeVerseNode): void {
     previousSibling &&
     !$isSomeVerseNode(previousSibling) &&
     !$isTextNode(previousSibling) &&
-    !$isUnknownNode(previousSibling)
+    !$isUnknownNode(previousSibling) &&
+    // Para marker prefixes carry their own trailing spacing (NBSP baked into the prefix text),
+    // and an inserted plain " " would be exporter-visible USJ content that shifts every content
+    // index in the paragraph (see PT-3835).
+    !$isParaMarkerPrefix(previousSibling)
   )
     node.insertBefore($createTextNode(" "));
 }
--- a/libs/shared-react/src/plugins/usj/TextSpacingPlugin.tsx
+++ b/libs/shared-react/src/plugins/usj/TextSpacingPlugin.tsx
@@ -7,7 +7,7 @@ import {
 } from "../../nodes/usj/node-react.utils";
 import { useLexicalComposerContext } from "@lexical/react/LexicalComposerContext";
 import { mergeRegister } from "@lexical/utils";
-import { $createTextNode, $isTextNode, LexicalEditor, TextNode } from "lexical";
+import { $createTextNode, $isTextNode, LexicalEditor, LexicalNode, TextNode } from "lexical";
 import { useEffect } from "react";
 import {
   $isCharNode,
@@ -112,16 +112,26 @@ function $textNodeInUnknownTransform(node: TextNode, editor: LexicalEditor): voi
 function $verseNodeTransform(node: SomeVerseNode): void {
   if (!node.isAttached()) return;
 
-  const previousSibling = node.getPreviousSibling();
+  // Annotation wrappers (TypedMarkNode) are presentation-only, so the spacing decision depends
+  // on the content INSIDE them: resolve to the wrapper's last content child (nested wrappers
+  // are transient but resolved for safety). An empty wrapper resolves to null and inserts
+  // nothing.
+  let previousSibling: LexicalNode | null = node.getPreviousSibling();
+  while ($isTypedMarkNode(previousSibling)) previousSibling = previousSibling.getLastChild();
+
   if (
     previousSibling &&
     !$isSomeVerseNode(previousSibling) &&
-    !$isTextNode(previousSibling) &&
     !$isUnknownNode(previousSibling) &&
-    // Para marker prefixes carry their own trailing spacing (NBSP baked into the prefix text),
-    // and an inserted plain " " would be exporter-visible USJ content that shifts every content
-    // index in the paragraph (see PT-3835).
-    !$isParaMarkerPrefix(previousSibling)
+    // Para marker prefixes are presentation scaffolding, not USJ content, so no structural
+    // space belongs after them — and an inserted plain " " would be exporter-visible USJ
+    // content that shifts every content index in the paragraph (see PT-3835). Their visual
+    // separation comes from the prefix nodes' own text.
+    !$isParaMarkerPrefix(previousSibling) &&
+    // Bare text before a verse gets its structural space from $textNodeTrailingSpaceTransform;
+    // text inside an annotation wrapper can't (that transform skips TypedMarkNode parents), so
+    // the space is inserted here instead and coalesces onto the same USJ text run.
+    (!$isTextNode(previousSibling) || $isTypedMarkNode(previousSibling.getParent()))
   )
     node.insertBefore($createTextNode(" "));
 }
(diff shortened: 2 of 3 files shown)


jolierabideau (dismissed):
Thanks Ira, one small NIT from Claude.

> @jolierabideau reviewed 3 files and all commit messages, and made 2 comments.
> *[Reviewable](https://reviewable.io/reviews/eten-tech-foundation/scripture-editors/497#-:-OwxZneg310OD7xriNaf:b-pdziky)* status: all files reviewed, 1 unresolved discussion (waiting on irahopkinson).

___
*[`libs/shared-react/src/plugins/usj/TextSpacingPlugin.tsx` line 121 at r1](https://reviewable.io/reviews/eten-tech-foundation/scripture-editors/497#-OwxZWt0CHrMtHrdIieC:-OwxZWt0CHrMtHrdIieD:b-bejrir) ([raw file](https://github.com/eten-tech-foundation/scripture-editors/blob/0d9de5f0d7555d5bc4574b941dce14c65dfcfac8/libs/shared-react/src/plugins/usj/TextSpacingPlugin.tsx#L121)):*
> ```XML
>     !$isTextNode(previousSibling) &&
>     !$isUnknownNode(previousSibling) &&
>     // Para marker prefixes carry their own trailing spacing (NBSP baked into the prefix text),
> ```

NIT From Claude -

_"Para marker prefixes carry their own trailing spacing (NBSP baked into the prefix text)"_ isn't quite right for either flavor. An editable `MarkerNode` is just `\marker` with no trailing spacing (`openingMarkerText`), and the visible marker's trailing separator is a **plain space** added by `getVisibleOpenMarkerText` (the NBSP there sits _between_ the marker and its content, not at the end). The guard itself is correct — the real justification is the next clause: an inserted plain `" "` becomes exporter-visible USJ content and shifts every downstream content index. Since the guard fires for both marker modes (`$isParaMarkerPrefix` = `$isMarkerNode || $isVisibleMarkerNode`), I'd just reword so a future reader doesn't rely on a trailing NBSP that isn't there. Visual separation still holds in visible mode via that trailing plain space.

katherinejensen00 (dismissed):
Thanks for continuing this fix, Ira! Just a couple of small things to consider. Approving because this could be merged as is and be good, there just might be a couple small improvements.

> @katherinejensen00 reviewed 3 files and all commit messages, and made 3 comments.
> *[Reviewable](https://reviewable.io/reviews/eten-tech-foundation/scripture-editors/497#-:-Owxcg-q-Bm8GF0vXM92:bsor96d)* status: all files reviewed, 3 unresolved discussions (waiting on irahopkinson).

___
*[`libs/shared-react/src/plugins/usj/TextSpacingPlugin.tsx` line 124 at r1](https://reviewable.io/reviews/eten-tech-foundation/scripture-editors/497#-OwxcXwG95wXYo_aO6Uj:-OwxcXwG95wXYo_aO6Uk:b-8ing5c) ([raw file](https://github.com/eten-tech-foundation/scripture-editors/blob/0d9de5f0d7555d5bc4574b941dce14c65dfcfac8/libs/shared-react/src/plugins/usj/TextSpacingPlugin.tsx#L124)):*
> ```XML
>     // and an inserted plain " " would be exporter-visible USJ content that shifts every content
>     // index in the paragraph (see PT-3835).
>     !$isParaMarkerPrefix(previousSibling)
> ```

NIT Claude Finding

The fix is right (confirmed `MarkerNode extends TextNode`, so this correctly targets the gutter/visible `ImmutableTypedTextNode` prefix). I went down a path worrying the same shift was open for any `TypedMarkNode` predecessor, but I traced it through `$getLogicalContentItems` and the _common_ case doesn't reproduce: when an annotation wraps plain text before a verse, the inserted `" "` coalesces onto the existing text run, so the verse's logical index is unchanged. No off-by-one there.

The one case that _does_ still shift is narrow: an annotation whose trailing content is a non-text element (`CharNode`/`NoteNode`) sitting immediately before a verse — e.g. a verse ending in `\nd LORD` that gets annotated through. There the `CharNode` breaks the coalesced run, so the inserted space becomes a standalone content item and pushes the verse index by one, the same way the para-marker prefix did. Reachable in principle, but edge-y, so I'd treat it as hardening rather than a live bug.

Two ways to close it if you think it's worth it: add `!$isTypedMarkNode(previousSibling)` to this same condition (one line, no behavior change for anything else), or — the design angle — note that "which nodes are transparent to USJ content" is now encoded both here and in #496's `$shouldIgnoreNodeForContentIndexes`, and the two can drift as node types are added (this `TypedMarkNode` gap is the first instance). Deriving the exclusion from a shared predicate, or inverting to an allowlist of nodes that genuinely need a structural space (the `NoteNode`/`CharNode` cases you validated in P9), would make new transparent types default to safe. Not a blocker — the current fix is correct for what shipped.

I have a two-case unit test that pins exactly which scenario shifts — the first passes today, the second fails today (returns 5 instead of 4) and would pass with the `!$isTypedMarkNode` guard, so it documents both the boundary and the fix. Happy to fold it into `TextSpacingPlugin.test.tsx` if you'd like the edge case captured:

```tsx
it("annotation over plain text before a verse does NOT shift the verse's logical index", async () => {
  const { editor } = await testEnvironment(() => {
    $getRoot().append(
      $createParaNode().append(
        $createImmutableVerseNode("1"),
        $createTextNode("the "),
        $createTypedMarkNode({ t: ["1"] }).append($createTextNode("beginning")),
        $createImmutableVerseNode("2"),
      ),
    );
  });
  await act(async () => editor.update(() => undefined));
  editor.getEditorState().read(() => {
    const para = $getRoot().getFirstChild();
    if (!$isParaNode(para)) throw new Error("Expected a ParaNode");
    // Space coalesces into "the beginning " → verse 2 stays at logical index 2.
    expect($getLogicalContentItems(para)).toHaveLength(3);
  });
});

it("annotation ending on a CharNode before a verse DOES shift (the guard's real target)", async () => {
  const { editor } = await testEnvironment(() => {
    $getRoot().append(
      $createParaNode().append(
        $createImmutableVerseNode("1"),
        $createTextNode("text "),
        $createTypedMarkNode({ t: ["1"] }).append(
          $createCharNode("nd").append($createTextNode("LORD")),
        ),
        $createImmutableVerseNode("2"),
      ),
    );
  });
  await act(async () => editor.update(() => undefined));
  editor.getEditorState().read(() => {
    const para = $getRoot().getFirstChild();
    if (!$isParaNode(para)) throw new Error("Expected a ParaNode");
    // CharNode breaks the run → inserted space is a standalone item → verse 2 shifts to index 4.
    // Fails today (returns 5); passes with `!$isTypedMarkNode` added to $verseNodeTransform.
    expect($getLogicalContentItems(para)).toHaveLength(4);
  });
});
```

(Needs `$getLogicalContentItems` imported from `shared`; the other factories are already imported in the test file.)
___
*[`packages/platform/src/editor/annotation-pending-rewrap.test.tsx` line 8 at r1](https://reviewable.io/reviews/eten-tech-foundation/scripture-editors/497#-OwxbjdG3bUHy8XDBsUv:-OwxbjdG3bUHy8XDBsUw:b8g3wx4) ([raw file](https://github.com/eten-tech-foundation/scripture-editors/blob/0d9de5f0d7555d5bc4574b941dce14c65dfcfac8/packages/platform/src/editor/annotation-pending-rewrap.test.tsx#L8)):*
> ```XML
>  * "Failed to find start or end node of the annotation."
>  *
>  * MINIMAL FAILING CONFIGURATION: the failure needs the REAL platform Editor (full plugin
> ```

NIT 
This is a dense comment with a lot of good information. I wonder if it might be more readable if there is a summary or a TLDR section at the top followed by the more detailed explanation. Here is a potential summary:
```
 * Reproduces ONLY with: real platform Editor + formatted view + hasGutterParaMarkers: true.
 * Does NOT reproduce with:
 *   - a minimal LexicalComposer mounting only AnnotationPlugin (even with identical view options)
 *   - the real Editor without the gutter option
 *   - showCharMarkerTitles / hasActiveTextFocusBox alone (see the passing control)
```

irahopkinson (commented):
I've already used more than the time I have for coding in this sprint. However I have addressed the 3 comments. For further work either approve as is and add your own follow-up PR or take over the branch and fix yourself.

> @irahopkinson made 4 comments and resolved 3 discussions.
> *[Reviewable](https://reviewable.io/reviews/eten-tech-foundation/scripture-editors/497#-:-Ox-FV1cAmRW0u8db2qA:b-h5ch3y)* status: 0 of 3 files reviewed, all discussions resolved (waiting on jolierabideau and katherinejensen00).

___
*[`libs/shared-react/src/plugins/usj/TextSpacingPlugin.tsx` line 121 at r1](https://reviewable.io/reviews/eten-tech-foundation/scripture-editors/497#-OwxZWt0CHrMtHrdIieC:-Ox-RGrG4v6b82tJG5Wc:bajpptt) ([raw file](https://github.com/eten-tech-foundation/scripture-editors/blob/0d9de5f0d7555d5bc4574b941dce14c65dfcfac8/libs/shared-react/src/plugins/usj/TextSpacingPlugin.tsx#L121)):*

NIT From Claude -

_"Para marker prefixes carry their own trailing spacing (NBSP baked into the prefix text)"_ isn't quite right for either flavor. An editable `MarkerNode` is just `\marker` with no trailing spacing (`openingMarkerText`), and the visible marker's trailing separator is a **plain space** added by `getVisibleOpenMarkerText` (the NBSP there sits _between_ the marker and its content, not at the end). The guard itself is correct — the real justification is the next clause: an inserted plain `" "` becomes exporter-visible USJ content and shifts every downstream content index. Since the guard fires for both marker modes (`$isParaMarkerPrefix` = `$isMarkerNode || $isVisibleMarkerNode`), I'd just reword so a future reader doesn't rely on a trailing NBSP that isn't there. Visual separation still holds in visible mode via that trailing plain space.
</blockquote></details>

Done. Good catch — the NBSP claim was wrong as written (editable `MarkerNode` is bare `\marker` with its trailing space in a separate node, and the visible marker's trailing separator is a plain space). Reworded: the comment now leads with the real justification — marker prefixes aren't USJ content, so no structural space belongs after them, and an inserted plain `" "` becomes exporter-visible content that shifts every content index.
___
*[`libs/shared-react/src/plugins/usj/TextSpacingPlugin.tsx` line 124 at r1](https://reviewable.io/reviews/eten-tech-foundation/scripture-editors/497#-OwxcXwG95wXYo_aO6Uj:-Ox-RaT04sY_71yZn2A3:b7vf29g) ([raw file](https://github.com/eten-tech-foundation/scripture-editors/blob/0d9de5f0d7555d5bc4574b941dce14c65dfcfac8/libs/shared-react/src/plugins/usj/TextSpacingPlugin.tsx#L124)):*

NIT Claude Finding

The fix is right (confirmed `MarkerNode extends TextNode`, so this correctly targets the gutter/visible `ImmutableTypedTextNode` prefix). I went down a path worrying the same shift was open for any `TypedMarkNode` predecessor, but I traced it through `$getLogicalContentItems` and the _common_ case doesn't reproduce: when an annotation wraps plain text before a verse, the inserted `" "` coalesces onto the existing text run, so the verse's logical index is unchanged. No off-by-one there.

The one case that _does_ still shift is narrow: an annotation whose trailing content is a non-text element (`CharNode`/`NoteNode`) sitting immediately before a verse — e.g. a verse ending in `\nd LORD` that gets annotated through. There the `CharNode` breaks the coalesced run, so the inserted space becomes a standalone content item and pushes the verse index by one, the same way the para-marker prefix did. Reachable in principle, but edge-y, so I'd treat it as hardening rather than a live bug.

Two ways to close it if you think it's worth it: add `!$isTypedMarkNode(previousSibling)` to this same condition (one line, no behavior change for anything else), or — the design angle — note that "which nodes are transparent to USJ content" is now encoded both here and in #496's `$shouldIgnoreNodeForContentIndexes`, and the two can drift as node types are added (this `TypedMarkNode` gap is the first instance). Deriving the exclusion from a shared predicate, or inverting to an allowlist of nodes that genuinely need a structural space (the `NoteNode`/`CharNode` cases you validated in P9), would make new transparent types default to safe. Not a blocker — the current fix is correct for what shipped.

I have a two-case unit test that pins exactly which scenario shifts — the first passes today, the second fails today (returns 5 instead of 4) and would pass with the `!$isTypedMarkNode` guard, so it documents both the boundary and the fix. Happy to fold it into `TextSpacingPlugin.test.tsx` if you'd like the edge case captured:

```tsx
it("annotation over plain text before a verse does NOT shift the verse's logical index", async () => {
  const { editor } = await testEnvironment(() => {
    $getRoot().append(
      $createParaNode().append(
        $createImmutableVerseNode("1"),
        $createTextNode("the "),
        $createTypedMarkNode({ t: ["1"] }).append($createTextNode("beginning")),
        $createImmutableVerseNode("2"),
      ),
    );
  });
  await act(async () => editor.update(() => undefined));
  editor.getEditorState().read(() => {
    const para = $getRoot().getFirstChild();
    if (!$isParaNode(para)) throw new Error("Expected a ParaNode");
    // Space coalesces into "the beginning " → verse 2 stays at logical index 2.
    expect($getLogicalContentItems(para)).toHaveLength(3);
  });
});

it("annotation ending on a CharNode before a verse DOES shift (the guard's real target)", async () => {
  const { editor } = await testEnvironment(() => {
    $getRoot().append(
      $createParaNode().append(
        $createImmutableVerseNode("1"),
        $createTextNode("text "),
        $createTypedMarkNode({ t: ["1"] }).append(
          $createCharNode("nd").append($createTextNode("LORD")),
        ),
        $createImmutableVerseNode("2"),
      ),
    );
  });
  await act(async () => editor.update(() => undefined));
  editor.getEditorState().read(() => {
    const para = $getRoot().getFirstChild();
    if (!$isParaNode(para)) throw new Error("Expected a ParaNode");
    // CharNode breaks the run → inserted space is a standalone item → verse 2 shifts to index 4.
    // Fails today (returns 5); passes with `!$isTypedMarkNode` added to $verseNodeTransform.
    expect($getLogicalContentItems(para)).toHaveLength(4);
  });
});
```

(Needs `$getLogicalContentItems` imported from `shared`; the other factories are already imported in the test file.)
</blockquote></details>

Done. Thanks for the careful trace — the coalescing analysis for the plain-text case is spot on, and your two-case test made the boundary really clear.

I went with a variant of your suggestion: instead of `!$isTypedMarkNode(previousSibling)`, the transform now resolves the wrapper to its last content child and decides from that. Reason: I tested the char/note-before-verse case in Paratext 9 — if you delete the space between them, P9 puts it back. That space is structural (USJ→USFM conversion needs it), so canonical USJ genuinely has the standalone `" "` item there, and inserting it matches the exported shape rather than shifting it. A blanket `!$isTypedMarkNode` guard would have made annotated content export differently from the same content unannotated. So:
 - wrapper ending in text → insert (coalesces onto the run, no index shift — your case 1, still 3 items)
 - wrapper ending in char/note → insert (canonical structural space — your case 2, pinned at 5 items with a comment explaining why)
 - empty wrapper → skip (the actual behavior fix; previously it added an exporter-visible space for a presentation-only node)

All three are now pinned in `TextSpacingPlugin.test.tsx` (your tests, adapted — thank you!). On the design point about "which nodes are USJ-transparent" living in two places: agreed it can drift; I'd like to keep that as a follow-up rather than grow this PR — the look-through keeps this transform aligned with the model's transparency rule for now.
___
*[`packages/platform/src/editor/annotation-pending-rewrap.test.tsx` line 8 at r1](https://reviewable.io/reviews/eten-tech-foundation/scripture-editors/497#-OwxbjdG3bUHy8XDBsUv:-Ox-SabE1vYxr9oP83Xx:b3ttnjn) ([raw file](https://github.com/eten-tech-foundation/scripture-editors/blob/0d9de5f0d7555d5bc4574b941dce14c65dfcfac8/packages/platform/src/editor/annotation-pending-rewrap.test.tsx#L8)):*

NIT 
This is a dense comment with a lot of good information. I wonder if it might be more readable if there is a summary or a TLDR section at the top followed by the more detailed explanation. Here is a potential summary:
```
 * Reproduces ONLY with: real platform Editor + formatted view + hasGutterParaMarkers: true.
 * Does NOT reproduce with:
 *   - a minimal LexicalComposer mounting only AnnotationPlugin (even with identical view options)
 *   - the real Editor without the gutter option
 *   - showCharMarkerTitles / hasActiveTextFocusBox alone (see the passing control)
```
</blockquote></details>

Done. Agreed — added your TLDR block at the top verbatim (reproduces-only-with / does-not-reproduce-with), with the prose detail kept below it. 2a0f7f7e.

katherinejensen00 (approved):
> @katherinejensen00 reviewed 11 files and all commit messages.
> *[Reviewable](https://reviewable.io/reviews/eten-tech-foundation/scripture-editors/497)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved.

PR #498, opened 2026-07-07
platform: stop stale book echo when the host navigates across books

tjcouch-sil:
## Problem

When a host application (Platform.Bible) navigates the scripture reference across a book boundary — e.g. NUM 1 → "previous chapter" → LEV 27:1 — the editor echoes a corrupted reference back through `onScrRefChange`: the OLD book with the NEW chapter/verse (`NUM 27:1`). The host's navigation is overwritten and the user lands on chapter 27 of the wrong book.

## What broke, and where

#457 added the BookNode **book-sync listener** (`skipInitialization: false`) to `ScriptureReferencePlugin`, so the ref's `book` follows the loaded document — a legitimate and still-desired behavior. That PR also added `scrRefRef`/`onScrRefChangeRef` refs specifically so the listener could read current values **without re-registering** (its comment says it should "only run when BookNodes are created/updated").

However, the listener was registered inside the pre-existing `useEffect` keyed on `[editor, chapterNum, verseNum]`. Every chapter/verse-only navigation therefore tears down and re-registers the listener, and because it uses `skipInitialization: false`, each re-registration replays initialization against the **still-mounted old document** (the new book's content loads async in the host). During that window `scrRefRef.current` already holds the new ref (LEV 27:1) while the stale document's BookNode still reads `NUM` → the listener fires `onScrRefChange({ ...current, book: 'NUM' })` → NUM 27:1.

## Fix — preserving #457's intent

Split the two BookNode listeners into separate effects:

- The **book-sync listener** now registers once per editor (`[editor]` deps), reading current values through the refs #457 already added for exactly this purpose. Its mount-time initialization still performs the legitimate book correction #457 introduced, and it still reacts to genuine BookNode created/updated mutations — it just no longer replays against a stale document on every chapter/verse navigation.
- The **cursor-move listener** keeps its original deps and behavior, byte-identical.

This completes the design #457's own refs and comment describe, rather than changing what the listener is for.

## Testing

- New regression test: ref moves to a different book while the old document is still loaded → `onScrRefChange` must NOT fire with the old book. Fails on the pre-fix code with exactly the production echo shape; passes with the fix.
- The three pre-existing book-sync tests (mount-time correction) pass **unmodified**; full platform-editor suite 116 passed / 3 skipped; whole-monorepo test/lint/typecheck/format clean.
- Live-verified in Platform.Bible via a local yalc build: NUM 1 → previous chapter lands and stays on LEV 27:1 (polled ~5s for late echoes); same-book navigation, in-editor verse-click ref updates, in-editor book jumps, and fresh-editor loads all behave.

AI-assisted (Claude Code)

- - -
This change is [Reviewable](https://reviewable.io/reviews/eten-tech-foundation/scripture-editors/498)

### DIFF
--- a/packages/platform/src/editor/ScriptureReferencePlugin.test.tsx
+++ b/packages/platform/src/editor/ScriptureReferencePlugin.test.tsx
@@ -90,7 +90,7 @@ describe("ScriptureReferencePlugin", () => {
       // async load). scrRef.book already matches the content at mount, so no sync fires yet.
       const { setScrRef } = await testEnvironment(scrRef, mockOnScrRefChange);
 
-      // Host navigates across books (e.g. NUM -> LEV): scrRef now targets a different book (and
+      // Host navigates across books (here GEN -> EXO): scrRef now targets a different book (and
       // a verse that still exists in the stale document, isolating this test to the book-sync
       // listener rather than the separate cursor-placement/BCV path), but the editor still
       // contains the OLD book's BookNode.
--- a/packages/platform/src/editor/ScriptureReferencePlugin.test.tsx
+++ b/packages/platform/src/editor/ScriptureReferencePlugin.test.tsx
@@ -83,6 +83,23 @@ describe("ScriptureReferencePlugin", () => {
 
       expect(mockOnScrRefChange).not.toHaveBeenCalled();
     });
+
+    it("should not echo the stale document's book when scrRef navigates to a different book while the old document is still loaded", async () => {
+      // Content stays on GEN (the old document hasn't been swapped in yet - simulates the
+      // window between a host navigating across books and the new book's USJ finishing its
+      // async load). scrRef.book already matches the content at mount, so no sync fires yet.
+      const { setScrRef } = await testEnvironment(scrRef, mockOnScrRefChange);
+
+      // Host navigates across books (e.g. NUM -> LEV): scrRef now targets a different book (and
+      // a verse that still exists in the stale document, isolating this test to the book-sync
+      // listener rather than the separate cursor-placement/BCV path), but the editor still
+      // contains the OLD book's BookNode.
+      await setScrRef({ book: "EXO", chapterNum: 1, verseNum: 2 });
+
+      // The plugin must not echo the stale document's book back to the host combined with the
+      // new chapter/verse - that would corrupt the navigation (e.g. EXO 1:2 -> GEN 1:2).
+      expect(mockOnScrRefChange).not.toHaveBeenCalled();
+    });
   });
 
   describe("Selection Change", () => {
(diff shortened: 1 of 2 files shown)


lyonsil (approved):
lgtm - This appears to fix the bug flagged by TJ without causing new bugs

lyonsil (commented):
> @lyonsil reviewed 2 files and all commit messages.
> *[Reviewable](https://reviewable.io/reviews/eten-tech-foundation/scripture-editors/498)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved.

katherinejensen00 (dismissed):
This bug fix looks really helpful. Thanks for working on this, TJ! Just one small optional comment clarity change. I will go ahead and approve this.

> @katherinejensen00 reviewed 2 files and all commit messages, and made 2 comments.
> *[Reviewable](https://reviewable.io/reviews/eten-tech-foundation/scripture-editors/498#-:-Ox1RvlD0Gp8DZYxXWHn:b-kve0ag)* status: all files reviewed, 1 unresolved discussion (waiting on tjcouch-sil).

___
*[`packages/platform/src/editor/ScriptureReferencePlugin.tsx` line 98 at r2](https://reviewable.io/reviews/eten-tech-foundation/scripture-editors/498#-Ox1Rrla-ARoJzh6wRm3:-Ox1Rrla-ARoJzh6wRm4:b-cwh6kt) ([raw file](https://github.com/eten-tech-foundation/scripture-editors/blob/be63e3bbece2e860e6a516d82c15da0bcd5310c4/packages/platform/src/editor/ScriptureReferencePlugin.tsx#L98)):*
NIT Claude Finding

> Small accuracy fix: this listener actually fires on _every_ non-`destroyed` BookNode mutation (including in-editor `"updated"` ones), not only mount + document swap. What keeps the stray firings harmless is the `code !== current.book` guard, not a limited firing set — worth wording it that way so nobody later relies on a firing-frequency invariant that doesn't hold.

**Suggested rewrite of the book-sync comment** (replacing lines 90–99), corrected and tightened:

```ts
  // Book node sync: correct scrRef.book to match the currently loaded document's BookNode.
  // Fires on every non-destroyed BookNode mutation (initial-mount replay + later swaps); the
  // `code !== current.book` guard below is what makes stray firings harmless. Registered once
  // per editor ([editor] only), NOT re-registered on chapterNum/verseNum change: the refs
  // already keep values fresh, and re-registering would replay skipInitialization: false against
  // the still-mounted (stale) document on every navigation, echoing the OLD book with the NEW
  // chapter/verse back to the host.
```

katherinejensen00 (approved):
Thanks for improving that comment, TJ! Looks good.

> @katherinejensen00 reviewed 1 file and all commit messages, made 1 comment, and resolved 1 discussion.
> *[Reviewable](https://reviewable.io/reviews/eten-tech-foundation/scripture-editors/498#-:-Ox2UN5fAzoPaFaKfldK:bq07iv6)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved.

PR #500, opened 2026-07-14
Clean up ScriptureReferencePlugin after PR #499

irahopkinson:
# Code Review Summary

**Branch**: cleanup

**Base**: origin/main

**Date**: 2026-07-14

**Review model**: Claude Sonnet 5

**Files changed**: 3

## Overview

This branch is a cleanup pass following PR #499 ("rebuild ScriptureReferencePlugin as a navigation state machine"), touching only `packages/platform/src/editor/Editor.tsx`, `ScriptureReferencePlugin.tsx`, and `ScriptureReferencePlugin.test.tsx`. It reorganizes helper functions to sit below the main plugin export (matching the convention used elsewhere in the package), renames a couple of helpers for clarity (`$moveCursorToVerseStart` → `$moveCaretToVerseStart`, `schedulePlacement` → `schedulePlacingCaretAtVerseStart`), switches the plugin from a `default` export to a named export, and replaces `editor.getEditorState().read()` with `editor.read()` at the two sites where that's the correct pattern. It also restores a chunk of pre-PR-#499 logic in `$moveCaretToVerseStart`, described in the commit message as restoring "affordance for more than one chapter in USJ."

That last item turned out to be the one substantive discussion point: three independent analysis passes, followed by a detailed manual trace (including reading `$findNextChapter`'s actual implementation and diffing all three historical versions of the function — pre-#499, PR #499, and this cleanup), confirmed the "restored" guard was behaviorally identical to the PR #499 code it replaced — it didn't deliver the stated affordance, just reordered computation. The author agreed and had this hunk reverted to the PR #499 form during the review (see Findings). A genuine but unrelated pre-existing edge-case bug in `$findNextChapter` was surfaced along the way (see Suggested Review Focus) but is not part of this diff and was intentionally left alone.

## API Changes

None. `ScriptureReferencePlugin.tsx` is an internal editor-implementation file, not re-exported through `packages/platform/src/index.ts`, and is not tracked by the package's `api-extractor` report (`packages/platform/etc/platform-editor.api.md`), which has no diff versus the merge base. `Editor.tsx` only changed one import line to match the new named export. No changes touch `libs/shared/`, `libs/shared-react/`, `packages/utilities/`, or any package's public surface.

None.

- [x] **The "restored" guard in `$moveCaretToVerseStart` (lines 375–389) didn't deliver its stated purpose** — it was logically/behaviorally equivalent to the simpler PR #499 form it replaced, confirmed via truth-table analysis and a three-way diff of pre-#499/PR-#499/cleanup versions of the function, including tracing `$findNextChapter`'s `isCurrentChapterAtFirstNode` argument through both branches. _(fixed during review: reverted to the PR #499 form — early `if (!chapterNode) return;` followed by a hardcoded `true` argument to `$findNextChapter`, removing the redundant combined condition and the wasted pre-guard computation.)_
- [x] **Stale references to the pre-rename function name `$moveCursorToVerseStart`** in the top-of-file "HOW THIS PLUGIN WORKS" doc comment (`ScriptureReferencePlugin.tsx`, under "SAFE (additive) changes") and in a comment in `ScriptureReferencePlugin.test.tsx`, left over after the rename to `$moveCaretToVerseStart` elsewhere in this same diff. _(fixed during review: both updated to `$moveCaretToVerseStart`.)_

Author confirmed the guard restoration was "probably more restore to match old code" than an intentional behavior fix, and asked a substantive follow-up question about `$findNextChapter`'s second argument before agreeing to revert it — see Interview Notes.

### Minor — Consider

None.

#### Shared Regions Modified

None. No `#region shared with <url>` markers in any changed file.

#### Extension Config Changes

Not applicable — this repository has no `extensions/`-directory/template-propagation convention.

## Positive Observations

- The reorg (helpers moved below the exported component) matches the established convention elsewhere in the package (`ArrowNavigationPlugin.tsx`, `NoteNodePlugin.tsx`, `ParaMarkerPrefixGuardPlugin.tsx`, `ActiveTextPlugin.tsx`, `TreeViewPlugin.tsx`).
- `default export` → named `export function ScriptureReferencePlugin` matches how other plugins in this codebase are exported/consumed; confirmed no other file imports the old default export.
- `editor.read(...)` replacing `editor.getEditorState().read(...)` was applied at exactly the two correct sites, and the one site that must *not* use this pattern (`$resolvePosition`, inline in the `SELECTION_CHANGE_COMMAND` handler) was correctly left alone, with its explanatory comment about why intact.
- Renames (`schedulePlacement` → `schedulePlacingCaretAtVerseStart`, `$moveCursorToVerseStart` → `$moveCaretToVerseStart`, "window" → "navigation window" in prose) were applied consistently everywhere except the two stale spots noted above (now fixed).
- All 37 existing tests in `ScriptureReferencePlugin.test.tsx` pass unchanged; `nx lint`/`nx typecheck` clean for the package — confirmed both before and after the in-review fixes.

## Interview Notes

Author's stated purpose (from the branch's commit message): "cleanup after PR #499" — reorg, doc clarity, renames, named export, `editor.read()` adoption, and restoring the multi-chapter guard.

The one substantive discussion was around the restored guard in `$moveCaretToVerseStart`. When asked whether it fixed a real observed bug or was a "restore to match old code" move, the author said it was probably the latter. When told the guard reorder appeared to be behaviorally inert, the author pushed back with a sharp, specific technical question — whether moving the return check to after computing `nextChapterNode` could matter given what `$findNextChapter`'s second argument does. That was a good challenge and led to tracing the actual implementations of `removeNodesBeforeNode` and `$findNextChapter`, and diffing all three historical versions of the function (pre-#499 commit `6bb03956`, PR #499 commit `b315cac9`, and this cleanup). That trace confirmed no behavioral difference in either branch of the guard, but surfaced a real, pre-existing edge case: when the current chapter has zero verses (immediately followed by another chapter node), `$findNextChapter`'s hardcoded `isCurrentChapterAtFirstNode=true` skips index 0 of the sliced node array and misses that next chapter node, causing verse lookup to leak into the following chapter's content. This bug exists identically in both PR #499 and pre-#499 code — it predates this diff and is unrelated to it, so it was intentionally not fixed here (see Suggested Review Focus for a follow-up).

Based on this, the author agreed to revert the guard hunk to the PR #499 form, and separately asked for the two stale `$moveCursorToVerseStart` references to be fixed. Both were done during the review (see Findings).

No unresolved items — the author engaged directly and substantively with the one non-trivial technical question raised, rather than deferring or expressing uncertainty. No areas where the author did not understand their own changes.

Separately: the branch's commit was amended and pushed to `origin/cleanup` mid-review by the author directly (not by any action taken during this review) — confirmed by the author and by the subagent that ran quality checks, which reported it took no `git commit`/`git commit --amend`/`git push` actions of its own.

## In-Review Quality Check

After the two in-review fixes (guard revert, stale-name fixes), a subagent ran the package's checks against `@eten-tech-foundation/platform-editor` (the Nx project for `packages/platform`):

- **Typecheck** (`nx typecheck`) — passed, no errors.
- **Lint** (`nx lint`) — passed clean, no auto-fix needed.
- **Format** (`prettier --write` on the two edited files) — both already correctly formatted, no changes.
- **Tests** (`nx test`) — 148 passed, 3 skipped (unrelated data-generation tests), 0 failed. All 37 tests in `ScriptureReferencePlugin.test.tsx` passed.

No unfixable failures. No issues found that need the author's attention.

## Suggested Review Focus

- [x] Confirm the guard revert to the PR #499 form (already applied during this review) reads correctly — it's a pure simplification with no behavior change.
- [x] Pre-existing edge-case bug surfaced during review (not part of this diff, not fixed here): in `$findNextChapter` (libs/shared/src/nodes/usj/node.utils.ts), a current chapter with zero verses immediately followed by another chapter node causes the hardcoded `isCurrentChapterAtFirstNode=true` skip-index-0 behavior to miss that next chapter node, letting verse lookup leak into the following chapter's content. Exists identically in both PR #499 and pre-#499 code. Worth a follow-up ticket/test if empty chapters are a real scenario in production USJ documents.
- [x] Sanity-check the two stale-name fixes (doc comment + test comment) for correctness — both were mechanical, low-risk changes.

- - -
This change is [Reviewable](https://reviewable.io/reviews/eten-tech-foundation/scripture-editors/500)

### DIFF
--- a/packages/platform/src/editor/ScriptureReferencePlugin.tsx
+++ b/packages/platform/src/editor/ScriptureReferencePlugin.tsx
@@ -151,8 +151,6 @@ interface ResolvedPosition {
   verse: string | undefined;
 }
 
-const MAX_PENDING_ECHOES = 8;
-
 /**
  * A component (plugin) that keeps the Scripture reference updated.
  * @param scrRef - Scripture reference.
@@ -200,8 +198,8 @@ export function ScriptureReferencePlugin({
         (nodeMutations) => {
           const kinds = [...nodeMutations.values()];
           if (kinds.every((kind) => kind === "destroyed")) return;
-          const code = getCommittedBookCode(editor);
-          onDocumentChanged(machineRef.current, editor, code, {
+          const bookCode = getCommittedBookCode(editor);
+          onDocumentChanged(machineRef.current, editor, bookCode, {
             hasCreated: kinds.includes("created"),
             hasDestroyed: kinds.includes("destroyed"),
           });
@@ -218,6 +216,8 @@ export function ScriptureReferencePlugin({
         SELECTION_CHANGE_COMMAND,
         () => {
           const machine = machineRef.current;
+          // $resolvePosition MUST be called inline here, in the command's active editor-state
+          // context - see onSelectionSettled's doc before folding it into the helper.
           if (machine.phase === "idle") onSelectionSettled(machine, $resolvePosition());
           return false;
         },
@@ -322,52 +322,50 @@ function getCommittedBookCode(editor: LexicalEditor): string | undefined {
 
 /** First BookNode of the root (documents put it first; strays after content are ignored). */
 function $getFirstBookNode(): BookNode | undefined {
-  let node = $getRoot().getFirstChild();
-  while (node) {
-    if ($isBookNode(node)) return node;
-    node = node.getNextSibling();
-  }
-  return undefined;
+  return $getRoot().getChildren().find($isBookNode);
 }
 
 /** `documentChanged`: the loaded document's book identity was established or changed. */
 function onDocumentChanged(
   machine: Machine,
   editor: LexicalEditor,
-  code: string | undefined,
+  bookCode: string | undefined,
   batch: { hasCreated: boolean; hasDestroyed: boolean },
 ) {
+  // Captured before the write below so the MOUNT branch (b) sees the pre-batch value.
+  const isFirstDocument = batch.hasCreated && !machine.sawDocument;
+  if (batch.hasCreated) machine.sawDocument = true;
+
   if (machine.phase === "navigating") {
     // Silence - except the arrival of the document the navigation is waiting for.
-    if (batch.hasCreated && code && code === machine.scrRef.book) {
+    if (batch.hasCreated && bookCode && bookCode === machine.scrRef.book) {
       schedulePlacingCaretAtVerseStart(machine, editor);
     }
-    if (batch.hasCreated) machine.sawDocument = true;
     return;
   }
 
   if (batch.hasCreated && batch.hasDestroyed) {
-    // REPLACEMENT (LoadStatePlugin's setEditorState swap, in-editor book jump): the swap's own
+    // (a) REPLACEMENT (LoadStatePlugin's setEditorState swap, in-editor book jump): the swap's own
     // synthetic selection settle fires before the deferred placement and must be silenced -
     // regardless of book match, since a same-book reload's transient chapter-top settle is the
     // R2 clobber this branch exists to prevent. The correction below (d), if any, runs after.
     schedulePlacingCaretAtVerseStart(machine, editor);
     machine.phase = "navigating";
-  } else if (batch.hasCreated && !machine.sawDocument) {
-    // MOUNT: fresh editor, no navigation window needed - a null selection or one already at scrRef.
+  } else if (isFirstDocument) {
+    // (b) MOUNT: fresh editor, no navigation window needed - a null selection or one already at
+    // scrRef.
     schedulePlacingCaretAtVerseStart(machine, editor);
   }
-  // else: a later pure-created BookNode (cross-editor paste, undo recreating a BookNode) while a
-  // document is already loaded - neither placement nor a navigation window. A paste must not
+  // (c) else: a later pure-created BookNode (cross-editor paste, undo recreating a BookNode) while
+  // a document is already loaded - neither placement nor a navigation window. A paste must not
   // teleport the caret to verse start.
-  if (batch.hasCreated) machine.sawDocument = true;
 
-  // Book correction: "keep your position, fix the book". Emitting one opens a navigation window
-  // (the swap's own synthetic selection settle fires before the deferred placement and must be
-  // silenced); the correction's echo closes it. Runs after (a)-(c); under (a) phase is already
+  // (d) Book correction: "keep your position, fix the book". Emitting one opens a navigation
+  // window (the swap's own synthetic selection settle fires before the deferred placement and must
+  // be silenced); the correction's echo closes it. Runs after (a)-(c); under (a) phase is already
   // "navigating" here, so this assignment is harmless.
-  if (code && code !== machine.scrRef.book) {
-    if (report(machine, { ...machine.scrRef, book: code })) machine.phase = "navigating";
+  if (bookCode && bookCode !== machine.scrRef.book) {
+    if (report(machine, { ...machine.scrRef, book: bookCode })) machine.phase = "navigating";
   }
 }
 
@@ -385,6 +383,8 @@ function schedulePlacingCaretAtVerseStart(machine: Machine, editor: LexicalEdito
   });
 }
 
+/** Moves the caret to the start of `verseNum` in `chapterNum`. No-op when the caret is already
+ * inside a verse range containing `verseNum` (a range is one location), or the target is absent. */
 function $moveCaretToVerseStart(chapterNum: number, verseNum: number) {
   const startNode = getSelectionStartNode($getSelection());
   const selectedVerse = $findThisVerse(startNode)?.getNumber();
@@ -467,6 +467,10 @@ function positionToScrRef(
   if (hostRef.versificationStr != null) ref.versificationStr = hostRef.versificationStr;
   return ref;
 }
+
+/** Bound on the pendingEchoes FIFO: an echo that hasn't returned within this many subsequent
+ * emissions is presumed lost and dropped (oldest first). */
+const MAX_PENDING_ECHOES = 8;
 
 /** The only caller of `onScrRefChange`. Drops emissions equal to the current prop or to a
  * pending echo. Returns whether an emission actually happened. */
--- a/packages/platform/src/editor/ScriptureReferencePlugin.tsx
+++ b/packages/platform/src/editor/ScriptureReferencePlugin.tsx
@@ -1,3 +1,89 @@
+/*
+ * ============================================================================================
+ * HOW THIS PLUGIN WORKS - READ BEFORE CHANGING ANYTHING
+ * ============================================================================================
+ *
+ * Contract: the host owns `scrRef`; this plugin (a) applies scrRef changes to the caret and
+ * (b) reports genuine user caret moves back via `onScrRefChange`. Exactly two emission shapes
+ * exist, both produced only by `report()`:
+ *   - position report: chapter + verse read from the loaded document in ONE snapshot, with the
+ *     book from that same snapshot's BookNode - falling back to the scrRef prop's book ONLY
+ *     when the document has no book code (a book-less document cannot name its own book, and
+ *     corrections never fire there, so the prop is the sole authority) - and the prop's
+ *     versificationStr riding along (a document states no versification);
+ *   - book correction: { ...scrRef, book: <document's book> } - the one deliberate mixed-source
+ *     shape ("keep your position, fix the book"), for documents whose book differs from scrRef.
+ *
+ * Why this file is shaped the way it is - three async signals meet here, plus a feedback loop:
+ *   1. the scrRef prop (the host's intent - never wrong, but everything else lags it);
+ *   2. document content (arrives a USJ-load later than the intent it belongs to; a SUPERSEDED
+ *      navigation's document can land after a NEWER navigation started);
+ *   3. selection events (browser macrotasks; can describe positions that are already obsolete);
+ *   4. our own emissions return as the scrRef prop (the echo loop).
+ *
+ * The races these caused (regression tests exist for each):
+ *   R1 echo treated as navigation      -> caret yanked mid-typing
+ *   R2 settle reported as a user move  -> verse resets to 0 (e.g. 78:9 -> 78:0 on reattach)
+ *   R3 stale queued selectionchange    -> rapid navigation hijacked back to a previous target
+ *   R4 superseded document lands late  -> navigation hijacked to an abandoned target
+ *   R5 mixed-source emissions          -> references that describe no real position
+ *   R6 doc mutations mid-navigation    -> old book echoed during cross-book navigation
+ *
+ * The machine: phase is "idle" or "navigating" (see NavPhase). We call the span where
+ * phase === "navigating" the "navigation window". While "navigating", this plugin emits NOTHING
+ * (kills R2, R3, R4, R6). The navigation window opens on any external scrRef change, after emitting
+ * a book correction, or on a document REPLACEMENT while idle (a same-book reload's own settle
+ * transiently resolves verse 0 - reporting it is the R2 clobber; pinned by test). It closes on real
+ * user input (pointerdown/keydown/beforeinput on the root - beforeinput catches IME/voice/paste
+ * input paths that move the selection without a pointerdown or keydown) or when one of our own
+ * emissions echoes back as the prop. `pendingEchoes` (a FIFO, deliberately not a
+ * boolean or a single slot) recognizes echoes (kills R1). `report()` enforces the two emission
+ * shapes and dedupes (kills R5). `$resolvePosition` refuses to describe positions the document
+ * cannot address; content before the first chapter of a loaded document addresses as chapter 1
+ * verse 0 by USFM convention - verse 0 is DATA here, never a sentinel.
+ *
+ * Invariants (each backed by a named test in ScriptureReferencePlugin.test.tsx):
+ *   I1 single-source        position reports come from one document snapshot (book falls back
+ *                           to the prop only when the document has no book code;
+ *                           versificationStr always rides from the prop - a document has none)
+ *   I2 navigating-silence   zero emissions while navigating
+ *   I3 echo-inert           echoed props never move the caret nor open a navigation window
+ *   I4 user-report          after user input, the next differing selection reports (sole
+ *                           exception: the pending-echo dedupe trade-off below)
+ *   I5 resolved-position    never emit a position the document cannot address
+ *   I6 placement-book-gate  prop-driven placement never moves the caret in a different book's
+ *                           document; arrival-driven placement targets the arriving document only
+ *   I7 mount-correction     a mounted document with a mismatched book corrects the host once
+ *   I8 verse-0-is-data      verse 0 reports from idle and applies as a target
+ *
+ * Documented trade-offs (deliberate; do not "fix" one without weighing its counterpart):
+ *   - A late echo arriving after a newer external navigation cleared the queue is treated as a
+ *     real navigation (pinned: "treats an echo arriving after a newer external navigation...").
+ *   - A user selection equal to a STILL-PENDING echo of an earlier report is deduped, not
+ *     re-emitted - report() cannot distinguish it from the echo. The host may briefly hold a
+ *     newer ref than the caret; the next differing selection self-heals. Emitting instead would
+ *     leave a never-consumed echo in the FIFO that would later swallow a REAL navigation to that
+ *     ref - strictly worse. This is the sole exception to I4.
+ *   - A cross-editor paste or undo that (re)creates a BookNode while a document is already loaded
+ *     triggers neither placement nor a navigation window: pastes must not teleport the caret to
+ *     verse start (pinned).
+ *
+ * UNSAFE without re-running the full test suite AND the live Platform.Bible repros:
+ *   - re-registering any listener on chapter/verse deps (skipInitialization replay fires against
+ *     whatever document happens to be mounted - this exact mistake caused the NUM 27:1 bug);
+ *   - emitting anything while phase === "navigating" (no "genuine jump" carve-outs: every such
+ *     carve-out forces settle-shape classification, which is how verse-0 sentinels are born);
+ *   - collapsing pendingEchoes to a boolean or single slot (two reports can be in flight);
+ *   - building an emission from more than one source (except the book-correction shape and
+ *     I1's documented no-book fallback).
+ *
+ * SAFE (additive) changes:
+ *   - adding more events that close the navigation window;
+ *   - placement internals ($moveCaretToVerseStart), as long as it stays a pure caret move;
+ *   - additional validity checks before report().
+ * ============================================================================================
+ */
+
 import { useLexicalComposerContext } from "@lexical/react/LexicalComposerContext";
 import { mergeRegister } from "@lexical/utils";
 import { SerializedVerseRef } from "@sillsdev/scripture";
@@ -33,96 +119,12 @@ import {
   ImmutableVerseNode,
 } from "shared-react";
 
-/*
- * ============================================================================================
- * HOW THIS PLUGIN WORKS - READ BEFORE CHANGING ANYTHING
- * ============================================================================================
- *
- * Contract: the host owns `scrRef`; this plugin (a) applies scrRef changes to the caret and
- * (b) reports genuine user caret moves back via `onScrRefChange`. Exactly two emission shapes
- * exist, both produced only by `report()`:
- *   - position report: chapter + verse read from the loaded document in ONE snapshot, with the
- *     book from that same snapshot's BookNode - falling back to the scrRef prop's book ONLY
- *     when the document has no book code (a book-less document cannot name its own book, and
- *     corrections never fire there, so the prop is the sole authority) - and the prop's
- *     versificationStr riding along (a document states no versification);
- *   - book correction: { ...scrRef, book: <document's book> } - the one deliberate mixed-source
- *     shape ("keep your position, fix the book"), for documents whose book differs from scrRef.
- *
- * Why this file is shaped the way it is - three async signals meet here, plus a feedback loop:
- *   1. the scrRef prop (the host's intent - never wrong, but everything else lags it);
- *   2. document content (arrives a USJ-load later than the intent it belongs to; a SUPERSEDED
- *      navigation's document can land after a NEWER navigation started);
- *   3. selection events (browser macrotasks; can describe positions that are already obsolete);
- *   4. our own emissions return as the scrRef prop (the echo loop).
- *
- * The races these caused (regression tests exist for each):
- *   R1 echo treated as navigation      -> caret yanked mid-typing
- *   R2 settle reported as a user move  -> verse resets to 0 (e.g. 78:9 -> 78:0 on reattach)
- *   R3 stale queued selectionchange    -> rapid navigation hijacked back to a previous target
- *   R4 superseded document lands late  -> navigation hijacked to an abandoned target
- *   R5 mixed-source emissions          -> references that describe no real position
- *   R6 doc mutations mid-navigation    -> old book echoed during cross-book navigation
- *
- * The machine: phase is "idle" or "navigating" (see NavPhase). While "navigating", this plugin
- * emits NOTHING (kills R2, R3, R4, R6). The window opens on any external scrRef change, after
- * emitting a book correction, or on a document REPLACEMENT while idle (a same-book reload's own
- * settle transiently resolves verse 0 - reporting it is the R2 clobber; pinned by test). It
- * closes on real user input (pointerdown/keydown/beforeinput on the root - beforeinput catches
- * IME/voice/paste input paths that move the selection without a pointerdown or keydown) or when
- * one of our own emissions echoes back as the prop. `pendingEchoes` (a FIFO, deliberately not a
- * boolean or a single slot) recognizes echoes (kills R1). `report()` enforces the two emission
- * shapes and dedupes (kills R5). `$resolvePosition` refuses to describe positions the document
- * cannot address; content before the first chapter of a loaded document addresses as chapter 1
- * verse 0 by USFM convention - verse 0 is DATA here, never a sentinel.
- *
- * Invariants (each backed by a named test in ScriptureReferencePlugin.test.tsx):
- *   I1 single-source        position reports come from one document snapshot (book falls back
- *                           to the prop only when the document has no book code;
- *                           versificationStr always rides from the prop - a document has none)
- *   I2 navigating-silence   zero emissions while navigating
- *   I3 echo-inert           echoed props never move the caret nor open a window
- *   I4 user-report          after user input, the next differing selection reports (sole
- *                           exception: the pending-echo dedupe trade-off below)
- *   I5 resolved-position    never emit a position the document cannot address
- *   I6 placement-book-gate  prop-driven placement never moves the caret in a different book's
- *                           document; arrival-driven placement targets the arriving document only
- *   I7 mount-correction     a mounted document with a mismatched book corrects the host once
- *   I8 verse-0-is-data      verse 0 reports from idle and applies as a target
- *
- * Documented trade-offs (deliberate; do not "fix" one without weighing its counterpart):
- *   - A late echo arriving after a newer external navigation cleared the queue is treated as a
- *     real navigation (pinned: "treats an echo arriving after a newer external navigation...").
- *   - A user selection equal to a STILL-PENDING echo of an earlier report is deduped, not
- *     re-emitted - report() cannot distinguish it from the echo. The host may briefly hold a
- *     newer ref than the caret; the next differing selection self-heals. Emitting instead would
- *     leave a never-consumed echo in the FIFO that would later swallow a REAL navigation to that
- *     ref - strictly worse. This is the sole exception to I4.
- *   - A cross-editor paste or undo that (re)creates a BookNode while a document is already loaded
- *     triggers neither placement nor a window: pastes must not teleport the caret (pinned).
- *
- * UNSAFE without re-running the full test suite AND the live Platform.Bible repros:
- *   - re-registering any listener on chapter/verse deps (skipInitialization replay fires against
- *     whatever document happens to be mounted - this exact mistake caused the NUM 27:1 bug);
- *   - emitting anything while phase === "navigating" (no "genuine jump" carve-outs: every such
- *     carve-out forces settle-shape classification, which is how verse-0 sentinels are born);
- *   - collapsing pendingEchoes to a boolean or single slot (two reports can be in flight);
- *   - building an emission from more than one source (except the book-correction shape and
- *     I1's documented no-book fallback).
- *
- * SAFE (additive) changes:
- *   - adding more events that close the window;
- *   - placement internals ($moveCursorToVerseStart), as long as it stays a pure caret move;
- *   - additional validity checks before report().
- * ============================================================================================
- */
-
 /** "idle": selection changes are the user's. "navigating": an external scrRef change (or a book
  * correction we emitted) is still being applied; every emission is suppressed until real user
- * input on the editor ends the window or our own echo returns. */
+ * input on the editor ends the navigation window or our own echo returns. */
 type NavPhase = "idle" | "navigating";
 
-/** All plugin state plus the latest wiring values, one mutable object shared by the handlers. */
+/** All plugin state plus the latest props. One mutable object shared by the handlers. */
 interface Machine {
   phase: NavPhase;
   /** Emissions not yet seen back as the `scrRef` prop (FIFO, newest last). A queue, not a slot:
@@ -131,42 +133,10 @@ interface Machine {
   scrRef: SerializedVerseRef;
   onScrRefChange: (scrRef: SerializedVerseRef) => void;
   /** Whether a BookNode-created mutation has ever been observed. Distinguishes the mount of a
-   * fresh editor (no window needed - a null selection or one already at scrRef) from a later
-   * pure-created BookNode arriving while a document is already loaded (paste/undo - must not
-   * teleport the caret). */
+   * fresh editor (no navigation window needed - a null selection or one already at scrRef) from a
+   * later pure-created BookNode arriving while a document is already loaded (paste/undo - must
+   * not teleport the caret to verse start). */
   sawDocument: boolean;
-}
-
-const MAX_PENDING_ECHOES = 8;
-
-function refsEqual(a: SerializedVerseRef, b: SerializedVerseRef): boolean {
-  return (
-    a.book === b.book &&
-    a.chapterNum === b.chapterNum &&
-    a.verseNum === b.verseNum &&
-    (a.verse ?? undefined) === (b.verse ?? undefined)
-  );
-}
-
-/** Consume `ref` as an echo of an earlier emission. Consuming closes any open window: a returned
- * echo means the host and editor agree. Returns false when `ref` is not ours. */
-function consumePendingEcho(machine: Machine, ref: SerializedVerseRef): boolean {
-  const index = machine.pendingEchoes.findIndex((echo) => refsEqual(echo, ref));
-  if (index < 0) return false;
-  machine.pendingEchoes.splice(0, index + 1);
-  machine.phase = "idle";
-  return true;
-}
-
-/** The only caller of `onScrRefChange`. Drops emissions equal to the current prop or to a
- * pending echo. Returns whether an emission actually happened. */
-function report(machine: Machine, ref: SerializedVerseRef): boolean {
-  if (refsEqual(ref, machine.scrRef)) return false;
-  if (machine.pendingEchoes.some((echo) => refsEqual(echo, ref))) return false;
-  machine.pendingEchoes.push(ref);
-  if (machine.pendingEchoes.length > MAX_PENDING_ECHOES) machine.pendingEchoes.shift();
-  machine.onScrRefChange(ref);
-  return true;
 }
 
 interface ResolvedPosition {
@@ -177,166 +147,7 @@ interface ResolvedPosition {
   verse: string | undefined;
 }
 
-/** First BookNode of the root (documents put it first; strays after content are ignored). */
-function $getFirstBookNode(): BookNode | undefined {
-  let node = $getRoot().getFirstChild();
-  while (node) {
-    if ($isBookNode(node)) return node;
-    node = node.getNextSibling();
-  }
-  return undefined;
-}
-
-/** Resolve the selection to a position the document can address, or undefined (no selection, or
- * an empty document). Content before the first chapter of a loaded document addresses as
- * chapter 1 per USFM convention (its verse resolves to 0). */
-function $resolvePosition(): ResolvedPosition | undefined {
-  const selection = $getSelection();
-  const startNode = getSelectionStartNode(selection);
-  if (!startNode) return undefined;
-
-  // The unaddressable check needs the NODE's existence - a BookNode with an empty code still
-  // makes a document addressable.
-  const bookNode = $getFirstBookNode();
-  const chapterNode = $findThisChapter(startNode);
-  if (!chapterNode && !bookNode) return undefined;
-  const chapterNum = chapterNode ? parseInt(chapterNode.getNumber() ?? "1", 10) : 1;
-  if (Number.isNaN(chapterNum)) return undefined; // unaddressable - consistent with I5
-
-  const verseNode = $resolveVerseNode(startNode, selection);
-  const { verseNum, verse } = $getEffectiveVerseForBcv(verseNode ?? undefined, selection);
-  if (Number.isNaN(verseNum)) return undefined; // unaddressable - consistent with I5
-  return { book: bookNode?.getCode() || undefined, chapterNum, verseNum, verse };
-}
-
-/** isVerseInRange throws on malformed ranges (reversed, 3-part) from imported USFM; a malformed
- * range simply doesn't contain the verse. */
-function verseInRangeSafe(verseNum: number, verseRange: string): boolean {
-  try {
-    return isVerseInRange(verseNum, verseRange);
-  } catch {
-    return false;
-  }
-}
-
-/** Whether the resolved position is already what `scrRef` describes (verse-range aware). */
-function positionMatchesScrRef(position: ResolvedPosition, scrRef: SerializedVerseRef): boolean {
-  if (position.book && position.book !== scrRef.book) return false;
-  if (position.chapterNum !== scrRef.chapterNum) return false;
-  return position.verse
-    ? verseInRangeSafe(scrRef.verseNum, position.verse)
-    : scrRef.verseNum === position.verseNum;
-}
-
-function positionToScrRef(
-  position: ResolvedPosition,
-  hostRef: SerializedVerseRef,
-): SerializedVerseRef {
-  const ref: SerializedVerseRef = {
-    book: position.book || hostRef.book,
-    chapterNum: position.chapterNum,
-    verseNum: position.verseNum,
-  };
-  if (position.verse != null) ref.verse = position.verse;
-  // A document states no versification, so the host's rides along on every position report;
-  // stripping it could map the reference to the wrong physical verse under a divergent
-  // versification. refsEqual ignores it, so echo detection is unaffected.
-  if (hostRef.versificationStr != null) ref.versificationStr = hostRef.versificationStr;
-  return ref;
-}
-
-/** `propChanged`: echo of our own report, or an external navigation. */
-function onPropChanged(machine: Machine, editor: LexicalEditor, newRef: SerializedVerseRef) {
-  if (consumePendingEcho(machine, newRef)) return; // echoes never move the caret
-
-  machine.phase = "navigating";
-  machine.pendingEchoes.length = 0;
-  // Prop-driven placement gate: never move the caret inside a different book's (stale) document.
-  const contentBook = editor
-    .getEditorState()
-    .read(() => $getFirstBookNode()?.getCode() || undefined);
-  if (!contentBook || contentBook === newRef.book) {
-    editor.update(() => $moveCursorToVerseStart(newRef.chapterNum, newRef.verseNum), {
-      tag: CURSOR_CHANGE_TAG,
-    });
-  }
-}
-
-/** `documentChanged`: the loaded document's book identity was established or changed. */
-function onDocumentChanged(
-  machine: Machine,
-  editor: LexicalEditor,
-  code: string | undefined,
-  batch: { hasCreated: boolean; hasDestroyed: boolean },
-) {
-  if (machine.phase === "navigating") {
-    // Silence - except the arrival of the document the navigation is waiting for.
-    if (batch.hasCreated && code && code === machine.scrRef.book) {
-      schedulePlacement(machine, editor);
-    }
-    if (batch.hasCreated) machine.sawDocument = true;
-    return;
-  }
-
-  if (batch.hasCreated && batch.hasDestroyed) {
-    // REPLACEMENT (LoadStatePlugin's setEditorState swap, in-editor book jump): the swap's own
-    // synthetic selection settle fires before the deferred placement and must be silenced -
-    // regardless of book match, since a same-book reload's transient chapter-top settle is the
-    // R2 clobber this branch exists to prevent. The correction below (d), if any, runs after.
-    schedulePlacement(machine, editor);
-    machine.phase = "navigating";
-  } else if (batch.hasCreated && !machine.sawDocument) {
-    // MOUNT: fresh editor, no window needed - a null selection or one already at scrRef.
-    schedulePlacement(machine, editor);
-  }
-  // else: a later pure-created BookNode (cross-editor paste, undo recreating a BookNode) while a
-  // document is already loaded - neither placement nor a window. A paste must not teleport the
-  // caret.
-  if (batch.hasCreated) machine.sawDocument = true;
-
-  // Book correction: "keep your position, fix the book". Emitting one opens a navigation window
-  // (the swap's own synthetic selection settle fires before the deferred placement and must be
-  // silenced); the correction's echo closes it. Runs after (a)-(c); under (a) phase is already
-  // "navigating" here, so this assignment is harmless.
-  if (code && code !== machine.scrRef.book) {
-    if (report(machine, { ...machine.scrRef, book: code })) machine.phase = "navigating";
-  }
-}
-
-/** Mutation listeners must not call editor.update synchronously (repo rule); defer one
- * microtask, as LoadStatePlugin does. Reads the latest scrRef at execution time. Deliberately
- * NOT book-gated at fire time: on an idle cross-book REPLACEMENT the book correction's echo is
- * still in flight when this fires, so scrRef.book still names the old book and a gate here
- * would skip the placement that makes the caret match the just-emitted correction. */
-function schedulePlacement(machine: Machine, editor: LexicalEditor) {
-  queueMicrotask(() => {
-    editor.update(
-      () => $moveCursorToVerseStart(machine.scrRef.chapterNum, machine.scrRef.verseNum),
-      { tag: CURSOR_CHANGE_TAG },
-    );
-  });
-}
-
-/** `selectionSettled`: the caret is somewhere; the phase decides whose action that was.
- * `position` must be resolved by the caller (inline in the SELECTION_CHANGE_COMMAND listener,
- * not re-fetched in here via `editor.getEditorState()`): dispatchCommand always invokes that
- * listener inside an active editor-state context - a fresh implicit update, or, for the native
- * selectionchange path, the surrounding not-yet-committed update's pending state reused in place.
- * Reading `editor.getEditorState()` from a plain helper like this one would instead see the
- * *last committed* state, which on that nested/native path is one selection change stale - the
- * very commit still in flight - so a genuine settle would compare against the old position and be
- * silently dropped. */
-function onSelectionSettled(machine: Machine, position: ResolvedPosition | undefined) {
-  if (machine.phase === "navigating") return; // settles of an in-flight navigation are noise
-  if (!position) return; // never report a position the document cannot address
-  if (positionMatchesScrRef(position, machine.scrRef)) return;
-  report(machine, positionToScrRef(position, machine.scrRef));
-}
-
-/** `userInteracted`: real input ends settling; what follows is the user's. */
-function onUserInteracted(machine: Machine) {
-  machine.phase = "idle";
-}
+const MAX_PENDING_ECHOES = 8;
 
 /**
  * A component (plugin) that keeps the Scripture reference updated.
@@ -344,7 +155,7 @@ function onUserInteracted(machine: Machine) {
  * @param onScrRefChange - Callback function when the Scripture reference has changed.
  * @returns null, i.e. no DOM elements.
  */
-export default function ScriptureReferencePlugin({
+export function ScriptureReferencePlugin({
   scrRef,
   onScrRefChange,
 }: {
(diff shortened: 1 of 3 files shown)


katherinejensen00 (dismissed):
Thanks for making these fixes, Ira! The code looks good so I will go ahead and approve. There are two documentation notes that maybe you can fold into another PR later, but no worries if you don't.

> @katherinejensen00 reviewed 3 files and all commit messages, and made 3 comments.
> *[Reviewable](https://reviewable.io/reviews/eten-tech-foundation/scripture-editors/500#-:-OxW0ZA97fKiN5krjwmP:bln2p04)* status: all files reviewed, 2 unresolved discussions (waiting on irahopkinson).

___
*[`packages/platform/src/editor/ScriptureReferencePlugin.tsx` line 340 at r1](https://reviewable.io/reviews/eten-tech-foundation/scripture-editors/500#-OxW0T0I1itmmf7zKUut:-OxW0T0I1itmmf7zKUuu:b6fjyak) ([raw file](https://github.com/eten-tech-foundation/scripture-editors/blob/97172c85a6b617bef5df3a6dcc58f0ae4cef3d30/packages/platform/src/editor/ScriptureReferencePlugin.tsx#L340)):*
> ```XML
>     // synthetic selection settle fires before the deferred placement and must be silenced -
>     // regardless of book match, since a same-book reload's transient chapter-top settle is the
>     // R2 clobber this branch exists to prevent. The correction below (d), if any, runs after.
> ```

NIT Claude Finding

**The `(a)`–`(d)` cross-references in `onDocumentChanged` are used in prose but never anchored to their code branches.**

The REPLACEMENT branch says "The correction below (d), if any, runs after." The book correction block says "Runs after (a)–(c); under (a) phase is already 'navigating' here." But no comment in the code is actually labeled `// (a)`, `// (b)`, etc. — a reader has to manually count branches to decode the letter scheme. If (a) is REPLACEMENT, (b) is MOUNT, (c) is the paste/undo no-op, and (d) is the book correction block, that can be inferred, but it takes real effort. This predates your PR, so it doesn't need to block merge — but a one-liner label at the start of each branch (`// (a) REPLACEMENT: ...`, etc.) would make the cross-references self-evident.
___
*[`packages/platform/src/editor/ScriptureReferencePlugin.tsx` line 375 at r1](https://reviewable.io/reviews/eten-tech-foundation/scripture-editors/500#-OxW0Iwt8iOI613Zaf7S:-OxW0Iwt8iOI613Zaf7T:b-jowbbb) ([raw file](https://github.com/eten-tech-foundation/scripture-editors/blob/97172c85a6b617bef5df3a6dcc58f0ae4cef3d30/packages/platform/src/editor/ScriptureReferencePlugin.tsx#L375)):*
> ```XML
> }
> 
> function $moveCaretToVerseStart(chapterNum: number, verseNum: number) {
> ```

NIT Claude Finding
**`$moveCaretToVerseStart` is the only named handler in the file without a JSDoc, and it's the most complex one.**

Every other handler in the file — `onPropChanged`, `consumePendingEcho`, `$resolvePosition`, `onDocumentChanged`, `schedulePlacingCaretAtVerseStart`, `onSelectionSettled`, `positionMatchesScrRef`, `verseInRangeSafe`, `positionToScrRef`, `report`, `refsEqual`, `onUserInteracted` — has a function-level JSDoc comment. `$moveCaretToVerseStart` doesn't, and it's the function with the most state assumptions: it must be called inside a Lexical editor-state update, it has an early-return when the caret is already in a verse range, and it handles the "no verse node, fall back to para" case. The file-top HOW-THIS-PLUGIN-WORKS block explains the overall algorithm but doesn't substitute for a function signature–adjacent description that callers read in context. Since this PR renamed the function and moved it, this is a good moment to close that gap with even two or three sentences.

lyonsil (commented):
## Review: the reorg is clean, but two `editor.read()` swaps are not cleanup

I verified the mechanical half of this PR and it is solid: every moved function body is byte-identical apart from the two swaps below, no other file imports the old default export, no barrel/mock/lazy-import references it, and `nx extract-api` is not needed (the plugin is not on the package's public surface). No CLAUDE.md violations, no over-length lines, no em dashes.

The concern is the two `editor.getEditorState().read(cb)` -> `editor.read(cb)` swaps (lines 199 and 270). In Lexical 0.43 those are **not** synonyms (`Lexical.dev.mjs:10935`):

```js
read(callbackFn) {
  $commitPendingUpdates(this);   // <-- synchronously FLUSHES pending updates
  return this.getEditorState().read(callbackFn, { editor: this });
}
```

There is no `_updating` guard - the very guard Lexical applies to itself in `setEditorState` (*"otherwise this will cause a second commit with an already read-only state and selection"*) and warns about in `triggerCommandListeners` (*"...will potentially corrupt the nodeMap by doing GC too early"*).

And the flush window is real in production: `LoadStatePlugin` calls `editor.setEditorState()` **inside** `editor.update()`, so `_updating === true`, `setEditorState`'s own commit is skipped, and `_pendingEditorState` survives - holding the new document - until a later microtask.

**Recommendation: revert both swaps to `editor.getEditorState().read(...)`.** At line 199 the flush is *provably* incapable of changing the value read (Lexical nulls `_pendingEditorState` at `:8504`, before firing mutation listeners at `:8605`), so it is pure risk for zero benefit. At line 270 it is actively wrong: it commits a pending document and then reads it, which is exactly what the gate on the next line exists to prevent. The other two mutation listeners in this editor (`AnnotationPlugin.tsx:85`, `CommentPlugin.tsx:862`) both still use `getEditorState().read(...)`, so reverting restores consistency rather than breaking it.

If the terseness is the goal, extract a named helper instead - see the comment on line 199.

Comments below: **1-6** are introduced by this PR; **14-25** are cleanup notes held to the PR's own "cleanup" billing. (I also found 7 pre-existing defects from #499 in the functions this PR moved - empty-book-code placement, `sawDocument` vs book-less documents, a stale `verse` range riding into book corrections, and others. Those are not this PR's doing and I have left them off this review; happy to file them separately.)

Verified against Lexical 0.43.0 and React 18.3.1 (note: repo CLAUDE.md still says React 19.1.0 - it resolves 18.3.1).

packages/platform/src/editor/ScriptureReferencePlugin.tsx
```diff
@@ -451,7 +260,119 @@ export default function ScriptureReferencePlugin({
...
 
-function $moveCursorToVerseStart(chapterNum: number, verseNum: number) {
+/** `propChanged`: echo of our own report, or an external navigation. */
+function onPropChanged(machine: Machine, editor: LexicalEditor, newRef: SerializedVerseRef) {
+  if (consumePendingEcho(machine, newRef)) return; // echoes never move the caret
+
+  machine.phase = "navigating";
+  machine.pendingEchoes.length = 0;
+  // Prop-driven placement gate: never move the caret inside a different book's (stale) document.
+  const bookCode = editor.read(() => $getFirstBookNode()?.getCode() || undefined);
```
lyonsil: **[1/18 - CONFIRMED] The flush makes `useDecorators` call `flushSync` inside React's commit phase.**

`onPropChanged` runs from a React passive effect. When this `editor.read()` flush is non-inert, `$commitPendingUpdates` fires Lexical's decorator listener, and `@lexical/react`'s `useDecorators` (`LexicalRichTextPlugin.dev.mjs:72-77`) calls `flushSync`. That now runs inside `flushPassiveEffects`, so React logs *"flushSync was called from inside a lifecycle method"* and **silently defers the decorator render** instead of flushing it.

Reproduced end-to-end against real React 18.3.1 + Lexical 0.43.0 with a `LexicalComposer` + `RichTextPlugin` + a DecoratorNode. This is not a corner case: `ImmutableVerseNode`, `ImmutableChapterNode`, `ImmutableNoteCallerNode`, `MilestoneNode` all extend `DecoratorNode`, so any document-load `FULL_RECONCILE` populates `_pendingDecorators`.

Worth noting `LoadStatePlugin` already guards this exact hazard on purpose: *"Use queueMicrotask to defer the editor update outside of React's lifecycle, preventing flushSync warnings when this is triggered by a parent component update."* This line reintroduces precisely what that `queueMicrotask` exists to avoid.

This is the one consequence of the swap that is live and confirmed rather than latent.

**Fix: `editor.getEditorState().read(...)`.**

lyonsil: **[2/18 - PLAUSIBLE, most severe] `editor.read()` has no `_updating` guard, so it can commit a half-built editor state.**

`report()` calls the host's `onScrRefChange` from inside `onSelectionSettled`, which Lexical runs inside an **open `$beginUpdate`** (`_updating === true`, `_pendingEditorState` a live mutable clone) - the file's own comment on `onSelectionSettled` says as much. If the host responds with a synchronous re-render (`flushSync`, a sync external store), React flushes the pending passive effect re-entrantly, `onPropChanged` runs, and this `editor.read()` commits that **half-built** state: `$commitPendingUpdates` sets `_readOnly = true` and freezes the selection / swaps the node map for throwing stubs while `$beginUpdate` is still between `updateFn()` and `$applyAllTransforms`/`$garbageCollectDetachedNodes`.

Reproduced under a `flushSync` host:

```
MODE=pr   (editor.read)           -> errors: ['Cannot call set() on a frozen Lexical node map']
                                     committed text: "GEN"   <-- enclosing update's work LOST
MODE=main (getEditorState().read) -> errors: []
                                     committed text: "GEN\n\nWORK-AFTER-THE-READ"  <-- survives
```

The enclosing update's tail-scheduled commit then no-ops, because `_pendingEditorState` is already null - so the edit is silently dropped. **And the freeze guard is dev-only** (`grep -c "frozen Lexical node map" Lexical.prod.mjs` -> 0), so production corrupts silently where dev throws.

This is exactly the hazard Lexical guards in `setEditorState` with `if (!this._updating) { $commitPendingUpdates(this); }`. `editor.read()` has no such guard.

Marked PLAUSIBLE only because the trigger depends on whether the Platform.Bible host commits `scrRef` synchronously - which I cannot see from this repo. **Could you confirm whether the host wraps the scrRef update in `flushSync` or a sync store?** If it does, this is a crash-and-lose-the-edit bug.

lyonsil: **[3/18 - PLAUSIBLE] Gate inversion: the flush commits the pending document, then reads it - defeating invariant I6.**

The comment right above this line states the purpose: *"never move the caret inside a different book's (stale) document"*, and I6 says *"prop-driven placement never moves the caret in a different book's document"*. That question can only be answered by the **currently loaded** document - which is what `getEditorState().read()` returned.

`editor.read()` commits the pending swap **first**, then reads it. Demonstrated against real Lexical 0.43.0, replicating `LoadStatePlugin`'s `queueMicrotask(() => editor.update(() => editor.setEditorState(...)))`:

```
_pendingEditorState !== null : true   _updating: false
OLD  editor.getEditorState().read() -> "GEN"
NEW  editor.read()                  -> "EXO"
GATE INPUT CHANGED BY THE PR        : true
```

So the gate on line 271 passes where it previously (correctly) failed, and a prop-driven placement runs inside a document the user cannot see yet. The gate also stops being a pure predicate: evaluating it *installs* the document it then approves.

The read is also demonstrably reachable with a live pending state and a **plain** host (no `flushSync`): Lexical dispatches `SELECTION_CHANGE_COMMAND` synchronously inside the DOM `selectionchange` event, which React 18 treats as `DiscreteEventPriority` -> SyncLane, and a SyncLane commit flushes its passive effects synchronously (`react-dom.development.js:26973`). I measured 3 hits of this line with `_pendingEditorState !== null`. What I did not reproduce end-to-end is that window coinciding with a *cross-book* swap - hence PLAUSIBLE rather than CONFIRMED.

No test pins this: all 37 stay green either way.

packages/platform/src/editor/ScriptureReferencePlugin.tsx
```diff
@@ -451,7 +260,119 @@ export default function ScriptureReferencePlugin({
...
+  if (!bookCode || bookCode === newRef.book) {
+    editor.update(() => $moveCaretToVerseStart(newRef.chapterNum, newRef.verseNum), {
```
lyonsil, line 276: **[5/18 - CONFIRMED] Side effect of the line-270 flush: it splits one commit into two.**

Without the flush, this `editor.update()` merges into the still-pending document state (`$beginUpdate` sees pending non-null and not `_readOnly`, so `editorStateWasCloned = false`) and the **document and the caret move commit together**.

With the flush, the document commits first - firing every update/mutation/decorator listener with the caret **not yet placed** - and the caret move then commits separately. So `DeltaOnChangePlugin`, `HistoryPlugin` and the annotation plugins all observe an intermediate state that previously never existed.

This resolves itself if line 270 goes back to `getEditorState().read(...)`.

packages/platform/src/editor/ScriptureReferencePlugin.tsx
```diff
@@ -385,9 +196,7 @@ export default function ScriptureReferencePlugin({
         (nodeMutations) => {
           const kinds = [...nodeMutations.values()];
           if (kinds.every((kind) => kind === "destroyed")) return;
-          const code = editor
-            .getEditorState()
-            .read(() => $getFirstBookNode()?.getCode() || undefined);
+          const code = editor.read(() => $getFirstBookNode()?.getCode() || undefined);
```
lyonsil: **[4/18 - CONFIRMED] At this site the swap provably buys nothing, and adds a re-entrancy.**

On the normal path this listener runs from inside `$commitPendingUpdates`, which nulls `_pendingEditorState` at `Lexical.dev.mjs:8504` and installs the new state at `:8505` - **100 lines before** it fires mutation listeners at `:8605`. So `editor.read()`'s `$commitPendingUpdates` hits the early return at `:8486-8488`, and `getEditorState()` already holds exactly the state it would have returned. **The value read cannot differ.** Zero benefit.

The risk is not zero, though. This listener explicitly opts into `{ skipInitialization: false }`, and Lexical's `registerMutationListener` inserts the listener into `_listeners.mutation` (`:10648`) **before** calling `initializeMutationListener` (`:10653`), which invokes it synchronously from inside the `useEffect` - **not** inside a commit, so `_pendingEditorState` is not nulled there. If a pending update exists at that moment, this `editor.read()` runs a full commit whose `triggerMutationListeners` iterates `Array.from(editor._listeners.mutation)` (`:8646`) - which already contains **this listener** - and re-enters it. `onDocumentChanged` then runs twice for one document arrival, mutating `machine.phase` / `sawDocument` / `pendingEchoes` under itself.

The pre-PR `getEditorState().read()` could not do this. Note the file's own header calls the `skipInitialization` replay out as the source of the NUM 27:1 bug.

**Fix: revert to `editor.getEditorState().read(...)`** - which is also what `AnnotationPlugin.tsx:85` and `CommentPlugin.tsx:862` still use.

lyonsil: **[6/18 - PLAUSIBLE, latent] Cross-plugin listener-ordering hazard introduced here.**

`triggerMutationListeners` (`Lexical.dev.mjs:8645-8661`) - unlike `triggerListeners` - never sets `editor._updating`. So a mutation listener registered **before** this one that calls `editor.update()` runs `$beginUpdate` immediately and returns leaving `_pendingEditorState` non-null (commit deferred to a microtask). This `editor.read()` then commits it **re-entrantly, mid-loop, from inside the outer `$commitPendingUpdates`** - and the outer commit's `update` listener subsequently fires with an `editorState` that is no longer `editor.getEditorState()`. Reproduced:

```
[0] reported="seed\n\n-TRIGGER\n\n-FROM_LISTENER1"  live="...-FROM_LISTENER1"  stale=false
[1] reported="seed\n\n-TRIGGER"                      live="...-FROM_LISTENER1"  stale=true
```

`DeltaOnChangePlugin` derives USJ ops from exactly that `prevEditorState -> editorState` pair, and `HistoryPlugin` keys undo/redo on it.

**Not reachable today** - this BookNode listener happens to register first (SRP is at `Editor.tsx:423`, ahead of `AnnotationPlugin` 452, `NoteNodePlugin` 458, `{children}` 469, all passive effects), and none of the others calls `editor.update()` synchronously. But it is armed by ordinary future edits: SRP is **conditionally mounted** (`{scrRef && onScrRefChange && ...}`), so a host that supplies `scrRef` late mounts it *last*; and any listener registered in a `useLayoutEffect` beats every passive effect regardless of JSX order.

The file's header already treats mutation-listener registration order as load-bearing, so this adds an undocumented ordering dependency to a file that documents its ordering dependencies.

lyonsil: **[17/18 - cleanup] The book-code read is duplicated, and the `contentBook` -> `bookCode` rename landed 1-of-3.**

The identical expression `read(() => $getFirstBookNode()?.getCode() || undefined)` is now written out twice - here and at line 270 - assigned to two different names (`code` here, `bookCode` there), while `onDocumentChanged`'s parameter (line 324) is still `code`. So the rename this PR advertises left the file *more* inconsistent than it found it.

It also matters beyond style: the `|| undefined` (empty code means "no book") is load-bearing and pinned by a test, yet it has to be kept in sync by hand across sites - and having the decision written out twice is precisely why the same read-vs-flush choice got made twice in one commit.

Suggested shape, which also fixes findings 4 and 1-3 in one move:

```ts
/** The loaded document's book code, from committed state. Empty code means no book. */
function getCommittedBookCode(editor: LexicalEditor): string | undefined {
  return editor.getEditorState().read(() => $getFirstBookNode()?.getCode() || undefined);
}
```

Then both sites are one call, the "committed, never flush" requirement is explicit and greppable, and `onDocumentChanged`'s param can be renamed to `bookCode` to finish the job.

packages/platform/src/editor/ScriptureReferencePlugin.tsx
```diff
@@ -481,3 +402,80 @@ function $moveCursorToVerseStart(chapterNum: number, verseNum: number) {
     else verseOrParaNode.select(0, 0);
   } else verseOrParaNode.selectNext(0, 0);
 }
+
+/** `selectionSettled`: the caret is somewhere; the phase decides whose action that was.
```
lyonsil, line 419: **[14/18 - cleanup, the one I'd most want changed in this PR]** 

This 9-line doc block constrains the **caller**, not this function: it says `position` must be resolved inline in the `SELECTION_CHANGE_COMMAND` listener and not re-fetched in here. That caller is line 217 - now **~190 lines away**, and labelled only `// selectionSettled`.

Before this PR the block sat 18 lines above the component containing the listener. Now someone tidying

```ts
if (machine.phase === "idle") onSelectionSettled(machine, $resolvePosition());
```

into the obvious-looking `onSelectionSettled(machine)` (the helper already has `machine`) sees **no warning at all** at the site where the mistake is made, and silently reintroduces the stale-position drop that R2/R3 exist to prevent.

This file opens with "READ BEFORE CHANGING ANYTHING" - its comment placement is a safety mechanism, not formatting, and I don't think a generic "helpers below the component" convention should have outranked it. Either move the constraint **into** line 217, or let this file be the documented exception to the layout convention.

packages/platform/src/editor/ScriptureReferencePlugin.tsx
```diff
@@ -481,3 +402,80 @@ function $moveCursorToVerseStart(chapterNum: number, verseNum: number) {
...
+ * `position` must be resolved by the caller (inline in the SELECTION_CHANGE_COMMAND listener,
+ * not re-fetched in here via `editor.getEditorState()`): dispatchCommand always invokes that
+ * listener inside an active editor-state context - a fresh implicit update, or, for the native
+ * selectionchange path, the surrounding not-yet-committed update's pending state reused in place.
+ * Reading `editor.getEditorState()` from a plain helper like this one would instead see the
```
lyonsil, line 424: **[16/18 - cleanup] This surviving warning now teaches the opposite of what the code requires.**

The comment warns against reading state via `editor.getEditorState()` - while the same PR made `editor.read()` house style two functions away, at lines 199 and 270. A reader reconciling the two concludes that `editor.read()` is the blessed modern replacement for `getEditorState()` throughout this file.

The truth is the opposite: `getEditorState().read()` is what lines 199 and 270 **require**, and this warning is scoped narrowly to the nested/native `selectionchange` path, where the *pending* state is the one that matters. Its stated rationale ("would see the last committed state") is also now factually wrong about `editor.read()`, which flushes.

The file ends up with two contradictory signals about the same API pair and no statement of the actual rule. Something like: *"read committed state from listeners and effects (`getEditorState().read()`); only the SELECTION_CHANGE path may rely on the active context; never `editor.read()` in this file - it commits."*

packages/platform/src/editor/ScriptureReferencePlugin.tsx
```diff
@@ -111,18 +78,53 @@ import {
  *     I1's documented no-book fallback).
  *
  * SAFE (additive) changes:
- *   - adding more events that close the window;
- *   - placement internals ($moveCursorToVerseStart), as long as it stays a pure caret move;
+ *   - adding more events that close the navigation window;
```
lyonsil, line 85: **[15/18 - cleanup] The UNSAFE / SAFE lists were edited cosmetically but not extended with the hazard this PR introduces.**

These lists are the mechanism by which this file protects invariants that no test can pin. This PR touched them ("window" -> "navigation window") without adding the one new hazard it created - which is worse than not touching them, because the diff now shows the hazard lists as reviewed-as-part-of-this-change, and a future reader will reasonably assume they are current.

The gap is real and provable: I applied every variant of the `editor.read()` mistake available in this file and **all 37 tests stayed green** every time. A maintainer who follows this file's own instructions - make the change, re-run the suite, see green - ships an R2/R6-class regression.

Suggested UNSAFE entry:

> - calling `editor.read()` anywhere in this file (it runs `$commitPendingUpdates` - a full synchronous commit that fires every listener, including ours); use `editor.getEditorState().read()` when you want the committed snapshot.

And honestly, a comment may be the wrong altitude for a constraint 37 tests cannot enforce. An ESLint `no-restricted-properties` rule banning `editor.read(` in this file would make the wrong thing hard to type. (The repo currently has zero `no-restricted-*` rules, so this is a genuinely absent mechanism rather than a duplicated one.)

packages/platform/src/editor/ScriptureReferencePlugin.tsx
```diff
@@ -177,174 +147,15 @@ interface ResolvedPosition {
...
-  if (!position) return; // never report a position the document cannot address
-  if (positionMatchesScrRef(position, machine.scrRef)) return;
-  report(machine, positionToScrRef(position, machine.scrRef));
-}
-
-/** `userInteracted`: real input ends settling; what follows is the user's. */
-function onUserInteracted(machine: Machine) {
-  machine.phase = "idle";
-}
+const MAX_PENDING_ECHOES = 8;
```
lyonsil: **[18/18 - cleanup] `MAX_PENDING_ECHOES` is now stranded ~310 lines from its only use.**

Its sole consumer is `report()` at line 464. Before this PR, `pendingEchoes` -> `MAX_PENDING_ECHOES` -> `refsEqual` -> `consumePendingEcho` -> `report` were one contiguous ~30-line block - the whole R1 echo-FIFO mechanism readable on one screen. The reorg scattered them to lines 150 / 469 / 280 / 460.

The "helpers below the component" rule was applied to the functions but not to the constant they use, so this one got the worst of both: a reader at line 464 has to scroll ~310 lines to learn that `MAX_PENDING_ECHOES` is even a constant rather than a prop, and a reader at line 150 has no context for what a "pending echo" is.

Module-scope consts have no ordering constraint - just move it down next to `report()`.

packages/platform/src/editor/ScriptureReferencePlugin.tsx
```diff
@@ -451,7 +260,119 @@ export default function ScriptureReferencePlugin({
...
+    schedulePlacingCaretAtVerseStart(machine, editor);
+  }
+  // else: a later pure-created BookNode (cross-editor paste, undo recreating a BookNode) while a
+  // document is already loaded - neither placement nor a navigation window. A paste must not
+  // teleport the caret to verse start.
+  if (batch.hasCreated) machine.sawDocument = true;
+
+  // Book correction: "keep your position, fix the book". Emitting one opens a navigation window
+  // (the swap's own synthetic selection settle fires before the deferred placement and must be
+  // silenced); the correction's echo closes it. Runs after (a)-(c); under (a) phase is already
```
lyonsil: **[19/18 - cleanup] Dangling `(a)`-`(d)` branch-label references.**

This comment says *"Runs after (a)-(c); under (a) phase is already 'navigating' here"*, and line 340 says *"The correction below (d), if any, runs after"* - but **no branch in this function is labelled (a), (b), (c) or (d)**. A reader who greps the file for `(a)` lands on the header's completely unrelated enumeration (*"this plugin (a) applies scrRef changes to the caret and (b) reports genuine user caret moves"*) and is actively misled.

The claim is also unverifiable as written: it only holds if (a) means the REPLACEMENT branch, since the `phase === "navigating"` branch returns at line 333 and can never reach line 356.

This PR rewrote the wording of both comments without fixing the references. Either label the cases inline (`// (a) REPLACEMENT`, `// (b) MOUNT`, `// (c) paste/undo - no placement`, `// (d) book correction`) or drop the letters and name the branches.

packages/platform/src/editor/ScriptureReferencePlugin.tsx
```diff
@@ -451,7 +260,119 @@ export default function ScriptureReferencePlugin({
...
+    schedulePlacingCaretAtVerseStart(machine, editor);
+    machine.phase = "navigating";
+  } else if (batch.hasCreated && !machine.sawDocument) {
+    // MOUNT: fresh editor, no navigation window needed - a null selection or one already at scrRef.
+    schedulePlacingCaretAtVerseStart(machine, editor);
+  }
+  // else: a later pure-created BookNode (cross-editor paste, undo recreating a BookNode) while a
+  // document is already loaded - neither placement nor a navigation window. A paste must not
+  // teleport the caret to verse start.
+  if (batch.hasCreated) machine.sawDocument = true;
```
lyonsil: **[20/18 - cleanup] `sawDocument` is assigned in both branches (332 and here), and the obvious consolidation is a trap.**

Hoisting a single `if (batch.hasCreated) machine.sawDocument = true;` above the `phase === "navigating"` check - the natural next cleanup - makes `!machine.sawDocument` on line 343 always false whenever `hasCreated` is true, **killing the MOUNT branch** and with it mount-time caret placement. Nothing in the code says so; the correctness of the current shape rests on an emergent property of statement order (the read at 343 happening before the write at 350).

A form that is both single-write and safe:

```ts
const isFirstDocument = batch.hasCreated && !machine.sawDocument;
if (batch.hasCreated) machine.sawDocument = true;
...
} else if (isFirstDocument) {
```

One write site instead of two, and the read-before-write ordering becomes explicit rather than a landmine.

packages/platform/src/editor/ScriptureReferencePlugin.tsx
```diff
@@ -177,174 +147,15 @@ interface ResolvedPosition {
...
+const MAX_PENDING_ECHOES = 8;
 
 /**
  * A component (plugin) that keeps the Scripture reference updated.
  * @param scrRef - Scripture reference.
  * @param onScrRefChange - Callback function when the Scripture reference has changed.
  * @returns null, i.e. no DOM elements.
  */
-export default function ScriptureReferencePlugin({
+export function ScriptureReferencePlugin({
```
lyonsil, line 160: **[21/18 - cleanup] The `export default` -> named export migration is half-applied, and it creates a name collision.**

**Half-applied:** `TreeViewPlugin` is still `export default` and is imported on the line directly below this one's import in `Editor.tsx` (`import TreeViewPlugin from "./TreeViewPlugin";`, line 9). `CommentPlugin` and `Editor` itself are also still default exports. A convention applied to one of four plugins in a directory is a diff, not a convention - either finish it or leave it.

**Collision:** `libs/shared-react/src/plugins/ScriptureReferencePlugin.tsx` exports a *different* plugin under the *same name* (props `{book?, onChangeReference?}`, both optional), re-exported from the `shared-react` barrel that `Editor.tsx` already imports ~25 names from. Two identically-named exports now sit in `Editor.tsx`'s import scope; IDE auto-import will offer both, and because both of the other plugin's props are optional, `<ScriptureReferencePlugin />` from the wrong module **type-checks silently** and yields a no-op plugin. The default export was the only thing preventing this.

Not a blocker - just worth knowing the migration bought a footgun that didn't exist before.

packages/platform/src/editor/ScriptureReferencePlugin.tsx
```diff
@@ -481,3 +402,80 @@ function $moveCursorToVerseStart(chapterNum: number, verseNum: number) {
...
+  if (position.book && position.book !== scrRef.book) return false;
+  if (position.chapterNum !== scrRef.chapterNum) return false;
+  return position.verse
+    ? verseInRangeSafe(scrRef.verseNum, position.verse)
+    : scrRef.verseNum === position.verseNum;
+}
+
+/** isVerseInRange throws on malformed ranges (reversed, 3-part) from imported USFM; a malformed
+ * range simply doesn't contain the verse. */
+function verseInRangeSafe(verseNum: number, verseRange: string): boolean {
```
lyonsil, line 446: **[22/18 - cleanup/reuse] One root cause, guarded twice in this file.**

`isVerseInRange` (`libs/shared/src/nodes/usj/node.utils.ts`) throws on malformed ranges. This file catches that single throw in two different ways: this `verseInRangeSafe` wrapper (used at lines 378 and 427), **and** a separate ad-hoc `try/catch` around `$findVerseOrPara` at lines 391-396 - which only throws because `$findVerseInNode` (`libs/shared-react/src/nodes/usj/node-react.utils.ts`) calls the same unguarded `isVerseInRange`. The comment above that try/catch admits as much.

So the guard is at the wrong altitude: two mechanisms here for one root cause, and `$findVerseOrPara` / `$findVerseInNode` stay throw-prone for every other consumer of the `shared-react` barrel.

Better: export an `isVerseInRangeSafe` sibling next to `isVerseInRange` in `shared`, have `$findVerseInNode` use it, then delete both this wrapper and the try/catch at 391-396. (Don't change `isVerseInRange` itself - its throw is pinned by `node-utils.test.ts`.)

Out of scope for a cleanup PR if you'd rather not - but this PR *is* the cleanup PR.

packages/platform/src/editor/ScriptureReferencePlugin.tsx
```diff
@@ -451,7 +260,119 @@ export default function ScriptureReferencePlugin({
...
+  if (Number.isNaN(chapterNum)) return undefined; // unaddressable - consistent with I5
+
+  const verseNode = $resolveVerseNode(startNode, selection);
+  const { verseNum, verse } = $getEffectiveVerseForBcv(verseNode ?? undefined, selection);
+  if (Number.isNaN(verseNum)) return undefined; // unaddressable - consistent with I5
+  return { book: bookNode?.getCode() || undefined, chapterNum, verseNum, verse };
+}
+
+/** First BookNode of the root (documents put it first; strays after content are ignored). */
+function $getFirstBookNode(): BookNode | undefined {
```
lyonsil, line 324: **[23/18 - cleanup/reuse] This hand-rolled sibling walk is `$getRoot().getChildren().find($isBookNode)`.**

Six lines of `getFirstChild()` / `getNextSibling()` for what the codebase already expresses as a one-line `find` with a type guard - which is the established idiom here: `$getFirstPara` (`node-react.utils.ts`), `$findChapter` and `$findNextChapter` (`node.utils.ts`) are all written that way, and **this plugin's own test file already writes exactly `$getRoot().getChildren().find($isBookNode)`** (`ScriptureReferencePlugin.test.tsx:471`).

`$moveCaretToVerseStart` in this same file already calls `$getRoot().getChildren()` (line 382), so the file now carries two different root-scan shapes for no behavioral gain.

packages/platform/src/editor/ScriptureReferencePlugin.tsx
```diff
@@ -451,7 +260,119 @@ export default function ScriptureReferencePlugin({
...
+ * an empty document). Content before the first chapter of a loaded document addresses as
+ * chapter 1 per USFM convention (its verse resolves to 0). */
+function $resolvePosition(): ResolvedPosition | undefined {
+  const selection = $getSelection();
+  const startNode = getSelectionStartNode(selection);
+  if (!startNode) return undefined;
+
+  // The unaddressable check needs the NODE's existence - a BookNode with an empty code still
+  // makes a document addressable.
+  const bookNode = $getFirstBookNode();
```
lyonsil, line 302: **[24/18 - cleanup/efficiency] `$resolvePosition` re-walks the root children on every selection change.**

This runs on every idle selection change: every caret move, every typed character, every click, continuously during a mouse drag, plus the synthetic dispatch this plugin's own verse listener fires on every verse created/destroyed batch. Each call re-derives the first BookNode by walking root children.

In a well-formed document the BookNode is the first child, so it's cheap. But for the book-less document this file explicitly supports (the I1 prop-book fallback), the loop visits **every** root child before returning undefined - one per para, ~40 for a typical chapter, 190+ for a Psalm-119-shaped one - on every caret move.

The value is invariant between BookNode mutations, and `registerMutationListener(BookNode, ...)` on line 196 is already the perfect invalidation hook - it computes exactly this value and throws it away. Caching it on `Machine` would delete the walk from the hot path entirely. (Would need invalidating on the destroy-only early return at line 198 too.)

Low priority - flagging it because the mechanism to fix it already exists 100 lines up.

packages/platform/src/editor/ScriptureReferencePlugin.tsx
```diff
@@ -451,7 +260,119 @@ export default function ScriptureReferencePlugin({
...
+function schedulePlacingCaretAtVerseStart(machine: Machine, editor: LexicalEditor) {
+  queueMicrotask(() => {
+    editor.update(
+      () => $moveCaretToVerseStart(machine.scrRef.chapterNum, machine.scrRef.verseNum),
+      { tag: CURSOR_CHANGE_TAG },
+    );
+  });
+}
+
+function $moveCaretToVerseStart(chapterNum: number, verseNum: number) {
```
lyonsil, line 388: **[25/18 - cleanup] The only helper in the file with no doc comment - and its first branch contradicts its name.**

Every other helper here carries a `/** */` (`report`, `consumePendingEcho`, `$resolvePosition`, `positionMatchesScrRef`, `verseInRangeSafe`, ...). This one has none, and lines 376-380 deliberately **do not move the caret** when it already sits inside a verse range containing `verseNum` (pinned by "should not move the cursor if already in range") - i.e. the function's first act is the opposite of what `$moveCaretToVerseStart` promises, with no explanation.

The other three no-op exits (no chapter, `$findVerseOrPara` throws, no target found) are unsurprising, so the name is fine - the gap is just the one-liner. Since this PR renamed this exact signature line, it seems like the moment to add it:

```ts
/** Moves the caret to the start of `verseNum` in `chapterNum`. No-op when the caret is already
 * inside a verse range containing `verseNum` (a range is one location), or the target is absent. */
```

lyonsil:
The previous comments were directly posted after a `/code-review max` pass. Regarding the "pre-existing bugs" note, most of them rely on the following:
- Books with missing IDs
- Documents with no book node
- Other plug-ins working differently than they do today

I'm double checking there aren't new, meaningful issues it says were latent. So far, I haven't found any.

It seems like Claude is most unhappy with the changes to remove `getEditorState()`. I didn't dig into each item in depth, but I did scan them to make sure they seemed plausible enough to consider.

packages/platform/src/editor/ScriptureReferencePlugin.tsx
lyonsil: **Update: I answered my own question - the trigger is NOT present. Downgrading this to latent.**

Rather than leave this hanging for you, I traced the host chain in `paranext-core`:

`platform-scripture-editor.web-view.tsx:1692` passes `onScrRefChange={setScrRefNoScroll}` -> `setScrRefNoScroll` (`:1127-1140`) -> `setScrRefWithScroll` from `useWebViewScrollGroupScrRef()` -> `useScrollGroupScrRef` (`use-scroll-group-scr-ref.hook.ts:99-114`) -> `setScrRefLocalIfDifferent` -> `setScrRefLocal`, a **plain `useState` setter** (`:67`).

`flushSync`, `useSyncExternalStore` and `unstable_batchedUpdates` appear **zero times** in the host's first-party source. The only `flushSync` in the dependency tree is Lexical's own `useDecorators`, which fires at the end of a commit rather than re-entrantly inside `onScrRefChange`. (`setScrRefSync` in `scroll-group.service-host.ts` is a red herring - "Sync" there means "not Promise-returning", nothing to do with React.) The feedback loop back into React is `useEvent`, a plain `useEffect` subscription, not `useSyncExternalStore`. The other three consumers (`resource-text-panel`, `model-text-panel`, `enhanced-resource`) go through the same hook.

This finding requires a re-entrant passive-effect flush **inside** an open Lexical update (`_updating === true`), and nothing in the current host produces one. **So it is latent, not live - please don't spend time on it.** It remains an argument for not using `editor.read()` here (the API gives you no guard if a host ever introduces a synchronous commit), but it is not a bug you can hit today.

**This does not affect findings 1 and 3 on this line.** Those need only `_pendingEditorState !== null` with `_updating === false` - i.e. after the stack has unwound - which the plain-`setState` host *does* produce: `onScrRefChange` fires from a `selectionchange`-driven Lexical command listener, React treats that as discrete/sync-lane and schedules its flush as a microtask, and that microtask can run before Lexical's own deferred commit microtask. Both still stand.

Sorry for putting an open question in front of you that was mine to answer.

lyonsil:
## Four pre-existing latent issues in `ScriptureReferencePlugin` (not introduced by this PR - none should block it)

While reviewing this PR I traced four defects that are **not this PR's doing**. They come from #499 and earlier; #500 only *moved* the code they live in, which is how they surfaced. I'm putting them here rather than inline so they don't clutter the diff.

**None of them causes a problem today.** Each needs a specific condition that no current code path or consumer produces (with one caveat on L1 below, which I'll be honest about). All four are reproduced against the unmodified plugin. Entirely your call whether to fold any in now or leave them for later - I'd merge this PR on its own merits regardless.

---

### L1. A BookNode with an empty code never gets the caret placed on navigation

`BookNode.getCode()` is typed `BookCode | ""` (`libs/shared/src/nodes/usj/BookNode.ts:90`), and the adaptor **deliberately tolerates** a bad code - `usj-editor.adaptor.ts:226-229` warns and then emits it anyway as `code ?? ("" as BookCode)`. Line 199 collapses that to `undefined`:

```ts
const code = editor.read(() => $getFirstBookNode()?.getCode() || undefined);   // "" -> undefined
```

and the arrival gate at line 329 requires a truthy code:

```ts
if (batch.hasCreated && code && code === machine.scrRef.book) {
  schedulePlacingCaretAtVerseStart(machine, editor);
}
```

so the placement never runs. Reproduced: control document coded `GEN`, navigating to GEN 2:9, caret lands on `"verse nine "`. Identical document coded `""`, caret lands on `"Test Book"` - the top of the document.

Because Platform.Bible documents are single-chapter, arrival-driven placement is the path **every** navigation takes, so in a project with a codeless `\id`, every navigation would leave the caret at the document top.

There's also a three-way inconsistency worth knowing about: the gate at line 271 (`!bookCode || ...`) treats an empty code as *"no book, so place anyway"*, line 329 treats it as *"not my document"*, and `$resolvePosition` (lines 296-300) takes a third position, keying addressability on the **node's existence** with a comment explaining why: *"a BookNode with an empty code still makes a document addressable."*

**Trigger:** a USJ `\id` with a missing or empty code. **Caveat: this is the one whose trigger is _data_-dependent rather than code-dependent**, so if such projects exist in the wild it is already live rather than latent. I couldn't determine whether they do - you'd know better than I would.

---

### L2. `sawDocument` is never set for a book-less document, so a later paste teleports the caret

`machine.sawDocument` is written only inside `onDocumentChanged`, which only ever runs from the BookNode mutation listener. Lexical's `initializeMutationListener` does not invoke a listener when no node of that class exists, so a document with **zero** BookNodes never fires it - `sawDocument` stays `false` even though a document *is* loaded.

A later cross-editor paste carrying a BookNode then arrives as `{hasCreated: true, hasDestroyed: false}` with phase `idle`, and line 343 matches:

```ts
} else if (batch.hasCreated && !machine.sawDocument) {
  // MOUNT: fresh editor, no navigation window needed
  schedulePlacingCaretAtVerseStart(machine, editor);
}
```

The plugin thinks it is mounting a fresh document and teleports the caret. Reproduced: caret jumped from `"bookless verse one "` to `"bookless verse three "`. Note `beforeinput` does **not** save you here - the guard on this path is `sawDocument`, not `phase`.

That contradicts the trade-off the file header calls pinned (lines 67-69): *"A cross-editor paste or undo that (re)creates a BookNode while a document is already loaded triggers neither placement nor a navigation window: pastes must not teleport the caret to verse start (pinned)."* The pinning test only covers the has-a-BookNode case.

**Trigger:** a USJ with no `book` element. Book-less documents are an explicitly *designed* input of this file (`ResolvedPosition.book` is "undefined when it has none"; the I1 prop-book fallback exists for exactly this), but I found no path in paranext-core that sends one and every fixture in the repo has a book. **If you can confirm the host always sends `\id`, this is dead code and you can dismiss it.**

---

### L3. A late mount over an already-loaded document emits a spurious verse-0 ref

The plugin is conditionally mounted (`Editor.tsx:422`) and both props are optional (`editor.model.ts:169-171`):

```tsx
{scrRef && onScrRefChange && <ScriptureReferencePlugin scrRef={scrRef} onScrRefChange={onScrRefChange} />}
```

So a host that writes `scrRef={isReady ? scrRef : undefined}` - a natural thing to write against optional props - mounts the plugin **over an already-loaded document**. Then, in a *single synchronous effects flush*:

1. The BookNode listener's `{ skipInitialization: false }` replay fires -> `onDocumentChanged` takes the **MOUNT branch** (line 343), which queues placement as a **microtask** and deliberately does **not** set `phase`. Same book, so no correction is emitted either. **Phase stays `"idle"`.**
2. The two verse listeners (lines 237-238) are registered with **no options** - and Lexical's `DEFAULT_SKIP_INITIALIZATION` is **`false`** (`Lexical.dev.mjs:10128`), so **they replay too**. `onVerseDestroyed` counts `"created"` as a trigger and fires `editor.dispatchCommand(SELECTION_CHANGE_COMMAND)` **synchronously**.
3. The command listener sees `phase === "idle"` -> `onSelectionSettled` -> the caret is at the chapter top -> resolves to **verse 0** -> `report()` emits.
4. **Only then** does the placement microtask run.

Steps 1-4 are one synchronous call stack, so `queueMicrotask` provably cannot run until after the emission. Reproduced against the unmodified plugin, with the assertion read on the line right after mount with **no `await` in between** (so no microtask can have run):

```
emitted: { book: "GEN", chapterNum: 1, verseNum: 0 }
```

That's the R2 verse-0 clobber, at mount, before placement.

The registration-order guard only half-saves this. It works for the **cross-book** late mount (the mount correction sets `phase = "navigating"` before the verse replay dispatches). But in the **same-book** case - the common one - no correction is emitted and the MOUNT branch never flips the phase, so ordering buys nothing. The header's note about *"a document swap must flip phase to navigating here before the verse listener's synchronous dispatch"* is about the REPLACEMENT branch; the MOUNT branch has no such flip.

**Why it's latent:** `Editor.tsx:172` passes `editorState: undefined`, so the composer is empty at plugin mount and both replays are no-ops (measured: 0 dispatches, 0 emissions). The document then arrives via `LoadStatePlugin`, whose new state carries a `null` selection, so the verse dispatch resolves to nothing and the placement microtask wins. And no current consumer late-mounts - paranext-core always passes `scrRef` from `useWebViewScrollGroupScrRef()`, which is never undefined.

**Trigger:** any host that gates the props on a loading state. **Also: Vite HMR already does this in dev** - editing the plugin module remounts it over a loaded document. If anyone has ever seen a phantom verse-0 after a hot reload, this is why.

The root cause is an implicit/explicit asymmetry: the BookNode listener opts into the replay **explicitly** (`{ skipInitialization: false }`), while the verse listeners 35 lines below get the same replay **implicitly** from Lexical's default - which reads as though they *don't* replay. The file's own UNSAFE list headlines that replay path as the cause of the NUM 27:1 bug.

**Cheapest fix:** have `onVerseDestroyed` ignore the registration-time replay (skip when the update tags indicate `registerMutationListener`, or pass `{ skipInitialization: true }` if the replay isn't wanted there), which kills the dispatch at source without touching any of the six pinned races. Alternatively, have the MOUNT branch set `phase = "navigating"` when a non-matching selection already exists.

---

### L4. REPLACEMENT detection is load-bearing on destroy+create arriving in ONE mutation batch

Line 198 discards a destroy-only batch outright, recording nothing that a destroy happened:

```ts
if (kinds.every((kind) => kind === "destroyed")) return;
```

`sawDocument` is a one-way latch, and REPLACEMENT is detected only at line 336 (`batch.hasCreated && batch.hasDestroyed`). So if a document swap ever arrived as **two commits** (a clear-then-fill), batch 1 dies at line 198 and batch 2 is create-only with `sawDocument === true`, falling through *both* branches into the trailing "cross-editor paste" case: **no placement, and `phase` is never set to `"navigating"`** - so the refill's own settle is reported as a user move at verse 0.

Reproduced: the single-batch control (`setEditorState`) emits nothing; split across two commits it emits `{GEN, 1, 0}` while the host is at 1:2.

**Trigger: none today.** I checked all four candidate paths and every one is single-commit or unmounts instead:
- `LoadStatePlugin` wraps one `editor.setEditorState()` - a single `BookNode:destroyed,created` batch (verified).
- Undo/redo restores via a single `setEditorState`, and `CLEAR_HISTORY_COMMAND` fires after every load anyway, so undo never spans a swap.
- No production code calls `$getRoot().clear()`.
- The host **unmounts** the editor rather than blanking it - `renderEditor()` swaps in a spinner or an error div when the usj is empty or the book doesn't exist, and a remount is a MOUNT, not a split swap.

So this is an undocumented invariant rather than a defect. I'm recording it because a split swap would be silently **misclassified as the documented "cross-editor paste" branch**, which is exactly the sort of thing a future `LoadStatePlugin` change could reintroduce with no test catching it. A `sawDestroy` flag set by the destroy-only batch and consumed by the next create-only batch would close it.

---

Happy to open separate issues for any of these, or to contribute failing tests (I have working repros for all four), if that's more useful than a comment. And happy to be told L2 and L4 are unreachable by construction - you'd know the host's data guarantees better than I do.

packages/platform/src/editor/ScriptureReferencePlugin.tsx
irahopkinson: Fixed in 36bceac8 - both sites reverted to `editor.getEditorState().read()` via a `getCommittedBookCode` helper that documents the committed-snapshot requirement.

irahopkinson: Resolved by the revert in 36bceac8 - `editor.read()` is gone from the file, and the header's UNSAFE list now bans it. Agreed it's latent given the host's plain `useState` setter chain; the revert removes the exposure either way.

irahopkinson: Fixed in 36bceac8 - the gate reads the committed snapshot again (`getCommittedBookCode`), so evaluating it can no longer install the document it then approves.

packages/platform/src/editor/ScriptureReferencePlugin.tsx
irahopkinson: Fixed in 36bceac8 - with the flush gone, the caret move merges into the still-pending document state again, so the document and caret commit together.

packages/platform/src/editor/ScriptureReferencePlugin.tsx
irahopkinson: Fixed in 36bceac8 - reverted via the shared `getCommittedBookCode` helper, which documents why `editor.read()` is off-limits here (including the skipInitialization replay re-entry).

irahopkinson: Resolved by the revert in 36bceac8. The UNSAFE list entry added there documents the `editor.read()` hazard so this ordering dependency can't come back silently.

irahopkinson: Took the suggested shape: `getCommittedBookCode` landed in 36bceac8, and c8459c73 finishes the rename - `onDocumentChanged`'s parameter and the listener local are now `bookCode`.

packages/platform/src/editor/ScriptureReferencePlugin.tsx
irahopkinson: Addressed in c8459c73 with the lighter variant: the block stays on `onSelectionSettled`, and the call site now carries an inline warning that `$resolvePosition` must stay inline there, pointing at that doc.

packages/platform/src/editor/ScriptureReferencePlugin.tsx
irahopkinson: Resolved by the revert - the contradiction was with the `editor.read()` usage, which is gone. The actual rule now lives on `getCommittedBookCode`'s doc and the header's UNSAFE entry.

packages/platform/src/editor/ScriptureReferencePlugin.tsx
irahopkinson: The UNSAFE entry landed in 36bceac8. Passing on the ESLint rule for now - introducing the repo's first `no-restricted-*` rule goes too far afield for a cleanup PR; happy to revisit separately.

packages/platform/src/editor/ScriptureReferencePlugin.tsx
irahopkinson: Moved next to `report()` with a TSDoc in c8459c73. For what it's worth, SCREAMING_SNAKE already signals a module const and hover gives the value - the missing doc comment was the real gap.

packages/platform/src/editor/ScriptureReferencePlugin.tsx
irahopkinson: Labelled the branches inline in c8459c73 - `(a) REPLACEMENT`, `(b) MOUNT`, `(c)` paste/undo, `(d) Book correction` - so the existing references resolve.

packages/platform/src/editor/ScriptureReferencePlugin.tsx
irahopkinson: Took the suggested `isFirstDocument` shape in c8459c73 - one write site, hoisted above the navigating branch (which also collapses the duplicate write there), with the read captured explicitly first. Tests stay green.

packages/platform/src/editor/ScriptureReferencePlugin.tsx
irahopkinson: Leaving as is for this PR. Scribe's `ScriptureReferencePlugin` is already a named export under the same name, so named exports are the established direction here. Finishing the migration (`TreeViewPlugin`, `CommentPlugin`, `Editor`) and dealing with the legacy shared-react plugin (whose only consumer is the perf-react demo) belongs in a follow-up.

packages/platform/src/editor/ScriptureReferencePlugin.tsx
irahopkinson: Agreed on the direction, but moving the guard into shared/shared-react changes malformed-range behavior for scribe and the demos too - that deserves its own PR and tests rather than riding on this one. Leaving both local guards for now.

packages/platform/src/editor/ScriptureReferencePlugin.tsx
irahopkinson: Declining, with context the repo doesn't show: the host splits USJ into chapters and loads one at a time, and only chapter 1's document carries a BookNode - so the full-walk miss is actually the *common* case here, not the book-less edge case. It stays bounded by paras-per-chapter (~40, ~190 for Psalm 119), i.e. microseconds per selection event. And the proposed Machine cache can't invalidate on chapter N -> N+1 swaps, which create/destroy no BookNode and fire no mutation at all - a stale-state hazard in exactly the file that exists to contain them.

packages/platform/src/editor/ScriptureReferencePlugin.tsx
irahopkinson: Added in c8459c73, essentially verbatim.

irahopkinson:
L1: the `code` attribute is required for `/id` (BookNode) in USFM but I have no idea if projects exist that don't have it for any book.
L2: we mostly display per-chapter so we split USJ per chapter and send to the editor so only chapter 1 has a `BookNode` and all the rest are bookless. With the new Text Collection in Simple we are also using the editor component to display a verse at a time so definitely bookless also.
L3 & L4: are your call @lyonsil since you know this new state machine better than anyone.

packages/platform/src/editor/ScriptureReferencePlugin.tsx
irahopkinson: Done.

irahopkinson (commented):
> @irahopkinson made 3 comments and resolved 2 discussions.
> *[Reviewable](https://reviewable.io/reviews/eten-tech-foundation/scripture-editors/500)* status: 2 of 3 files reviewed, 18 unresolved discussions (waiting on katherinejensen00 and lyonsil).

___
*[`packages/platform/src/editor/ScriptureReferencePlugin.tsx` line 340 at r1](https://reviewable.io/reviews/eten-tech-foundation/scripture-editors/500#-OxW0T0I1itmmf7zKUut:-OxY1OVMD9O2g9G8peUl:b-896fix) ([raw file](https://github.com/eten-tech-foundation/scripture-editors/blob/97172c85a6b617bef5df3a6dcc58f0ae4cef3d30/packages/platform/src/editor/ScriptureReferencePlugin.tsx#L340)):*

NIT Claude Finding

**The `(a)`–`(d)` cross-references in `onDocumentChanged` are used in prose but never anchored to their code branches.**

The REPLACEMENT branch says "The correction below (d), if any, runs after." The book correction block says "Runs after (a)–(c); under (a) phase is already 'navigating' here." But no comment in the code is actually labeled `// (a)`, `// (b)`, etc. — a reader has to manually count branches to decode the letter scheme. If (a) is REPLACEMENT, (b) is MOUNT, (c) is the paste/undo no-op, and (d) is the book correction block, that can be inferred, but it takes real effort. This predates your PR, so it doesn't need to block merge — but a one-liner label at the start of each branch (`// (a) REPLACEMENT: ...`, etc.) would make the cross-references self-evident.
</blockquote></details>

Done.
___
*[`packages/platform/src/editor/ScriptureReferencePlugin.tsx` line 375 at r1](https://reviewable.io/reviews/eten-tech-foundation/scripture-editors/500#-OxW0Iwt8iOI613Zaf7S:-OxY1Klt6EzgSE0U0AZY:b-896fix) ([raw file](https://github.com/eten-tech-foundation/scripture-editors/blob/97172c85a6b617bef5df3a6dcc58f0ae4cef3d30/packages/platform/src/editor/ScriptureReferencePlugin.tsx#L375)):*

NIT Claude Finding
**`$moveCaretToVerseStart` is the only named handler in the file without a JSDoc, and it's the most complex one.**

Every other handler in the file — `onPropChanged`, `consumePendingEcho`, `$resolvePosition`, `onDocumentChanged`, `schedulePlacingCaretAtVerseStart`, `onSelectionSettled`, `positionMatchesScrRef`, `verseInRangeSafe`, `positionToScrRef`, `report`, `refsEqual`, `onUserInteracted` — has a function-level JSDoc comment. `$moveCaretToVerseStart` doesn't, and it's the function with the most state assumptions: it must be called inside a Lexical editor-state update, it has an early-return when the caret is already in a verse range, and it handles the "no verse node, fall back to para" case. The file-top HOW-THIS-PLUGIN-WORKS block explains the overall algorithm but doesn't substitute for a function signature–adjacent description that callers read in context. Since this PR renamed the function and moved it, this is a good moment to close that gap with even two or three sentences.
</blockquote></details>

Done.

lyonsil (commented):
I included L3 and L4 out of completeness, but I don't think either is strictly needed. They both require the plug-in and editor to do things that neither does today. Fixing L4 is also potentially part of a larger refactor to slightly simplify the state machine further that I previously chose to defer given time constraints and complexity.

:lgtm: - Thanks for the clean ups! I think these updates will make the new plug-in code easier to understand.

> @lyonsil reviewed 3 files and all commit messages, made 1 comment, and resolved 18 discussions.
> *[Reviewable](https://reviewable.io/reviews/eten-tech-foundation/scripture-editors/500#-:-Ox_u9qxAyJSayoighJ7:bq6hsx9)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved.

PR #505, opened 2026-07-20
fix: allow whole verses deletion to be undone

mattgetgen:
This change is [Reviewable](https://reviewable.io/reviews/eten-tech-foundation/scripture-editors/505)


irahopkinson (commented):
Reviewed the verse-delete undo fix. Deferring the `SELECTION_CHANGE_COMMAND` dispatch out of the commit phase is the right call, and I confirmed it produces a clean single-undo entry: the deferred update has no dirty nodes, so `HistoryPlugin` takes the `HISTORY_MERGE` path rather than pushing a spurious history step. Three comments inline.

One process question, meant collegially: this file opens with "HOW THIS PLUGIN WORKS — READ BEFORE CHANGING ANYTHING." Did you get to read it through, and does the change sit comfortably with the invariants it documents — particularly the registration-order reasoning at lines 190–193 (flagged inline)?

packages/platform/src/editor/ScriptureReferencePlugin.tsx
```diff
@@ -235,7 +235,12 @@ export function ScriptureReferencePlugin({
       const hasCreatedOrDestroyedVerse = [...nodeMutations.values()].some(
         (m) => m === "created" || m === "destroyed",
       );
-      if (hasCreatedOrDestroyedVerse) editor.dispatchCommand(SELECTION_CHANGE_COMMAND, undefined);
+      // Defer one microtask: dispatching synchronously inside a mutation listener runs a reentrant
```
irahopkinson, line 241: This new deferral comment is clear, but it leaves the older comment at lines 190–193 stale. That one still describes the verse listener's `SELECTION_CHANGE_COMMAND` dispatch as *synchronous* and says reordering the two mutation listeners "reintroduces the R2 clobber." Now that the dispatch is `queueMicrotask`-deferred it runs after both mutation listeners complete, so R2-safety no longer depends on registration order. Could you update 190–193 to match the deferred model? Worth keeping accurate given the file's "READ BEFORE CHANGING ANYTHING" header.

packages/platform/src/editor/ScriptureReferencePlugin.tsx
```diff
@@ -235,7 +235,12 @@ export function ScriptureReferencePlugin({
...
+      // update while the triggering edit's history entry is still being recorded, corrupting the
+      // undo stack (PT-4102: undo did nothing after a verse-spanning delete). Same deferral rule as
+      // schedulePlacingCaretAtVerseStart below.
+      if (hasCreatedOrDestroyedVerse)
+        queueMicrotask(() => editor.dispatchCommand(SELECTION_CHANGE_COMMAND, undefined));
```
irahopkinson, line 251: Request (non-blocking): a short comment or guard noting the invariant this now relies on. Deferring to a microtask trades a self-limiting failure mode for a non-self-limiting one. Today nothing loops, because no `SELECTION_CHANGE_COMMAND` listener creates or destroys a verse node. But if one were ever added, the cycle (verse mutation → microtask → dispatch → mutate verse → …) would spin a fresh microtask each iteration, so the call stack unwinds every hop and Lexical's "Maximum update depth exceeded" guard never trips — the tab would silently freeze, whereas the old synchronous dispatch would at least throw visibly. A one-line note that no selection-change listener may create/destroy verse nodes would protect the next person.

packages/platform/src/editor/Editor.test.tsx
```diff
@@ -178,3 +180,99 @@ describe("setAnnotation overload", () => {
...
+  await act(async () => {
+    editor.dispatchCommand(
+      KEY_DOWN_COMMAND,
+      new KeyboardEvent("keydown", { key: "Delete", bubbles: true, cancelable: true }),
+    );
+  });
+}
+
+/** Lets queued microtasks (e.g. the deferred scripture-reference recompute) settle. */
+async function flushMicrotasks(): Promise<void> {
```
irahopkinson: Minor / non-blocking: `flushMicrotasks` here is byte-for-byte identical to `flushQueuedEvents` in `ScriptureReferencePlugin.test.tsx`. Could be extracted into a single shared test helper.

irahopkinson (commented):
> @irahopkinson reviewed 2 files and all commit messages.
> *[Reviewable](https://reviewable.io/reviews/eten-tech-foundation/scripture-editors/505)* status: all files reviewed, 3 unresolved discussions (waiting on mattgetgen).

irahopkinson (approved):
:lgtm: but I'd be happier if you did a `nx devpub platform-editor` from this branch and then in core `npm run build` and then `npm start` and tested it well in core. You are free to merge either way.

> @irahopkinson reviewed 4 files and all commit messages, made 1 comment, and resolved 3 discussions.
> *[Reviewable](https://reviewable.io/reviews/eten-tech-foundation/scripture-editors/505#-:-OypsZFaBedVRlNRspwV:bkgsjr)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved.

mattgetgen (commented):
I just verified that running those commands applies the changes in paranext-core, and it worked! Merging.

> @mattgetgen made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/eten-tech-foundation/scripture-editors/505#-:-OysVzCsCrqKQpc59xki:b-synemh)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved.

PR #510, opened 2026-07-27
docs: document the dependency security sweep procedure

irahopkinson:
Adds a written procedure for dependency updates and a Claude Code skill that follows it, plus two corrections to overrides already on `main`. Branched off `main`, independent of #512.

## Why

**Dependabot here only opens security PRs.** `.github/dependabot.yml` sets `open-pull-requests-limit: 0`, which disables version-update PRs while leaving security updates on. So there are no routine bumps — but the titles read exactly like routine bumps, and two sat unmerged for five weeks.

**Bumping a package doesn't necessarily close its advisory.** pnpm only re-resolves a transitive dep when the locked version stops satisfying its parent's range. Nothing in CI, the pre-commit hook, or any npm script runs `pnpm audit`, so a green build is not evidence either way.

## Two live instances, both found while writing this

Worth reading before reviewing — they're the reason the document exists, and both are in code already merged.

**#509 claimed to supersede #487 and did not close its advisory.** GHSA-4x5r-pxfx-6jf8 had two instances: the root manifest at 7.29.0, and a transitive `demos__perf-react > epitelete-html > @babel/core@7.28.3`. The bump fixed the first and left the second. The audit was checked by *total count* (45 → 30), so a still-open advisory hid inside a falling number. #512 closes it. This is why step 1 now baselines the advisory-**ID set**, not counts.

**An override key matches the range a parent _declares_, not the installed version.** `@svgr/plugin-svgo@8.1.0` declares `svgo: ^3.0.2`, but installed svgo is **4.0.1**. The override `svgo@>=3.0.0 <3.3.3: ">=3.3.3"` did that — its key never contained 4.0.1, it matched the declared `^3.0.2`, and the uncapped value then walked svgo across a major under a plugin asking for 3.x.

That one is **inert and needs no action**: svgo arrives only via `@svgr/webpack` ← `@nx/react`, and svgr is referenced nowhere in this Vite repo, so it never executes. But it means "cap the value" describes damage already done rather than a hypothetical, and #512 caps it at `<5`.

## What's here

- **`docs/dependency-updates.md`** — the single source of truth. Full-sweep and scoped modes, the procedure, how to write overrides, how to move a floor, how to retire stale ones, and why green CI isn't evidence.
- **`.claude/skills/update-dependencies/SKILL.md`** — the on-demand entry point for agent-driven runs. Deliberately does *not* repeat the procedure; it points at the doc, so each fact lives in one place.
- **`README.md`** — a short section linking both.
- **`pnpm-workspace.yaml` + `pnpm-lock.yaml`** — caps the `form-data` override from `">=4.0.6"` to `">=4.0.6 <5"`. It was uncapped and would have resolved to form-data 5.x on release. Resolution unchanged; the lockfile moves by one line.
- **`cspell.json`** — ignore GHSA identifiers by pattern rather than word by word.

## How it was validated

Four trial runs, each a fresh agent in an isolated worktree doing real work against real advisories — two full sweeps, two scoped to named PRs. **Every finding in the doc came from a run failing or struggling, not from review.** Among them:

- No step connected a PR to its advisory; titles name versions, not CVEs.
- Nothing said to check the PR's own lockfile diff before trusting it.
- Step 10's close condition was wrong — "reaching the version they propose" never happens when the fix was an override, which is the usual fix here. `@dependabot rebase` is the answer, and only once the fix is on `main`.
- The procedure described only a whole-repo sweep, so scoped work had no expression in it. Hence "Two modes".
- Two of four agents abandoned the verify step believing it had hung, so it now states the runtime and that nx buffers all output.

Two convergence checks worth noting: the two full sweeps, run independently against *different revisions* of the doc, produced byte-identical dependency resolutions; and the two scoped runs independently reached the same decision on #487. The doc appears to determine the outcome rather than the model improvising.

`nx format:check` and `cspell` both pass.

## Deliberate omission

No CI audit step. `pnpm audit` currently exits 126 with findings (0 when clean, so it *is* usable as a gate), but it can't be added without either a threshold or clearing the existing findings first — and sweeps are meant to be occasional rather than enforced per-PR. The doc says plainly that the manual re-audit is therefore the only gate.

## Merge order

Take this before #512 — its `pnpm-lock.yaml` change is one line against #512's ~980, so rebasing in that direction is far cheaper. #512 doesn't touch the `form-data` line, so this cap survives intact.

- - -
This change is [Reviewable](https://reviewable.io/reviews/eten-tech-foundation/scripture-editors/510)

### DIFF
--- a/.claude/skills/update-dependencies/SKILL.md
+++ b/.claude/skills/update-dependencies/SKILL.md
@@ -1,0 +1,32 @@
+---
+name: update-dependencies
+description: Run a dependency security sweep in this monorepo. USE WHEN the user wants to update dependencies, handle the open dependabot PRs, clear pnpm audit findings, or do a security pass before a release.
+---
+
+# Update Dependencies
+
+Read [`docs/dependency-updates.md`](../../../docs/dependency-updates.md) now and follow its
+**Procedure** section exactly, creating a todo per step. That doc is the single source of truth —
+this file deliberately does not repeat it. Don't improvise a shortcut; the steps encode two failures
+that have already happened in this repo.
+
+## Before you start
+
+Two things to get right, both explained in full in the doc:
+
+- **Every dependabot PR here is a security fix**, not a routine bump. The titles don't say so.
+- **A version bump does not necessarily close the advisory.** pnpm won't re-resolve a transitive dep
+  whose locked version still satisfies its parent's range. Never report a bump as a fix without
+  re-running `pnpm audit` and confirming that specific finding is gone.
+
+## Reporting
+
+Give the user:
+
+- before/after `pnpm audit` counts,
+- the version taken per package versus what dependabot proposed,
+- any advisory still open, and why.
+
+Report the real output of the verify step — a lockfile change can break things no unit test covers.
+And if an advisory could not be cleared, say so outright rather than letting a passing build imply
+otherwise: CI does not run `pnpm audit`, so green checks are not evidence here.
--- a/README.md
+++ b/README.md
@@ -211,6 +211,18 @@ Use Prettier directly rather than `nx format:write`. CI runs `nx format:check` p
 
 Note `.prettierignore` excludes some files that still get staged, notably `pnpm-lock.yaml` and `**/tsconfig*.json`. Prettier leaves those alone by design; that isn't the hook failing.
 
+## Dependency Updates
+
+Dependabot is configured to open **security PRs only**, so any dependabot PR on this repo is an
+advisory rather than a routine bump. We supersede those PRs with a single hand-made batch commit,
+because dependabot resolves pnpm monorepo trees poorly and because a version bump often fails to
+clear the advisory on its own.
+
+See [docs/dependency-updates.md](/docs/dependency-updates.md) for the procedure, how to write the
+`pnpm-workspace.yaml` overrides, and why a green CI run is not evidence that a vulnerability is
+fixed. Claude Code users can invoke the
+[`update-dependencies`](/.claude/skills/update-dependencies) skill, which follows that doc.
+
 ## TypeScript Code Intelligence
 
 This repo gives Claude Code a TypeScript language server (jump-to-definition, find-references, and
--- a/cspell.json
+++ b/cspell.json
@@ -15,6 +15,7 @@
   ],
   "dictionaryDefinitions": [],
   "dictionaries": [],
+  "ignoreRegExpList": ["GHSA(-[a-z0-9]{4}){3}"],
   "words": [
     "charis",
     "codesandbox",
@@ -23,6 +24,7 @@
     "deuterocanonical",
     "epitelete",
     "eten",
+    "ghsa",
     "hopkinson",
     "nrsv",
     "paratext",
--- a/docs/dependency-updates.md
+++ b/docs/dependency-updates.md
@@ -1,0 +1,279 @@
+# Dependency Updates
+
+## Every dependabot PR here is a security fix
+
+`.github/dependabot.yml` sets `open-pull-requests-limit: 0`, which disables version-update PRs while
+leaving security updates enabled. So this repo never receives a routine bump from dependabot — if a
+dependabot PR is open, an advisory is behind it.
+
+This matters because the titles don't say so. A PR reading `build(deps-dev): bump @babel/core from
+7.29.0 to 7.29.6` looks like housekeeping and is easy to leave sitting — two such PRs, opened in June
+2026, were still unmerged five weeks later.
+
+## When to run a sweep
+
+Before a release, or periodically. There is deliberately no scheduled audit job and CI does not run
+`pnpm audit` (see [CI does not audit](#ci-does-not-audit) below), so this is a manual, intentional
+activity rather than something that nags.
+
+## Why we don't just merge the dependabot PRs
+
+1. **Dependabot resolves pnpm monorepo trees poorly.** This is the original reason for handling them
+   by hand.
+2. **Its proposed version goes stale.** In the July 2026 sweep, three of four PRs were behind the
+   latest patch in their own major by the time anyone looked.
+3. **Its green checks go stale too.** An older PR ran CI against a much older `main`.
+4. **A dependabot bump often doesn't actually close the advisory.** This is the important one.
+
+## The trap: bumping a parent doesn't lift its locked transitive deps
+
+Worked example from the July 2026 sweep. Dependabot proposed axios `1.16.1` → `1.18.0` in
+`tools/usfm-markers`; we took `1.18.1`, the latest in that major. Either version fixes the axios
+advisories. But a _separate_, higher-severity finding —
+[GHSA-hmw2-7cc7-3qxx](https://github.com/advisories/GHSA-hmw2-7cc7-3qxx), CRLF injection in
+form-data, CVSS 8.7 — is reached through `tools__usfm-markers>axios>form-data`, and bumping axios
+left it wide open.
+
+The reason: axios 1.18.1 still declares `form-data: ^4.0.5`, and the already-locked 4.0.5 continues
+to satisfy that range. **pnpm only re-resolves a transitive dep when the locked version stops
+satisfying its parent's range.** A newer parent with an unchanged range changes nothing downstream.
+
+Clearing it took an explicit override:
+
+```yaml
+form-data@>=4.0.0 <4.0.6: ">=4.0.6 <5"
+```
+
+This is why the `overrides:` list in `pnpm-workspace.yaml` is long: most entries pin a transitive
+dependency that no parent bump would have reached. A few are different — `axios@<1.15.2` guards a
+_direct_ dependency, acting as a version floor rather than reaching past a parent. PR #477 added four
+entries in one pass, of both kinds.
+
+**The practical rule: never infer that a security bump worked. Re-run `pnpm audit` and confirm the
+specific advisory is gone.**
+
+## The same trap applies to overrides themselves
+
+The mechanism above is not special to parent ranges. It applies just as much to an override you
+already wrote, and this is the part that most often wastes a cycle.
+
+**An override key is matched against the range a parent _declares_, not against the version that is
+installed.** `pnpm audit` reports resolved versions, so the version it prints is usually the wrong
+thing to key on.
+
+Worked example from this repo, since fixed in #522. `@svgr/plugin-svgo@8.1.0` declares
+`svgo: ^3.0.2`, yet the installed svgo was **4.0.1**. What did that? The override:
+
+```yaml
+svgo@>=3.0.0 <3.3.3: ">=3.3.3"
+```
+
+The key `>=3.0.0 <3.3.3` does not contain 4.0.1. It does overlap the declared `^3.0.2`, which is what
+it actually matched. Then the uncapped value `">=3.3.3"` resolved to the newest svgo in existence and
+walked the package across a major boundary, under a plugin that asked for `^3.0.2`. Nobody reviewed
+that; it just happened.
+
+Two consequences worth remembering:
+
+- **Adding a second override keyed on the version `pnpm audit` printed usually does nothing.** If the
+  package already has an entry, that entry is the one in play. Raise its value; don't add a sibling.
+- **An uncapped value stops protecting you the moment the next advisory lands.** `pnpm audit` will
+  still report it — the override doesn't hide anything, since audit reads resolved versions from the
+  lockfile and never consults override ranges. What's missing is the automatic fix: the locked
+  version keeps satisfying `">=3.3.3"`, so `pnpm install` has no reason to re-resolve, and the
+  advisory sits there until someone runs a manual sweep. Given nothing in CI runs `pnpm audit`, "sits
+  there until someone looks" can be a long time. That's why capping isn't tidiness.
+
+## Two modes
+
+**Full sweep** — clear everything you reasonably can. Baseline and target are whole-repo counts.
+
+**Scoped run** — close the advisories behind one or more named PRs and _nothing else_. The
+deliverable is not a lower audit total; it is a specific set of GHSA IDs disappearing with no other
+change to the advisory set.
+
+Everything below applies to both. Steps marked _(scoped)_ differ.
+
+## Procedure
+
+1. **Baseline.** Capture the advisory-ID set, not a count:
+
+   ```bash
+   pnpm audit --json > tmp/audit-before.json   # tmp/ is gitignored
+   node -e "var a=require('./tmp/audit-before.json').advisories; for (var k in a) console.log(a[k].github_advisory_id)" | sort -u > tmp/ids-before.txt
+   ```
+
+   `.advisories` is keyed by pnpm's own numeric IDs, so the command reads each entry's
+   `github_advisory_id` field rather than the keys.
+
+   **Don't rely on counts.** One audit run yields three numbers that all disagree — at the time of
+   writing, a banner total of 38, 36 records in the JSON `advisories` map, and 34 unique GHSA IDs
+   (some advisories span two disjoint majors and appear twice). Worse, the totals drift on their own
+   as advisories are published: the same unchanged lockfile measured 30 one day and 39 a few days
+   later. A falling number is not evidence you fixed anything, and a steady one is not evidence you
+   didn't break something.
+
+2. **List the work.** `gh pr list --author app/dependabot --json number,title`
+3. **Identify the advisory behind each PR.** The title won't tell you — it names a version, not a
+   CVE. Use:
+
+   ```bash
+   gh api repos/eten-tech-foundation/scripture-editors/dependabot/alerts --paginate \
+     -q '.[] | select(.state=="open") | [.security_advisory.ghsa_id, .security_advisory.severity,
+         .dependency.package.name, .security_vulnerability.vulnerable_version_range,
+         .security_vulnerability.first_patched_version.identifier] | @tsv'
+   ```
+
+   This query is for attributing a PR to an advisory. **It is not the work list for a full sweep** —
+   it returned 23 alerts while `pnpm audit` found 28 records, missing five entirely. For a full
+   sweep, `pnpm audit` is authoritative.
+
+   The dependabot branch name encodes the version that _originally_ patched the advisory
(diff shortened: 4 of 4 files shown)


katherinejensen00 (changes requested):
Claude found 7 Warning level (lower than a Minor) findings to consider. Since they are so low, I thought I would run them by you but leave all but one as non blocking just so you have time to view them before merging. Thanks for improving the dependency management process!

> @katherinejensen00 reviewed 6 files and all commit messages, and made 8 comments.
> *[Reviewable](https://reviewable.io/reviews/eten-tech-foundation/scripture-editors/510#-:-OydH9E2BPqPsIwRlhQp:bkwjeie)* status: all files reviewed, 7 unresolved discussions (waiting on irahopkinson).

___
*[`pnpm-workspace.yaml` line 38 at r1](https://reviewable.io/reviews/eten-tech-foundation/scripture-editors/510#-OydD68DAA_2ndtSNw3g:-OydD68DAA_2ndtSNw3h:b-al523u) ([raw file](https://github.com/eten-tech-foundation/scripture-editors/blob/4a324212d408ac9d902ca45bc5b69c00aab6f6a1/pnpm-workspace.yaml#L38)):*
NIT Claude Finding - Warning (Not sure if this is right or if it is within the scope of this PR since this line didn't change)

### The document prohibits exact-version override pins, then ships with one unacknowledged in the file it governs

**Summary**: The doc states a hard rule against exact-version pins, explains why they don't work, and the repo already has one that contradicts it — with no note in either place.

The document is explicit:

> **Don't use exact-version overrides** (`brace-expansion@5.0.4: "5.0.5"`). They create a redirect while leaving the vulnerable package definition in the lockfile, where `pnpm audit` still finds it.

`pnpm-workspace.yaml` in this same PR contains:

```yaml
rollup@<4.59.0: "4.59.0"
```

This is the exact anti-pattern the doc prohibits, using the exact syntax the doc calls out (key is a range, value is a single version). The `cosmiconfig>yaml: "2.8.3"` entry also uses an exact version but is the parent→child form the doc explicitly allows as a fallback — that one is fine. The rollup entry is not that form.

A senior developer reading the document and then auditing `pnpm-workspace.yaml` against it will find this immediately. There are two plausible explanations — the rollup pin predates this rule and was never updated, or there was a specific reason at the time an exact pin was the only option — but neither is documented. The silence is worse than either explanation because it looks like an oversight rather than a deliberate exception.

The risk is concrete: per the doc's own description, `pnpm audit` will still flag advisories against rollup 4.59.0 even with this override, because the vulnerable package definition remains in the lockfile. If rollup 4.59.0 ever has an advisory, this override provides no protection and would mislead a developer into thinking it does.

**Fix**: Either replace the exact-version pin with a range form (`">=4.59.0 <5"`) or add a comment in `pnpm-workspace.yaml` explaining why this entry is an exception. The document could also briefly note that pre-existing exact-version pins from before this rule was established are candidates for cleanup in the next sweep.
___
*[`docs/dependency-updates.md` line 225 at r1](https://reviewable.io/reviews/eten-tech-foundation/scripture-editors/510#-OydGs-B5BLTLtvzIbAz:-OydGs-B5BLTLtvzIbB-:b-lhxtdq) ([raw file](https://github.com/eten-tech-foundation/scripture-editors/blob/4a324212d408ac9d902ca45bc5b69c00aab6f6a1/docs/dependency-updates.md#L225)):*
NIT Claude Finding (sounds bad, but I am not sure that it is right or if it is failing confidently. I had Claude double check the finding, and Claude still thought this finding was real, but I'm not sure if that is true)

**Summary**: The document tells developers to manually cap `pnpm audit --fix` output — but `--fix` already caps at the major. The entries that actually need manual capping are the ones written by hand.

The document says:

> `pnpm audit --fix` writes uncapped values, which is why most existing entries have no upper bound. Treat its output as a first draft and add the cap by hand.

Both claims are wrong. `pnpm audit --fix` writes **caret ranges** (`^X.Y.Z`), not bare `>=X.Y.Z`. This is the deliberate design: pnpm's source function is named `caretRangeForPatched()`, and its inline comment explains why — "`>=X.Y.Z` alone can silently promote a dep to a later breaking major; `^X.Y.Z` still satisfies the patch while preserving the major." The test suite asserts `'^0.18.1'`; the official docs show `"lodash@<2.1.0": "^2.1.0"`. Caret ranges already cap at the next major — the precise protection the doc says you have to add by hand.

The existing `>=` overrides in `pnpm-workspace.yaml` were written manually (or by an earlier unrelated tool), not by `--fix`. The doc's explanation reverses cause and effect.

This matters in practice for two reasons:

1.  A developer who runs `pnpm audit --fix` and sees `^` prefixes checks this document and finds an explanation describing `>=` syntax. The mismatch creates doubt about whether the warning applies — at the exact moment they're deciding whether to edit the output.
    
2.  The actionable advice points at the wrong target. `--fix` output is already safe at the major boundary. The entries that _do_ need manual review and capping are the hand-written `>=` entries already in the file — the ones with no upper bound that the doc correctly warns about elsewhere. Attributing those to `--fix` means a developer auditing their overrides file may not realize the `>=` entries are the ones requiring attention.
    

**Fix**: Replace the paragraph with an accurate description: `pnpm audit --fix` writes caret ranges that already cap at the current major, so its output is generally safe as written. The entries that need manual upper-bound capping are any `>=` overrides added by hand — those have no implicit ceiling.
___
*[`docs/dependency-updates.md` line 80 at r1](https://reviewable.io/reviews/eten-tech-foundation/scripture-editors/510#-Oyd7LZ26pOH4to4L_eD:-Oyd7LZ26pOH4to4L_eE:bm3nx7o) ([raw file](https://github.com/eten-tech-foundation/scripture-editors/blob/4a324212d408ac9d902ca45bc5b69c00aab6f6a1/docs/dependency-updates.md#L80)):*
> ```Markdown
> - **Adding a second override keyed on the version `pnpm audit` printed usually does nothing.** If the
>   package already has an entry, that entry is the one in play. Raise its value; don't add a sibling.
> - **An uncapped value silently absorbs every future advisory in that package.** The locked vulnerable
> ```

NIT Claude Finding - Warning

The document's explanation of why uncapped overrides are dangerous contains a mechanism error that could mislead a developer's mental model.

The document says: "An uncapped value silently absorbs every future advisory in that package. The locked vulnerable version keeps satisfying `>=3.3.3` forever, so nothing re-resolves — the override looks like protection while the vulnerable copy sits underneath it."

This is the wrong mechanism. `pnpm audit` reads the resolved version from the lockfile and checks it against advisory databases — it does not consult override ranges. A new advisory against the pinned version (whatever it resolved to) would still appear in `pnpm audit` output. The override does not conceal it.

What actually happens: `pnpm install` won't automatically re-resolve the pinned package when a new advisory drops, because the locked version still satisfies the override range. The advisory is visible; the automatic protection is absent. The distinction matters: "your tooling hides this from you" leads to a very different operational posture than "your tooling shows this; nothing automatically fixes it." The real danger, which the document correctly identifies elsewhere, is that `pnpm audit` isn't run regularly (CI doesn't run it). That's the exposure, not concealment.

**Fix**: Replace "silently absorbs" with a description of what actually happens — the override doesn't re-resolve automatically when a new advisory drops, so future vulnerabilities in the locked version require a manual sweep to clear.
___
*[`docs/dependency-updates.md` line 110 at r1](https://reviewable.io/reviews/eten-tech-foundation/scripture-editors/510#-Oyd6Qor14uf-9SepF46:-Oyd6Qor14uf-9SepF47:b-wk3mqk) ([raw file](https://github.com/eten-tech-foundation/scripture-editors/blob/4a324212d408ac9d902ca45bc5b69c00aab6f6a1/docs/dependency-updates.md#L110)):*
> ```Markdown
> 
>    ```bash
>    gh api repos/{owner}/{repo}/dependabot/alerts --paginate \
> ```

NIT Claude Finding - Warning 

**Summary**: Step 3 gives a command to list dependabot alerts that will always fail for anyone who runs it.

The command is:

```bash
gh api repos/{owner}/{repo}/dependabot/alerts --paginate \
  -q '.[] | select(.state=="open") | ...'
```

The `{owner}` and `{repo}` are template placeholders, not variables. The GitHub API treats them as literal path segments and returns a 404. There's no instruction anywhere in the document to substitute them. This is the step that maps PRs to their underlying advisories — the key diagnostic step that makes the rest of the procedure possible. An agent or developer copy-pasting this command (which the skill explicitly instructs them to do) cannot proceed.

**Fix**: Either use the real repo coordinates (`eten-tech-foundation/scripture-editors`) or add an explicit substitution note immediately above the command.
___
*[`docs/dependency-updates.md` line 139 at r1](https://reviewable.io/reviews/eten-tech-foundation/scripture-editors/510#-Oyd6c2R-5dd5sL3oS5G:-Oyd6c2R-5dd5sL3oS5H:bil9xu8) ([raw file](https://github.com/eten-tech-foundation/scripture-editors/blob/4a324212d408ac9d902ca45bc5b69c00aab6f6a1/docs/dependency-updates.md#L139)):*
> ```Markdown
>    not close the advisory, however green its checks are.
> 6. **Edit the manifests by hand**, then `pnpm install`.
> 7. **Re-audit and compare.** Diff the advisory-ID set against your baseline. For every advisory **in
> ```

NIT  Claude Finding - Warning

### Step 7 instructs a diff that it never shows how to perform — and this is the only security gate

**Summary**: The document's only security verification step is described but never demonstrated.

Step 1 explicitly says: "Keep the JSON — the GHSA ID list is what you diff at step 7." Step 7 says: "Diff the advisory-ID set against your baseline." The document also says plainly: "The re-audit at step 7 is therefore the only gate that exists. Don't skip it."

That gate is the difference between thinking you closed a vulnerability and actually having closed it. The document provides no command to extract GHSA IDs from the JSON or perform the diff. `pnpm audit --json` produces a large JSON object; extracting and comparing the advisory ID sets is not obvious, especially since the document itself notes that raw counts are misleading ("30 → 28 is equally consistent with 'fixed the two targets,' 'fixed two unrelated things,' and 'fixed three and introduced one'").

A concrete command closes this gap completely:

```bash
diff <(jq -r '.advisories | keys[]' tmp/audit-before.json | sort) \
     <(jq -r '.advisories | keys[]' tmp/audit-after.json | sort)
```
___
*[`docs/dependency-updates.md` line 168 at r1](https://reviewable.io/reviews/eten-tech-foundation/scripture-editors/510#-Oyd6om4BVoJm6GvEnIm:-Oyd6om4BVoJm6GvEnIn:b-9tnq5z) ([raw file](https://github.com/eten-tech-foundation/scripture-editors/blob/4a324212d408ac9d902ca45bc5b69c00aab6f6a1/docs/dependency-updates.md#L168)):*
> ```Markdown
> 10. **Let dependabot close its own PRs.** Closing one by hand only asserts you handled it; letting
>     dependabot close it confirms the fix really is in the base branch. Most self-close within a
>     couple of minutes of `main` reaching or exceeding the version they propose.
> ```

NIT  Claude Finding (This one I was less sure about but thought I would run past you)

### Dependabot PRs do not reliably self-close "within a couple of minutes"

**Summary**: The timing claim is contradicted by official dependabot-core behavior and risks a confusing experience for anyone following the procedure.

The document says: "Most self-close within a couple of minutes of `main` reaching or exceeding the version they propose." Dependabot uses a polling/scheduled model, not an event-driven one. [dependabot-core issue #3880](https://github.com/dependabot/dependabot-core/issues/3880) is titled "Dependabot does no longer close unnecessary PRs automatically" — closure depends on when the next scheduled evaluation cycle runs, which has no guaranteed window.

More importantly: the document itself documents two scenarios where auto-close never happens (override-only fixes and PRs older than 30 days), then gives `@dependabot rebase` as the fix for both. Since overrides are described as "the usual fix here… the common case, not the exception," `@dependabot rebase` is effectively the standard resolution path. Presenting it as a fallback for edge cases is backwards.

**Fix**: Remove the "couple of minutes" timing claim. Promote `@dependabot rebase` as the general recommendation after your commit is on `main`, with auto-close noted as what happens in the best case.
___
*[`docs/dependency-updates.md` line 215 at r1](https://reviewable.io/reviews/eten-tech-foundation/scripture-editors/510#-Oyd781IE7BZnole4Jf4:-Oyd781IE7BZnole4Jf5:bt7oi6c) ([raw file](https://github.com/eten-tech-foundation/scripture-editors/blob/4a324212d408ac9d902ca45bc5b69c00aab6f6a1/docs/dependency-updates.md#L215)):*
> ```Markdown
>   entry below that only when a specific package has actually burned you on a minor.
> 
>   An uncapped value has already done damage here, not just risked it: `svgo@>=3.0.0 <3.3.3:
> ```

NIT  Claude Finding - Warning

### The document teaches a rule, demonstrates it with a live example, and then ships without fixing that example

**Summary**: The svgo override is uncapped, this PR's documentation calls it out as present-day evidence of damage from not capping, and the fix is deferred to a separate PR with auto-merge disabled.

The document says: "An uncapped value has already done damage here, not just risked it: `svgo@>=3.0.0 <3.3.3: ">=3.3.3"` walked svgo from 3.x to 4.0.1 under a plugin declaring `^3.0.2`. It went unnoticed because nothing in this repo executes svgo."

That override is still uncapped in `pnpm-workspace.yaml` after this PR merges. The PR body attributes the fix to #512, but #512 has auto-merge disabled. If #512 stalls — for any reason, including the merge order the PR body itself requests — `main` will permanently host documentation that:

1.  Teaches capping as a hard rule
2.  Cites svgo as a live example of what happens when you don't cap
3.  Leaves the named example uncapped

Capping svgo in this PR is a one-line change (`">=3.3.3"` → `">=3.3.3 <4"`) with zero functional impact since svgr is unreferenced in the Vite repo. The PR body already acknowledges this.

irahopkinson:
Thanks — these were worth running by me. Six of the seven led to a change. One I've refuted with a test, and one I've fixed differently to what was suggested, both explained below.

Rebased onto `main` (now including #522) and force-pushed to `f9723e5e`. The PR is documentation only now — the `form-data` cap moved out to #522.

---

### 2. `pnpm audit --fix` writes caret ranges, not `>=` — **refuted, no change**

You were right to doubt this one. I tested it rather than reasoning about it: ran `pnpm audit --fix` on this repo at pnpm 10.31.0 and read what it wrote.

Every generated value is a bare `>=X.Y.Z`. **Zero carets.**

```
'@babel/core@<=7.29.0': '>=7.29.6'
adm-zip@<0.6.0: '>=0.6.0'
body-parser@<1.20.6: '>=1.20.6'
undici@>=7.0.0 <7.29.0: '>=7.29.0'
nx@>=17.0.4 <22.7.2: '>=22.7.2'
```

`git diff | grep -c '\^'` → `0`.

The `caretRangeForPatched()` behaviour may exist in another version or another code path, but it isn't what this repo's pnpm does, so the doc's original claim stands. I've kept it and attached the evidence.

The run confirmed the second half too — `--fix` appends one entry per advisory rather than editing the existing one, and a single pass left **four** overlapping `brace-expansion` entries, three `fast-uri` and four `js-yaml`. One useful detail I've added: it quotes its own additions with single quotes and leaves hand-written entries in double quotes, so you can tell at a glance what it touched.

### 5. Step 7 never shows how to diff the ID set — **accepted, different command**

Agreed, and this was the worst gap given it's the only gate. Your command doesn't work here though, for two reasons:

1. `.advisories` is keyed by **pnpm's own numeric IDs** (`1119108`, `1119502`, …), not GHSA IDs — so `keys[]` gives you the wrong identifiers. The GHSA ID is in each entry's `github_advisory_id` field.
2. `jq` isn't installed in this repo's toolchain.

Node always is, so I wrote and verified an equivalent, in both step 1 (capture) and step 7 (compare):

```bash
node -e "var a=require('./tmp/audit-before.json').advisories; for (var k in a) console.log(a[k].github_advisory_id)" | sort -u > tmp/ids-before.txt
```

Verified against a real audit: 36 records → 34 unique GHSA IDs.

### 3. "silently absorbs" is the wrong mechanism — **accepted**

Correct, and the distinction matters exactly as you describe. `pnpm audit` reads resolved versions from the lockfile and never consults override ranges, so nothing is concealed. What's missing is the *automatic* fix. Rewritten to say the advisory is visible but nothing re-resolves it, with the "CI doesn't run audit" point as the real exposure.

### 6. Dependabot doesn't reliably self-close in minutes — **accepted, one thing kept**

Agreed on the substance: it polls rather than reacting to events, and `@dependabot rebase` is the standard path since overrides are the usual fix. Step 10 now leads with the rebase comment.

I've kept the two-minute observation, demoted to the lucky case rather than the expectation — #507 and #504 genuinely did close about two minutes after their fix merged, and deleting an accurate observation to make the guidance tidier felt wrong.

### 4. `{owner}/{repo}` placeholders 404 — **accepted**

Plainly right, and it's the one command the skill tells you to copy-paste. Now uses real coordinates.

### 1. rollup exact-version pin & 7. svgo left uncapped — **fixed in #522**

Both resolved. #522 capped every entry in the file and replaced the rollup pin with `">=4.59.0 <5"`.

On svgo, worth flagging that the suggested fix was the wrong direction: capping `">=3.3.3"` to `">=3.3.3 <4"` would have **downgraded** svgo from 4.0.1 to 3.3.x — retreating across a major to escape the advisory rather than patching forward. svgo 4.0.1 is flagged with patched `>=4.0.2`, so #522 raised the floor instead: `svgo@<4.0.2: ">=4.0.2 <5"`.

Investigating that also turned up three more overrides that had silently crossed a major the same way — `path-to-regexp` 0.1.x → 8.4.2, `picomatch` 2.x → 4.0.4, `brace-expansion` 1.x/2.x → 5.0.5. So the finding was more productive than the specific fix. All four are capped on `main` now, and the doc's example is reframed as history rather than a live defect.

---

### One thing that changed independently

The doc previously cited a 30-finding baseline. `main` now reports 39 against the same resolutions — the advisory database moved, not the code. Rather than just refresh the number I've used it as evidence: the doc now says a falling count is not proof you fixed anything and a steady one is not proof you broke nothing, which is a better argument for diffing the ID set than the one I had.

katherinejensen00 (commented):
Very cool! Thank you for explaining all of that! Good learning opportunity. Thank you also for making those changes, Ira. They look great!

> @katherinejensen00 reviewed 4 files and all commit messages, and made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/eten-tech-foundation/scripture-editors/510#-:-OzCXpHi3nt0HQ44tSsZ:bgmtunj)* status: all files reviewed, 7 unresolved discussions (waiting on irahopkinson).

Issue #516, opened 2026-07-29
Lite entry point / bundle diet: importing Editorial ships the whole package (212 KB gz)


henrique221:
Follow-up to the Fluent RTE PoC size analysis (fluent-web#375, PR fluent-web#376), measured with esbuild against `@eten-tech-foundation/platform-editor` 0.8.14 (react/react-dom/yjs externalized).

**Importing only `Editorial` costs the same as importing everything: ~212 KB gz (678 KB min).** The dist is a single ESM file assembled through barrel exports, so nothing tree-shakes (211.9 KB gz for `Editorial` only vs 212.1 KB for `export *`).

Per-package attribution of those 678 KB min:

| KB min | package | needed for USJ editing? |
|---:|---|---|
| 190 | platform-editor own code | mostly yes |
| 134 | lexical core | yes |
| **89** | @xmldom/xmldom | no in the browser — `DOMParser` is native; only the USX conversion path uses it |
| **59** | @lexical/table | not for plain scripture editing |
| **33** | @lexical/yjs | collab binding, bundled even when yjs is absent |
| ~46 | @lexical/extension + devtools-core + link | no |
| ~40 | misc (quill-delta, lodash.isequal/clonedeep, rich-text, selection, utils…) | partly |

Source-side, the same picture: ~27k LOC of editor source, ~24k reachable from `Editorial` because of `export *` barrels; the genuinely needed USJ-editing core (editor, USJ nodes, adaptors, annotation/usj plugins, views) is ~15k LOC.

### Proposal

- split entries: core `Editorial` vs `Marginal`/comments vs collab (yjs) vs tables;
- use native `DOMParser` in browser builds instead of @xmldom (or lazy-load the USX path);
- publish a code-split, side-effect-annotated dist so consumers' bundlers can drop what they don't import;
- (packaging nicety) ship the editor CSS in the package instead of "copy from the repo" (README instruction today).

Realistic outcome for a USJ-editing consumer: **~145–150 KB gz**, from 212. For calibration, a full ProseMirror stack is ~62 KB gz — closing part of that gap removes one of the main arguments against SharedEditor on weak devices.

henrique221:
[...]

Issue #513, opened 2026-07-29
Editing materializes stray spaces into the document text (worst in complex scripts)


henrique221:
Found while integrating the Editorial component into Fluent for the RTE PoC (eten-tech-foundation/fluent-web#375, PoC PR eten-tech-foundation/fluent-web#376). Editing operations materialize whitespace into the document text, and the corrupted text flows out through `getUsj()` / `onUsjChange` — i.e. into anything the host app saves.

**Environment:** `@eten-tech-foundation/platform-editor` 0.8.14, Lexical 0.43, Chrome (headless and headful), `Editorial` with `hasExternalUI: true`, USJ loaded via `defaultUsj`/`setUsj`.

### Repro A — typing while external annotations are present

1. Load a chapter USJ, apply a few `editorRef.setAnnotation(...)` typed marks.
2. Type 3 characters into a marked verse.
3. Serialized text grows by **+10** instead of +3 — spaces appear at mark boundaries.

### Repro B — a mark ending exactly at a text node's end

1. `setAnnotation` with a range whose end offset equals the text piece length (e.g. a whole-verse highlight).
2. One space is appended to that text piece in the serialized USJ, per such mark. The wrap appears to pull the adjacent presentation space node inside the TypedMark, where it then serializes as real text (also released as text on unwrap).

In Fluent we work around this by keeping marks one character short of piece ends, but end-of-verse character diagnostics then can't be highlighted precisely.

### Repro C — zero annotations, plain typing (worst case: complex scripts)

1. No marks at all. Load Gujarati text (e.g. IRV Genesis 2). Type 3 characters into verse 1.
2. The edited verse gains a trailing space (+4 total instead of +3) — and an **untouched verse in the same document gains +2**, with a space materializing inside a word at a text-node boundary: `કર્યું` → `કર્ય ું`.
3. Same test with English text: only trailing spaces on the edited/previously-marked verses (which a consumer can trim); no internal corruption.

Repro C is the serious one: internal spaces inside untouched complex-script verses cannot be repaired downstream, and complex-script languages are exactly Fluent's target content.

### Mitigations we currently use in the PoC (fluent-web `src/features/rte`)

- strip all marks on the first editing keystroke, re-apply when idle;
- clamp annotation ranges one char short of text-piece ends;
- trim verse texts when deriving save payloads.

These reduce but cannot eliminate the corruption (Repro C survives). Happy to pair on this — we have deterministic repros and the fluent-web integration to test against.

[...]

PR #471, opened 2026-04-20
platform: no trailing para space

irahopkinson:
- don't automatically add space at the end of a para-like node
- move `$isParaLikeNode` to `node.utils.ts`

---

<a href="https://app.devin.ai/review/eten-tech-foundation/scripture-editors/pull/471" target="_blank">
  <picture>
    <source media="(prefers-color-scheme: dark)" srcset="https://static.devin.ai/assets/gh-open-in-devin-review-dark.svg?v=1">
    Open in Devin Review
  </picture>
</a>

- - -
This change is [Reviewable](https://reviewable.io/reviews/eten-tech-foundation/scripture-editors/471)

PR #517, opened 2026-07-29
fix(shared-react): don't insert a space between adjacent text nodes of the same run

henrique221:
Fixes the complex-script corruption case of #513.

## Root cause

`$textNodeTrailingSpaceTransform` treats any `TextNode` as a candidate for a structural trailing space unless its next sibling is a note/char/annotation. But when the **next sibling is another `TextNode`**, the two nodes are one logical text run — IME composition (complex scripts) and annotation-wrap splits leave runs as multiple adjacent nodes that Lexical won't merge (e.g. a `segmented` composition node). The transform then inserts a space **inside the word**, and the space serializes into USJ: typing in Gujarati turned `કર્યું` into `કર્ય ું`, including in verses the user never touched.

## Fix

Skip the trailing-space transform when the next sibling is a `TextNode`. This also protects a space-only node that sits between two text nodes from the placeholder cleanup — in that position it is real content, not scaffolding. The canonical pre-verse space is unaffected: once a composition segment settles into a normal node (or nodes merge), this same transform adds it as before.

## Verification

- New regression test (`should not add a space between adjacent text nodes of the same run`) fails on `main` with exactly the observed signature (`'ab cd'`) and passes with the fix.
- Full suites green: shared 107, shared-react 1270 (incl. the new test), platform-editor 124.
- End-to-end against the reporting app: built this branch, pointed fluent-web's RTE PoC (eten-tech-foundation/fluent-web#376) at the tarball, and re-ran the failing scenario — typing `xyz` into Gujarati Genesis 2 now yields exactly the typed characters plus the canonical structural spaces (mid-paragraph verses only, none at para end per #471), with **no internal corruption in any verse**. On 0.8.14 the same probe produced +7 stray characters across touched and untouched verses.

The remaining #513 cases (trailing space at para end, annotation-boundary spaces) were already addressed on `main` after 0.8.14 by #471/#496/#497 — with this change, all three #513 repros are covered. A release would get the whole set to consumers.

One unrelated packaging note found while testing the tarball: `exports["."].development` now points at `./src/index.ts`, which makes consumers' Vite dev servers resolve raw monorepo sources (bare `shared` imports) and crash — noted on #516.

- - -
This change is [Reviewable](https://reviewable.io/reviews/eten-tech-foundation/scripture-editors/517)

## Demo video

https://github.com/user-attachments/assets/41675112-0501-49d7-97db-db2977f240eb


kaseywright (approved):

kaseywright (commented):
> @kaseywright reviewed 2 files and all commit messages.
> *[Reviewable](https://reviewable.io/reviews/eten-tech-foundation/scripture-editors/517)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved.

irahopkinson (approved):
:lgtm: good find and thanks for this fix.

> @irahopkinson reviewed 2 files and all commit messages, and made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/eten-tech-foundation/scripture-editors/517#-:-OzP6k2k54V3QUaxK1hs:bhekwzt)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved.

PR #524, opened 2026-08-05
PT-4125: Regression test for native (Power-mode) verse delete undo

katherinejensen00:
## What

Adds a regression test covering **native (Power-mode) verse deletion + undo** in the platform editor, and hardens both verse-delete undo tests.

Fixes [PT-4125](https://paratextstudio.atlassian.net/browse/PT-4125).

## Why

PT-4125 reported that deleting verses in **PT10 Power mode** couldn't be undone (Undo / CTRL+Z did nothing). Power mode maps to `structureProtectionMode: "off"`, where `StructureKeyboardPlugin` registers nothing and deletion is fully **native Lexical** — a different path from the guarded two-step delete covered by the existing PT-4102 test.

Investigation found the reported bug is **already fixed** on `main` by PT-4102 (#505): deferring `ScriptureReferencePlugin`'s `SELECTION_CHANGE` dispatch off the verse mutation listener with `queueMicrotask`. That fix is mutation-listener based and therefore **gesture-agnostic**, so it covers the native path too. The original report was against Pkg 217, which predates the fix.

This PR closes the **test-coverage gap**: the native path had no regression test, even though it's the exact path the ticket reproduces.

## Changes

- New test: deletes a range spanning two adjacent verse markers (the Mark 1:7–8 shape) via `KEY_DELETE_COMMAND` in `"off"` mode, and asserts a single Undo restores the document.
- Extracted a shared `mountEditorForUndo()` helper and a single parameterized `pressDeleteKey()`, removing ~18 lines of duplicated mount boilerplate across both undo tests.
- Strengthened **both** undo tests (this and the PT-4102 sibling) to assert exact restoration — `getUsj()` `toEqual` a pre-delete snapshot — instead of loose substring checks, so a reordered / duplicated / partial undo now fails.

## Verification

- `nx test @eten-tech-foundation/platform-editor` — full suite green; lint + typecheck clean.
- **Protective:** both tests were confirmed to go **red** when the `queueMicrotask` fix is reverted, failing with the exact PT-4125 symptom (undo restores nothing).
- **Not brittle:** the `toEqual` snapshot assertion was verified against the actual USJ round-trip (stable) — it passes on fixed code and fails only on real corruption.

## Note

The collapsed-caret Backspace gesture can't be exercised in jsdom (it routes through Lexical's `deleteCharacter`, which needs `domSelection.modify`, unimplemented in jsdom). That gesture was **verified manually in-app**; the range-delete test drives the same verse-destruction → mutation-listener → history path.

- - -
This change is [Reviewable](https://reviewable.io/reviews/eten-tech-foundation/scripture-editors/524)

### DIFF
--- a/packages/platform/src/editor/Editor.test.tsx
+++ b/packages/platform/src/editor/Editor.test.tsx
@@ -2,13 +2,14 @@
 // eslint-disable-next-line @nx/enforce-module-boundaries
 import { usjGen1v1 } from "../../../utilities/src/converters/usj/converter-test.data";
 import Editor from "./Editor";
-import { EditorRef } from "./editor.model";
+import { EditorOptions, EditorRef } from "./editor.model";
 import Editorial from "../Editorial";
 import { flushQueuedEvents } from "./editor-test.utils";
 import { ContentJsonPath, Usj } from "@eten-tech-foundation/scripture-utilities";
 import { useLexicalComposerContext } from "@lexical/react/LexicalComposerContext";
+import { SerializedVerseRef } from "@sillsdev/scripture";
 import { act, render } from "@testing-library/react";
-import { KEY_DOWN_COMMAND, LexicalEditor } from "lexical";
+import { KEY_DELETE_COMMAND, KEY_DOWN_COMMAND, LexicalCommand, LexicalEditor } from "lexical";
 import { createRef, RefObject, useEffect } from "react";
 import { vi } from "vitest";
 
@@ -191,10 +192,44 @@ function GrabEditor({ onEditor }: { onEditor: (editor: LexicalEditor) => void })
   return null;
 }
 
-async function pressDelete(editor: LexicalEditor): Promise<void> {
+/** Renders the platform <Editor> and returns both the public ref and the underlying Lexical editor
+ * (needed to dispatch key commands). Covers the setup every delete/undo test repeats. */
+async function mountEditorForUndo(config: {
+  usj: Usj;
+  scrRef: SerializedVerseRef;
+  structureProtectionMode: EditorOptions["structureProtectionMode"];
+}): Promise<{ editorRef: EditorRef; lexicalEditor: LexicalEditor }> {
+  const ref = createRef<EditorRef>();
+  let editor: LexicalEditor | undefined;
+  await act(async () => {
+    render(
+      <Editor
+        ref={ref}
+        defaultUsj={config.usj}
+        scrRef={config.scrRef}
+        onScrRefChange={vi.fn()}
+        options={{ structureProtectionMode: config.structureProtectionMode }}
+      >
+        <GrabEditor onEditor={(e) => (editor = e)} />
+      </Editor>,
+    );
+  });
+  await flushQueuedEvents();
+  if (!ref.current || !editor) throw new Error("EditorRef did not mount");
+  return { editorRef: ref.current, lexicalEditor: editor };
+}
+
+/** Dispatches a synthetic Delete keydown as `command`. Guarded mode listens on KEY_DOWN_COMMAND
+ * (StructureKeyboardPlugin drives the two-step delete from there); Power mode ("off") has no such
+ * listener, so its native delete must go straight to KEY_DELETE_COMMAND — jsdom can't carry a real
+ * DOM keydown far enough to reach RichTextPlugin on its own. */
+async function pressDeleteKey(
+  editor: LexicalEditor,
+  command: LexicalCommand<KeyboardEvent>,
+): Promise<void> {
   await act(async () => {
     editor.dispatchCommand(
-      KEY_DOWN_COMMAND,
+      command,
       new KeyboardEvent("keydown", { key: "Delete", bubbles: true, cancelable: true }),
     );
   });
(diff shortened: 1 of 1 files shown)


mattgetgen (changes requested):
Great work! Claude thought this was a great PR, but it had a few notes on correctness and cleanup. It made two other notes about the added test, and I'll add it below, but this looks great!

- The new test dispatches KEY_DELETE_COMMAND directly, and the comment justifying that ("jsdom can't carry a real DOM keydown far enough") is empirically false — a verifier drove the same scenario with a real keydown and it worked. The PT-4102 sibling does route through KEY_DOWN_COMMAND, so only the native path has this gap.
- The fixture has no \v 6, but the test mounts at MRK 1:6, so the caret resolves to verse 0 — the "delete the verse you're referencing" condition the ticket describes isn't actually reached.

> @mattgetgen made 8 comments.
> *[Reviewable](https://reviewable.io/reviews/eten-tech-foundation/scripture-editors/524#-:-OzT771Q4KURqYWxOs4U:b-onc8ol)* status: 0 of 1 files reviewed, 7 unresolved discussions (waiting on katherinejensen00).

___
*[`packages/platform/src/editor/Editor.test.tsx` line 197 at r1](https://reviewable.io/reviews/eten-tech-foundation/scripture-editors/524#-OzT6oXZ9OMgGzeQzVHM:-OzT6oXZ9OMgGzeQzVHN:b-oqx110) ([raw file](https://github.com/eten-tech-foundation/scripture-editors/blob/5c31d0e237628edfe85a9665da974068f6ed9dd9/packages/platform/src/editor/Editor.test.tsx#L197)):*
> ```XML
> /** Renders the platform <Editor> and returns both the public ref and the underlying Lexical editor
>  * (needed to dispatch key commands). Covers the setup every delete/undo test repeats. */
> async function mountEditorForUndo(config: {
> ```

Cleanup: mountEditorForUndo duplicates the existing createEditorRefForTesting helper in the same file, with a different return convention.
___
*[`packages/platform/src/editor/Editor.test.tsx` line 218 at r1](https://reviewable.io/reviews/eten-tech-foundation/scripture-editors/524#-OzT6kEi0z-ah_ElpxZI:-OzT6kEjCXUa6HrvD3L0:bsko9fb) ([raw file](https://github.com/eten-tech-foundation/scripture-editors/blob/5c31d0e237628edfe85a9665da974068f6ed9dd9/packages/platform/src/editor/Editor.test.tsx#L218)):*
> ```XML
>   });
>   await flushQueuedEvents();
>   if (!ref.current || !editor) throw new Error("EditorRef did not mount");
> ```

Cleanup: The guard throws "EditorRef did not mount" even when the failure is the missing Lexical editor, misreporting the cause.
___
*[`packages/platform/src/editor/Editor.test.tsx` line 219 at r1](https://reviewable.io/reviews/eten-tech-foundation/scripture-editors/524#-OzT6eeG3OFy_Rpbb5t5:-OzT6eeG3OFy_Rpbb5t6:b9w82dk) ([raw file](https://github.com/eten-tech-foundation/scripture-editors/blob/5c31d0e237628edfe85a9665da974068f6ed9dd9/packages/platform/src/editor/Editor.test.tsx#L219)):*
> ```XML
>   await flushQueuedEvents();
>   if (!ref.current || !editor) throw new Error("EditorRef did not mount");
>   return { editorRef: ref.current, lexicalEditor: editor };
> ```

Correctness: `mountEditorForUndo` returns `ref.current` captured once at mount, but `Editor`'s `useImperativeHandle` has no dependency array, so React installs a fresh handle on every render and the helper hands back a stale one.
___
*[`packages/platform/src/editor/Editor.test.tsx` line 333 at r1](https://reviewable.io/reviews/eten-tech-foundation/scripture-editors/524#-OzT6TZg0-SvWON_1Sen:-OzT6TZg0-SvWON_1Seo:b5ijzip) ([raw file](https://github.com/eten-tech-foundation/scripture-editors/blob/5c31d0e237628edfe85a9665da974068f6ed9dd9/packages/platform/src/editor/Editor.test.tsx#L333)):*
> ```XML
>     });
>     // Snapshot the loaded document; a single undo must return to exactly this.
>     const original = editorRef.getUsj();
> ```

Correctness: `original = editorRef.getUsj()` is the same object reference as the module-level fixture, not an independent snapshot of the loaded document.
___
*[`packages/platform/src/editor/Editor.test.tsx` line 348 at r1](https://reviewable.io/reviews/eten-tech-foundation/scripture-editors/524#-OzT68k12dbeAewxB-aC:-OzT68k12dbeAewxB-aD:b-b5fcme) ([raw file](https://github.com/eten-tech-foundation/scripture-editors/blob/5c31d0e237628edfe85a9665da974068f6ed9dd9/packages/platform/src/editor/Editor.test.tsx#L348)):*
> ```XML
> 
>     // Precondition: both verse markers were actually deleted.
>     const afterDelete = JSON.stringify(editorRef.getUsj());
> ```

Correctness: The new test's post-delete precondition checks only that verse markers 7 and 8 vanished, dropping the text-removal check its PT-4102 sibling makes.
___
*[`packages/platform/src/editor/Editor.test.tsx` line 344 at r1](https://reviewable.io/reviews/eten-tech-foundation/scripture-editors/524#-OzT5qNX3Jewiara9wTq:-OzT5qNX3Jewiara9wTr:b8skvwu) ([raw file](https://github.com/eten-tech-foundation/scripture-editors/blob/5c31d0e237628edfe85a9665da974068f6ed9dd9/packages/platform/src/editor/Editor.test.tsx#L344)):*
> ```XML
>     });
> 
>     await pressDeleteKey(lexicalEditor, KEY_DELETE_COMMAND);
> ```

Correctness: The new native-delete test bypasses the KEY_DOWN_COMMAND chain that the guarded test exercises, on a stated jsdom justification that is factually wrong.
___
*[`packages/platform/src/editor/Editor.test.tsx` line 329 at r1](https://reviewable.io/reviews/eten-tech-foundation/scripture-editors/524#-OzT6M9DCzFvvfaYMMRc:-OzT6M9DCzFvvfaYMMRd:b57zwy3) ([raw file](https://github.com/eten-tech-foundation/scripture-editors/blob/5c31d0e237628edfe85a9665da974068f6ed9dd9/packages/platform/src/editor/Editor.test.tsx#L329)):*
> ```XML
>     const { editorRef, lexicalEditor } = await mountEditorForUndo({
>       usj: usjWithTwoAdjacentVerses,
>       scrRef: { book: "MRK", chapterNum: 1, verseNum: 6 },
> ```

Correctness: The PT-4125 fixture has no verse marker before "Six ", so `scrRef` verseNum 6 names a verse absent from the document and the post-delete caret resolves to verse 0 instead of a real verse — the "delete the verse you're referencing" state the bug report describes is never reached.

mattgetgen:
Apologies for all of the LGTMs, I'm not used to reviewable I guess

PR #525, opened 2026-08-06
PT-3203: resolve cursor disappearing issue when deleting text

mattgetgen:
This change is [Reviewable](https://reviewable.io/reviews/eten-tech-foundation/scripture-editors/525)

### DIFF
--- a/libs/shared-react/src/plugins/usj/TextSpacingPlugin.test.tsx
+++ b/libs/shared-react/src/plugins/usj/TextSpacingPlugin.test.tsx
@@ -3,6 +3,7 @@
 import { $expectSelectionToBe } from "../../../../../libs/shared/src/nodes/usj/test.utils";
 import { ImmutableVerseNode, $createImmutableVerseNode } from "../../nodes/usj/ImmutableVerseNode";
 import { $isSomeVerseNode } from "../../nodes/usj/node-react.utils";
+import { $getRangeFromUsjSelection } from "./annotation/selection.utils";
 import { TextSpacingPlugin } from "./TextSpacingPlugin";
 import {
   baseTestEnvironment,
@@ -12,14 +13,7 @@ import {
   typeTextAtSelection,
 } from "./react-test.utils";
 import { act } from "@testing-library/react";
-import {
-  $createTextNode,
-  $getRoot,
-  $getSelection,
-  $isTextNode,
-  TextNode,
-  $setSelection,
-} from "lexical";
+import { $createTextNode, $getRoot, $isTextNode, TextNode, $setSelection } from "lexical";
 import {
   $createCharNode,
   $createImmutableChapterNode,
@@ -106,11 +100,11 @@ describe("TextSpacingPlugin", () => {
     });
   });
 
-  it("should remove the structural space left after deleting the only text of an empty verse", async () => {
-    // The structural trailing space is scaffolding for real text. Once that text is gone the verse
-    // is empty again, and core's round trip to disk (ParatextData) drops the space — so keeping it
-    // makes the editor's USJ disagree with what core reads back. Core reacts to that disagreement
-    // by reloading the whole editor, which destroys the caret.
+  it("should remove an empty verse's space that an edit left behind", async () => {
+    // Covers the space arriving via a user EDIT (type then delete); the sibling test below covers
+    // it arriving with already-loaded content. See `$textNodeTrailingSpaceTransform` for why the
+    // space must go. Unit-level guard only — the cross-process reload this prevents lives in core
+    // and is not exercised here.
     const { editor } = await testEnvironment();
 
     // Type between the two empty verses; the transform adds the structural trailing space.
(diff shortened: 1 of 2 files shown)


katherinejensen00 (changes requested):
Thanks for your great work on this, Matt! This will be a helpful improvement. By the way, Claude really likes your fix ;)

"This is a clean, well-scoped, well-reasoned fix — the empty-verse space removal is correct, the transform pair provably converges (no oscillation), the loop protection terminates, and coverage for the still-preserved structural-space cases is intact. My notes are all about test strength, one comment accuracy fix, and one architectural question — no correctness bugs."

> @katherinejensen00 reviewed 2 files and all commit messages, and made 9 comments.
> *[Reviewable](https://reviewable.io/reviews/eten-tech-foundation/scripture-editors/525#-:-OzN7P9UEv4t_aeu3WM7:b-rvi9yr)* status: all files reviewed, 8 unresolved discussions (waiting on mattgetgen).

___
*[`libs/shared-react/src/plugins/usj/TextSpacingPlugin.tsx` line 87 at r1](https://reviewable.io/reviews/eten-tech-foundation/scripture-editors/525#-OzNB-rf3ymnYmdRIYeB:-OzNB-rf3ymnYmdRIYeC:b-2aqlom) ([raw file](https://github.com/eten-tech-foundation/scripture-editors/blob/6dad01e864dadf8ead616dda2107967eab21b7ac/libs/shared-react/src/plugins/usj/TextSpacingPlugin.tsx#L87)):*
**2 — This shared transform now encodes a ParatextData-specific assumption; confirm that's intended, and consider a follow-up for the real root cause.** — Severity: **Medium (design question)** 

> `TextSpacingPlugin` lives in `libs/shared-react` and is also consumed by `packages/scribe` and the demos, but the new rationale is framed as core/ParatextData behavior ("core's round trip to disk drops that space"). Is "an empty verse carries no space" _canonical USJ_ (in which case this rightly belongs in shared code) or a _ParatextData serialization quirk_ (in which case a non-Paratext consumer would lose a space it wants to keep)? Also: the underlying trigger — an editor↔core USJ mismatch forcing a full editor reload that destroys the caret — is untouched, so the next benign divergence will reproduce the symptom. Worth a follow-up ticket for a caret-preserving reconcile, and a note in the PR description so a reviewer sees it.
___
*[`libs/shared-react/src/plugins/usj/TextSpacingPlugin.tsx` line 95 at r1](https://reviewable.io/reviews/eten-tech-foundation/scripture-editors/525#-OzNBYZM27qJqwDQWsuI:-OzNBYZM27qJqwDQWsuJ:bqa2hv8) ([raw file](https://github.com/eten-tech-foundation/scripture-editors/blob/6dad01e864dadf8ead616dda2107967eab21b7ac/libs/shared-react/src/plugins/usj/TextSpacingPlugin.tsx#L95)):*
**7 — Document the scope of `isEmptyVerseContent` (NBSP / multi-space / annotation-wrapped).** — Severity: **Low**

> The check matches only exact `""`/`" "`. NBSP is reasonably out of scope (it's meaningful content, and `$addTrailingSpace` treats it as already-spaced), but an empty verse wrapped in a `TypedMarkNode` won't be detected because this reads `getPreviousSibling()` directly, whereas the sibling `$verseNodeTransform` resolves _through_ wrappers (line 135). A one-line comment stating the intended scope (or a follow-up for the wrapped case) would prevent a future reader assuming full coverage.
___
*[`libs/shared-react/src/plugins/usj/TextSpacingPlugin.tsx` line 99 at r1](https://reviewable.io/reviews/eten-tech-foundation/scripture-editors/525#-OzNBd0-4cPQIulUVw8K:-OzNBd0-4cPQIulUVw8L:b-m3qo4x) ([raw file](https://github.com/eten-tech-foundation/scripture-editors/blob/6dad01e864dadf8ead616dda2107967eab21b7ac/libs/shared-react/src/plugins/usj/TextSpacingPlugin.tsx#L99)):*
**8 — Confirm annotation index resolution across an empty verse (PT-3835 neighbor).** — Severity: **Low (verify)**

> Removing the space reduces the paragraph's logical content-item count by one, which feeds annotation/comment offset resolution — the same index-shift class PT-3835 guarded against in the sibling transform. I believe this is _correct_ here (it aligns the editor to core's 2-item shape, which is what annotations anchor to), but given that history it's worth a targeted test: a comment/annotation anchored after an empty verse should still resolve to the right run once the space is gone.
___
*[`libs/shared-react/src/plugins/usj/TextSpacingPlugin.test.tsx` line 109 at r1](https://reviewable.io/reviews/eten-tech-foundation/scripture-editors/525#-OzNBR5I7SH-7FB1k6Qd:-OzNBR5I7SH-7FB1k6Qe:b-uy2a77) ([raw file](https://github.com/eten-tech-foundation/scripture-editors/blob/6dad01e864dadf8ead616dda2107967eab21b7ac/libs/shared-react/src/plugins/usj/TextSpacingPlugin.test.tsx#L109)):*
**6 — Test A and Test B largely cover the same behavior.** — Severity: **Low** 

> Both drive the transform with `[v1," ",v2]` and assert it collapses to `[v1,v2]`; the only real difference is the arrival path (real deletion vs direct construction). Keeping A (the realistic user path) is clearly worth it — consider whether B adds enough beyond it, or differentiate its intent in the name/comment. Also, both comments describe the cross-process reload that these unit tests don't exercise; a one-line "unit-level guard only" note would set expectations.
___
*[`libs/shared-react/src/plugins/usj/TextSpacingPlugin.test.tsx` line 110 at r1](https://reviewable.io/reviews/eten-tech-foundation/scripture-editors/525#-OzNBC4zEyQVr3gyQaQY:-OzNBC4zEyQVr3gyQaQZ:b-tsd9hi) ([raw file](https://github.com/eten-tech-foundation/scripture-editors/blob/6dad01e864dadf8ead616dda2107967eab21b7ac/libs/shared-react/src/plugins/usj/TextSpacingPlugin.test.tsx#L110)):*
**4 — Consider stating the "why" once instead of ~4 times.** — Severity: **Low**

> The core round-trip/reload rationale is repeated near-verbatim in the TSDoc (:61-63), the inline comment (:87-92), and both test bodies (:110-113, :309-316). Keeping the full explanation once (the production comment) and having the tests point to it avoids four copies drifting out of sync. (Minor — self-documenting tests are partly a virtue.)
___
*[`libs/shared-react/src/plugins/usj/TextSpacingPlugin.test.tsx` line 141 at r1](https://reviewable.io/reviews/eten-tech-foundation/scripture-editors/525#-OzN7jGf-LSxcsbxc4x0:-OzN7jGf-LSxcsbxc4x1:b9igfmu) ([raw file](https://github.com/eten-tech-foundation/scripture-editors/blob/6dad01e864dadf8ead616dda2107967eab21b7ac/libs/shared-react/src/plugins/usj/TextSpacingPlugin.test.tsx#L141)):*
**1 — Strengthen the caret assertion; it currently passes on the buggy code too.** — Severity: **Low–Medium**

> `expect($getSelection()).not.toBe(null)` doesn't actually guard the caret regression. I traced the pre-fix code: after deleting `"a"`, the `" "` node's next sibling is verse 2, so the old removal branch doesn't fire, the node survives, and the selection stays non-null — the assertion passes on broken code. The real fix is caught only by the `$getLogicalContentItems(...).toHaveLength(2)` / `getTextContent()===""` lines. Since this file already uses `$expectSelectionToBe(node, offset)` for exactly this (test:74, :86), please swap in `$expectSelectionToBe(...)` to pin the caret's expected resting node+offset.
___
*[`libs/shared-react/src/plugins/usj/TextSpacingPlugin.test.tsx` line 339 at r1](https://reviewable.io/reviews/eten-tech-foundation/scripture-editors/525#-OzNBKVL14dP0L2Q0F_j:-OzNBKVL14dP0L2Q0F_k:bvj0ggu) ([raw file](https://github.com/eten-tech-foundation/scripture-editors/blob/6dad01e864dadf8ead616dda2107967eab21b7ac/libs/shared-react/src/plugins/usj/TextSpacingPlugin.test.tsx#L339)):*
**5 — Align the two tests' assertion style.** — Severity: **Low** 

> Test B asserts `para.getChildren().toHaveLength(2)`, which depends on Lexical GC'ing the emptied node; test A uses the GC-robust `$getLogicalContentItems(...).toHaveLength(2)`. Using `$getLogicalContentItems` in both makes B invariant to reconciliation internals (it also already asserts `getTextContent()===""`, so this is purely consistency).
___
*[`libs/shared-react/src/plugins/usj/TextSpacingPlugin.tsx` line 100 at r1](https://reviewable.io/reviews/eten-tech-foundation/scripture-editors/525#-OzNB4_P0m8gxGQfsqYy:-OzNB4_P0m8gxGQfsqYz:b-yr3rgt) ([raw file](https://github.com/eten-tech-foundation/scripture-editors/blob/6dad01e864dadf8ead616dda2107967eab21b7ac/libs/shared-react/src/plugins/usj/TextSpacingPlugin.tsx#L100)):*
> ```XML
>   // entire content.
>   if ((text === " " && !$isSomeVerseNode(nextSibling)) || isEmptyVerseContent) {
>     // Only write when there is something to clear: re-setting the same text marks the node dirty
> ```

**3 — Fix the loop-protection comment: it credits the wrong clause.** — Severity: **Low**

The comment says re-setting the text "would loop against `$addTrailingSpace` below," but the `if (text !== "")` guard actually prevents a _self-loop_ (`setTextContent("")` → `getWritable()` marks the node dirty even when unchanged → the transform re-runs). It's the separate `text === ""` clause in `isEmptyVerseContent` (line 95) that stops the `$addTrailingSpace` re-add loop. These two protections are coupled — please note that both are load-bearing so a future cleanup doesn't remove one and reintroduce an infinite transform cycle (which Lexical turns into a crash).

katherinejensen00 (dismissed):
> @katherinejensen00 reviewed 2 files and all commit messages, made 1 comment, and resolved 7 discussions.
> *[Reviewable](https://reviewable.io/reviews/eten-tech-foundation/scripture-editors/525)* status: all files reviewed, 1 unresolved discussion (waiting on mattgetgen).

___
*[`libs/shared-react/src/plugins/usj/TextSpacingPlugin.tsx` line 87 at r1](https://reviewable.io/reviews/eten-tech-foundation/scripture-editors/525#-OzNB-rf3ymnYmdRIYeB:-OzNmVf35IAJdlJQwudr:ba8fqhz) ([raw file](https://github.com/eten-tech-foundation/scripture-editors/blob/6dad01e864dadf8ead616dda2107967eab21b7ac/libs/shared-react/src/plugins/usj/TextSpacingPlugin.tsx#L87)):*

**2 — This shared transform now encodes a ParatextData-specific assumption; confirm that's intended, and consider a follow-up for the real root cause.** — Severity: **Medium (design question)** 

> `TextSpacingPlugin` lives in `libs/shared-react` and is also consumed by `packages/scribe` and the demos, but the new rationale is framed as core/ParatextData behavior ("core's round trip to disk drops that space"). Is "an empty verse carries no space" _canonical USJ_ (in which case this rightly belongs in shared code) or a _ParatextData serialization quirk_ (in which case a non-Paratext consumer would lose a space it wants to keep)? Also: the underlying trigger — an editor↔core USJ mismatch forcing a full editor reload that destroys the caret — is untouched, so the next benign divergence will reproduce the symptom. Worth a follow-up ticket for a caret-preserving reconcile, and a note in the PR description so a reviewer sees it.
</blockquote></details>

This is more of a design question rather than something that needs to be solved now. Will go ahead and approve this PR since everything else is fixed. Thanks for the great work, Matt!

PR #526, opened 2026-08-06
Fix chapter-number reinstatement: c-marker double-increment and applyUpdate undo loss

captaincrazybro:
## Summary

Two fixes supporting Platform.Bible's chapter-number reinstatement feature (paranext-core PR #2645), which lets a user re-add a missing `\c` marker to an effectively-blank chapter via `EditorRef.applyUpdate('local')`.

- **`\c` marker action double-increment**: the marker-action handler for inserting a chapter marker unconditionally incremented the chapter number, even when no `\c` node existed yet in the current chapter (the reinstatement case). Now it only increments when a chapter node is already present (`$findChapter(...) !== undefined`), matching the existing insert-a-chapter test's behavior while adding a regression test for the reinstatement path.
- **`applyUpdate('local')` inserts silently lost undo history**: `Editor.tsx`'s `viewOptions` was recomputed by reference on every render (`view ?? defaultViewOptions`), so `LoadStatePlugin`'s reload effect — which unconditionally dispatches `CLEAR_HISTORY_COMMAND` whenever any of its dependencies change by reference — fired on effectively every render, wiping Lexical's undo stack shortly after an `applyUpdate('local')` insert landed. Fixed by memoizing `viewOptions` against its own previous value via deep-equality (`Editor.tsx`), so the reload effect only re-fires when `view` is genuinely different. The same reload effect also depends on `logger`, which was passed through unmemoized — stabilized it the same way. Added a regression test (stable `view` across renders preserves history), a reverse-direction test (a genuinely different `view` object still triggers reload/history-clear), and an integration test that exercises the real `applyUpdate` → `onUsjChange` → parent-re-render round trip with a genuinely unmemoized `view` (verified it fails without the fix), plus a rationale comment for why `nodeOptions`/`contextMenuOptions` are intentionally left reference-only (not consumed by the reload effect).

## Known follow-up (not fixed here)

- A second, separate cause of undo-history loss was identified but not chased down in this branch: `paranext-core`'s `use-editor-pdp-sync.hook.ts` calls `EditorRef.setUsj()` on the PDP-echo success path after certain inserts, which also clears Lexical's undo history. This is tracked in the paranext-core PR and documented as a deliberate, decided-upon follow-up rather than a merge blocker for either PR.
- **This fix stabilizes `viewOptions`/`logger` at the `platform` package's `Editor.tsx` only — it does not touch the root cause.** `LoadStatePlugin`'s reload effect unconditionally clears undo history on every fire, with no check that the resulting state actually differs from what's loaded; per-consumer stabilization treats the symptom. `packages/scribe/src/editor/Editor.tsx` and `NoteEditor.tsx` also consume `LoadStatePlugin` and pass `viewOptions` through unmemoized — the identical exposure, in a package this PR deliberately doesn't touch. A real fix inside `LoadStatePlugin` itself (gate `CLEAR_HISTORY_COMMAND` on an actual document/state change) would close this for every consumer, including scribe, at once — left as a follow-up given its scope (a shared plugin used across packages) rather than expanding this PR.

## Test plan

- [x] `npx nx run @eten-tech-foundation/platform-editor:lint` — clean
- [x] `npx nx run @eten-tech-foundation/platform-editor:typecheck` — clean
- [x] `npx vitest run packages/platform/src/editor` — 155 passed, 3 skipped (pre-existing skips, unrelated)
- [x] Manually verified in a running Platform.Bible instance (via the paranext-core branch's yalc-vendored build) that reinstating a chapter number no longer double-inserts `\c`, and that a single `applyUpdate('local')` insert is now undoable immediately after insert.

AI-assisted — session generated by Claude Code (Claude Sonnet 5).

- - -
This change is [Reviewable](https://reviewable.io/reviews/eten-tech-foundation/scripture-editors/526)

### DIFF
--- a/packages/platform/src/editor/Editor.test.tsx
+++ b/packages/platform/src/editor/Editor.test.tsx
@@ -283,15 +283,10 @@ const blankUsj: Usj = {
 };
 
 describe("applyUpdate('local') undo-history retention", () => {
-  // A chapter-scaffold insert via `EditorRef.applyUpdate(ops, 'local')` must participate in the
-  // undo stack like any other local edit. It previously did not: `LoadStatePlugin`'s reload effect
-  // depends on `viewOptions`, which `Editor` derived with a plain `useMemo(() => view ??
-  // defaultViewOptions, [view])` - a reference-only memo. A parent re-render that passes a
-  // fresh-but-value-equal `options.view` object (e.g. one triggered by `applyUpdate`'s own
-  // `onUsjChange` round-trip completing, moments after the insert) changed `viewOptions`'s
-  // identity without changing its value, re-firing `LoadStatePlugin` and its unconditional
-  // `CLEAR_HISTORY_COMMAND` dispatch - wiping the just-created undo entry with no document or view
-  // change to justify it.
+  // This test verifies that a value-equal (but reference-different) `view` object does NOT clear
+  // undo history; the next test verifies the complementary case - that a genuinely different
+  // `view` still does. See the comment on `viewOptions`'s memoization in Editor.tsx for why this
+  // matters.
   it("stays undoable across a re-render that passes a fresh-but-equal `options.view` object", async () => {
     const ref = createRef<EditorRef>();
     let editor: LexicalEditor | undefined;
@@ -352,5 +347,69 @@ describe("applyUpdate('local') undo-history retention", () => {
     await flushQueuedEvents();
 
     expect(canUndo).toBe(true);
+  });
+
+  it("clears undo history across a re-render that passes a genuinely different `view` object", async () => {
+    const ref = createRef<EditorRef>();
+    let editor: LexicalEditor | undefined;
+    const optionsA: EditorOptions = {
+      view: { markerMode: "visible", hasSpacing: true, isFormattedFont: false },
+    };
+    // A real view-mode switch - `markerMode` differs from `optionsA` - so the memoized
+    // `viewOptions` must produce a new value and let `LoadStatePlugin` reload as intended.
+    const optionsDifferentView: EditorOptions = {
+      view: { markerMode: "hidden", hasSpacing: true, isFormattedFont: false },
+    };
+
+    let rerender: ((element: React.ReactElement) => void) | undefined;
+    await act(async () => {
+      const result = render(
+        <Editor ref={ref} defaultUsj={blankUsj} options={optionsA}>
+          <GrabEditor onEditor={(e) => (editor = e)} />
+        </Editor>,
+      );
+      rerender = result.rerender;
+    });
+    await flushQueuedEvents();
+    if (!rerender) throw new Error("render did not return a rerender function");
+    const rerenderEditor = rerender;
+    if (!ref.current || !editor) throw new Error("EditorRef did not mount");
+
+    let canUndo = false;
+    editor.registerCommand<boolean>(
+      CAN_UNDO_COMMAND,
+      (payload) => {
+        canUndo = payload;
+        return false;
+      },
+      COMMAND_PRIORITY_CRITICAL,
+    );
+
+    await act(async () => {
+      ref.current?.applyUpdate(
+        [
+          { insert: { chapter: { number: "1", style: "c" } } },
+          { insert: { verse: { number: "1", style: "v" } } },
+        ],
+        "local",
+      );
+    });
+    await flushQueuedEvents();
+    expect(canUndo).toBe(true);
+
+    // A genuinely different `view` must still trigger LoadStatePlugin's reload and clear the
+    // undo stack it just built - proving the memo isn't short-circuiting real changes to always
+    // report "equal" (which would pass the previous test while breaking every real view-mode
+    // switch in the app).
+    await act(async () => {
+      rerenderEditor(
+        <Editor ref={ref} defaultUsj={blankUsj} options={optionsDifferentView}>
+          <GrabEditor onEditor={(e) => (editor = e)} />
+        </Editor>,
+      );
+    });
+    await flushQueuedEvents();
+
+    expect(canUndo).toBe(false);
   });
 });
--- a/packages/platform/src/editor/Editor.test.tsx
+++ b/packages/platform/src/editor/Editor.test.tsx
@@ -2,13 +2,18 @@
 // eslint-disable-next-line @nx/enforce-module-boundaries
 import { usjGen1v1 } from "../../../utilities/src/converters/usj/converter-test.data";
 import Editor from "./Editor";
-import { EditorRef } from "./editor.model";
+import { EditorOptions, EditorRef } from "./editor.model";
 import Editorial from "../Editorial";
 import { flushQueuedEvents } from "./editor-test.utils";
 import { ContentJsonPath, Usj } from "@eten-tech-foundation/scripture-utilities";
 import { useLexicalComposerContext } from "@lexical/react/LexicalComposerContext";
 import { act, render } from "@testing-library/react";
-import { KEY_DOWN_COMMAND, LexicalEditor } from "lexical";
+import {
+  CAN_UNDO_COMMAND,
+  COMMAND_PRIORITY_CRITICAL,
+  KEY_DOWN_COMMAND,
+  LexicalEditor,
+} from "lexical";
 import { createRef, RefObject, useEffect } from "react";
 import { vi } from "vitest";
 
@@ -268,5 +273,84 @@ describe("undo after a verse-spanning delete (PT-4102 regression)", () => {
     expect(afterUndo).toContain("Alpha ");
     expect(afterUndo).toContain("Bravo");
     expect(afterUndo).toContain('"number":"2"');
+  });
+});
+
+const blankUsj: Usj = {
+  type: "USJ",
+  version: "3.1",
+  content: [{ type: "book", marker: "id", code: "GEN", content: ["Test Book"] }],
+};
+
+describe("applyUpdate('local') undo-history retention", () => {
+  // A chapter-scaffold insert via `EditorRef.applyUpdate(ops, 'local')` must participate in the
+  // undo stack like any other local edit. It previously did not: `LoadStatePlugin`'s reload effect
+  // depends on `viewOptions`, which `Editor` derived with a plain `useMemo(() => view ??
+  // defaultViewOptions, [view])` - a reference-only memo. A parent re-render that passes a
+  // fresh-but-value-equal `options.view` object (e.g. one triggered by `applyUpdate`'s own
+  // `onUsjChange` round-trip completing, moments after the insert) changed `viewOptions`'s
+  // identity without changing its value, re-firing `LoadStatePlugin` and its unconditional
+  // `CLEAR_HISTORY_COMMAND` dispatch - wiping the just-created undo entry with no document or view
+  // change to justify it.
+  it("stays undoable across a re-render that passes a fresh-but-equal `options.view` object", async () => {
+    const ref = createRef<EditorRef>();
+    let editor: LexicalEditor | undefined;
+    // Two different object references with the same content - simulates a parent re-render that
+    // recomputes `options`/`view` without (or despite) its own memoization.
+    const optionsA: EditorOptions = {
+      view: { markerMode: "visible", hasSpacing: true, isFormattedFont: false },
+    };
+    const optionsB: EditorOptions = {
+      view: { markerMode: "visible", hasSpacing: true, isFormattedFont: false },
+    };
+
+    let rerender: ((element: React.ReactElement) => void) | undefined;
+    await act(async () => {
+      const result = render(
+        <Editor ref={ref} defaultUsj={blankUsj} options={optionsA}>
+          <GrabEditor onEditor={(e) => (editor = e)} />
+        </Editor>,
+      );
+      rerender = result.rerender;
+    });
+    await flushQueuedEvents();
+    if (!rerender) throw new Error("render did not return a rerender function");
+    const rerenderEditor = rerender;
+    if (!ref.current || !editor) throw new Error("EditorRef did not mount");
+
+    let canUndo = false;
+    editor.registerCommand<boolean>(
+      CAN_UNDO_COMMAND,
+      (payload) => {
+        canUndo = payload;
+        return false;
+      },
+      COMMAND_PRIORITY_CRITICAL,
+    );
+
+    await act(async () => {
+      ref.current?.applyUpdate(
+        [
+          { insert: { chapter: { number: "1", style: "c" } } },
+          { insert: { verse: { number: "1", style: "v" } } },
+        ],
+        "local",
+      );
+    });
+    await flushQueuedEvents();
+    expect(canUndo).toBe(true);
+
+    // Mirror the paranext-core sequence: a later re-render (the PDP round-trip settling) passes a
+    // fresh `options` object whose `view` is value-equal to the original but a different reference.
+    await act(async () => {
+      rerenderEditor(
+        <Editor ref={ref} defaultUsj={blankUsj} options={optionsB}>
+          <GrabEditor onEditor={(e) => (editor = e)} />
+        </Editor>,
+      );
+    });
+    await flushQueuedEvents();
+
+    expect(canUndo).toBe(true);
   });
 });
--- a/packages/platform/src/editor/Editor.test.tsx
+++ b/packages/platform/src/editor/Editor.test.tsx
@@ -14,7 +14,7 @@ import {
   KEY_DOWN_COMMAND,
   LexicalEditor,
 } from "lexical";
-import { createRef, RefObject, useEffect } from "react";
+import { createRef, RefObject, useEffect, useState } from "react";
 import { vi } from "vitest";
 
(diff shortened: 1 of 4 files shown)


katherinejensen00 (changes requested):
Thanks for your hard work on this, Levi! Just a few things to look at and it will be there.

> @katherinejensen00 reviewed 4 files and all commit messages, and made 9 comments.
> *[Reviewable](https://reviewable.io/reviews/eten-tech-foundation/scripture-editors/526#-:-OzNluaW4H4-dai4S0EH:b-2a24mk)* status: all files reviewed, 8 unresolved discussions (waiting on captaincrazybro).

___
*[`packages/platform/src/editor/Editor.tsx` line 160 at r1](https://reviewable.io/reviews/eten-tech-foundation/scripture-editors/526#-OzNlkPg3VClJIoiSi25:-OzNlkPg3VClJIoiSi26:b-1rbvz9) ([raw file](https://github.com/eten-tech-foundation/scripture-editors/blob/6b558a7b4de7926b671e9a7d4e51213a58dae391/packages/platform/src/editor/Editor.tsx#L160)):*
**Low**

> **The explanatory comment is accurate but can lose ~40% without losing the point.** Keep the causal chain (reload-effect dep → unconditional history clear → fresh-but-equal `view` re-fires it) and the `nodeOptions`/`contextMenuOptions` contrast; drop the `applyUpdate` parenthetical and the two restatements.
___
*[`packages/platform/src/editor/Editor.test.tsx` line 285 at r1](https://reviewable.io/reviews/eten-tech-foundation/scripture-editors/526#-OzNlTX82iPckLTOQBmp:-OzNlTX82iPckLTOQBmq:b-x4vj5f) ([raw file](https://github.com/eten-tech-foundation/scripture-editors/blob/6b558a7b4de7926b671e9a7d4e51213a58dae391/packages/platform/src/editor/Editor.test.tsx#L285)):*
**Medium**

> **The tests prove the memo comparator works, but not the integration that motivated it.** They re-render manually with a fresh-but-equal `options`; consider one test that wires a parent recomputing `options` inside `onUsjChange` (the actual round-trip), asserting undo survives.
___
*[`packages/platform/src/editor/Editor.test.tsx` line 352 at r1](https://reviewable.io/reviews/eten-tech-foundation/scripture-editors/526#-OzNlqLcC5DlvKIiTt5N:-OzNlqLcC5DlvKIiTt5O:b-v17pfx) ([raw file](https://github.com/eten-tech-foundation/scripture-editors/blob/6b558a7b4de7926b671e9a7d4e51213a58dae391/packages/platform/src/editor/Editor.test.tsx#L352)):*
**Low**

> **The two undo tests share ~95% setup — factor the boilerplate, but keep the one differing line visible.** Extract the `CAN_UNDO` registration and `applyUpdate` payload into a helper; leave the `optionsB`\-vs-`optionsDifferentView` and `true`\-vs-`false` explicit in each test so the retain-vs-clear contrast stays obvious.
___
*[`packages/platform/src/editor/adaptors/usj-marker-action.utils.ts` line 59 at r1](https://reviewable.io/reviews/eten-tech-foundation/scripture-editors/526#-OzNlYVP0gyMVXJfdJDN:-OzNlYVP0gyMVXJfdJDO:bz0s8fl) ([raw file](https://github.com/eten-tech-foundation/scripture-editors/blob/6b558a7b4de7926b671e9a7d4e51213a58dae391/packages/platform/src/editor/adaptors/usj-marker-action.utils.ts#L59)):*
**Low**

> **The fix line would benefit from a one-line comment; right now the rationale only lives in the test.** Something like: `// Chapter node already present → next chapter; none present (reinstating a missing \c) → keep the current number, don't increment.` This prevents a future "simplify to `+1`" from silently reintroducing the double-increment. (Also note the fix is intentionally narrow — it doesn't address the pre-existing duplicate-`\c` case when the next number already exists.)
___
*[`packages/platform/src/editor/Editor.tsx` line 161 at r1](https://reviewable.io/reviews/eten-tech-foundation/scripture-editors/526#-OzNl7qyA3hjEthyJOD4:-OzNl7qyA3hjEthyJOD5:b-t4yfkc) ([raw file](https://github.com/eten-tech-foundation/scripture-editors/blob/6b558a7b4de7926b671e9a7d4e51213a58dae391/packages/platform/src/editor/Editor.tsx#L161)):*
> ```XML
>   //
>   // `viewOptions` needs a VALUE-based (not just reference-based) memo: it's a dependency of
>   // `LoadStatePlugin`'s reload effect, which unconditionally calls `setEditorState` +
> ```

Medium (follow up)
**The comment correctly identifies that this effect clears history on every fire — which is the actual root cause, and it lives in shared `LoadStatePlugin`, not here.** Stabilizing `viewOptions` at this consumer treats the symptom; a follow-up to gate `CLEAR_HISTORY_COMMAND` in `LoadStatePlugin` on an actual document/state change would fix this for all consumers (including scribe, finding 2) and close the `logger` gap (finding 1) at once.
___
*[`packages/platform/src/editor/Editor.tsx` line 177 at r1](https://reviewable.io/reviews/eten-tech-foundation/scripture-editors/526#-OzNlcWV4CqSQbcYlG3S:-OzNlcWV4CqSQbcYlG3T:b-hqorhm) ([raw file](https://github.com/eten-tech-foundation/scripture-editors/blob/6b558a7b4de7926b671e9a7d4e51213a58dae391/packages/platform/src/editor/Editor.tsx#L177)):*
> ```XML
>   const resolvedViewOptions = view ?? defaultViewOptions;
>   const viewOptionsRef = useRef(resolvedViewOptions);
>   if (!deepEqual(viewOptionsRef.current, resolvedViewOptions)) {
> ```

Low 

**The mutate-a-ref-during-render is correct, but a one-liner on _why it's safe_ would preempt a "rules-of-React" flag.** It's idempotent and derived purely from this render's props, so StrictMode double-invoke and discarded concurrent renders are safe — and unlike `useMemo` (which React may discard) the ref actually guarantees identity stability. Consider extracting a small `useDeepCompareMemo` hook so the reasoning lives in one place (and scribe can reuse it).
___
*[`packages/platform/src/editor/Editor.tsx` line 192 at r1](https://reviewable.io/reviews/eten-tech-foundation/scripture-editors/526#-OzNkoSQ2fyKJoMr0Rg6:-OzNkoSQ2fyKJoMr0Rg7:b-old2ac) ([raw file](https://github.com/eten-tech-foundation/scripture-editors/blob/6b558a7b4de7926b671e9a7d4e51213a58dae391/packages/platform/src/editor/Editor.tsx#L192)):*
> ```XML
>       namespace: "platformEditor",
>       theme: { ...editorTheme, showCharMarkerTitles: viewOptions.showCharMarkerTitles },
>       editable: !isReadonly,
> ```

Medium (scope)

**The scribe editor has this same undo-loss bug and isn't covered here.** It passes `viewOptions` fully unmemoized into `LoadStatePlugin`. Since this PR fixes exactly that class of bug, a one-line "scribe is out of scope / follow-up filed" note in the PR would make the boundary deliberate.
___
*[`packages/platform/src/editor/Editor.tsx` line 463 at r1](https://reviewable.io/reviews/eten-tech-foundation/scripture-editors/526#-OzNkejB05BMaEcHhujb:-OzNkejB05BMaEcHhujc:b-38zsy3) ([raw file](https://github.com/eten-tech-foundation/scripture-editors/blob/6b558a7b4de7926b671e9a7d4e51213a58dae391/packages/platform/src/editor/Editor.tsx#L463)):*
> ```XML
>             editorAdaptor={usjEditorAdaptor}
>             viewOptions={viewOptions}
>             logger={logger}
> ```

Medium 

**The undo-history fix stabilizes `viewOptions`, but `logger` is another unmemoized dependency of the same reload effect, so history can still be wiped if a parent passes a fresh `logger`.** `LoadStatePlugin`'s reload effect depends on `[editor, editorAdaptor, logger, scripture, scriptureRef, viewOptions]` and clears history unconditionally on any fire. `logger` is forwarded raw here. It's usually a stable singleton so this is latent, but worth either stabilizing it the same way or noting the assumption.

packages/platform/src/editor/Editor.test.tsx
```diff
@@ -270,3 +275,200 @@ describe("undo after a verse-spanning delete (PT-4102 regression)", () => {
...
+  // The two tests above prove the memo comparator itself works, by manually rerendering with a
+  // hand-constructed `options` object. This test instead exercises the actual integration that
+  // motivated the fix: `applyUpdate` invokes `onUsjChange` directly (see the call in
+  // `applyUpdate`'s imperative handle above), and a parent that reacts to it - the real
+  // paranext-core round trip is `applyUpdate` -> PDP save -> PDP echo -> parent re-render - can
+  // recompute `options.view` as a fresh object with no memoization at all. `Wrapper` below
+  // deliberately does not memoize `view`, so every one of its re-renders (including the one
+  // `onUsjChange` triggers) constructs a brand new, value-equal object - reproducing the actual
+  // bug shape rather than simulating its end state.
+  it("stays undoable through a real onUsjChange-triggered parent re-render with an unmemoized `view`", async () => {
```
captaincrazybro, line 577: **Fixed.** Added this test, which exercises the actual round trip: `applyUpdate` calls `onUsjChange` directly (see the call inside `applyUpdate`'s imperative handle), and `Wrapper` here reacts to it by re-rendering with a genuinely unmemoized `options.view` object — no manual `rerender()` with a hand-picked fresh-but-equal object. Verified it actually catches the regression: temporarily reverted the `viewOptions` stabilization and confirmed this test fails (`expected false to be true`) while the two existing tests still pass, then restored the fix and confirmed all pass again.

packages/platform/src/editor/adaptors/usj-marker-action.utils.ts
```diff
@@ -54,11 +56,17 @@ const markerActions: { [marker: string]: UsjMarkerAction } = {
   c: {
     action: (currentEditor) => {
       const { chapterNum } = currentEditor.reference;
-      const nextChapter = chapterNum + 1;
+      // Chapter node already present → next chapter; none present (reinstating a missing `\c`
```
captaincrazybro, line 66: **Fixed.** Added the rationale comment (using your suggested wording, lightly adapted) plus a note on the intentionally-narrow scope: this fix does not check whether `chapterNum + 1` already exists elsewhere before incrementing into it, so the pre-existing duplicate `\c` case in that scenario is unchanged.

packages/platform/src/editor/Editor.tsx
```diff
@@ -156,10 +156,44 @@ const Editor = forwardRef(function Editor<TLogger extends LoggerBasic>(
...
   const contextMenuOptions = useMemo(() => contextMenu, [contextMenu]);
 
+  // `logger` is also a dependency of `LoadStatePlugin`'s reload effect (see the `viewOptions`
+  // comment above for what that effect does on every fire), so the same reference-instability
+  // risk applies here if a caller ever passes a fresh-but-equivalent logger object. Deep-equality
+  // is still the right comparison for an object whose properties are mostly methods: two
+  // genuinely different loggers won't have the same function references and will correctly be
+  // treated as different, while a caller that re-wraps the same underlying stable methods in a
+  // new object each render will correctly be treated as unchanged.
+  const loggerRef = useRef(logger);
```
captaincrazybro, line 196: **Fixed.** Applied the same `useRef` + `deepEqual` stabilization used for `viewOptions`, and updated every downstream consumer to use the stabilized `stableLogger` instead of the raw prop. Checked whether `deepEqual` (fast-equals) is actually sound for a mostly-function-bearing object first: traced its source and confirmed nested function properties fall back to reference equality, so two genuinely different loggers (different method references) are still correctly treated as different — this only absorbs a caller that re-wraps the same underlying stable methods in a fresh object each render, which is exactly the risk you flagged.

packages/platform/src/editor/Editor.tsx
```diff
@@ -156,10 +156,44 @@ const Editor = forwardRef(function Editor<TLogger extends LoggerBasic>(
   // a fresh `options` object every render. Pairs with the per-instance `initialConfig` below -
   // any state derived from `options` should follow the same pattern to avoid cross-instance
   // surprises with multiple Editor instances in one WebView.
-  const viewOptions = useMemo(() => view ?? defaultViewOptions, [view]);
+  //
+  // `viewOptions` needs a VALUE-based (not just reference-based) memo: it's a dependency of
```
captaincrazybro, line 165: Skipped for now. Agreed it restates the "value-based memo" point once as thesis and once as conclusion. Leaving as-is rather than touching prose in this area again mid-fix-round — happy to trim in a follow-up doc pass.

packages/platform/src/editor/Editor.tsx
```diff
@@ -156,10 +156,44 @@ const Editor = forwardRef(function Editor<TLogger extends LoggerBasic>(
...
+  // `LoadStatePlugin`'s reload effect, which unconditionally calls `setEditorState` +
```
captaincrazybro, line 166: Skipped for now (deferred, not dismissed). Confirmed this independently: `LoadStatePlugin`'s reload effect unconditionally clears history on every fire with no check that the resulting state actually differs from what's loaded. Fixing it there would close this for every consumer at once, including scribe (see #7's reply). It's a bigger, more invasive change to a shared plugin used across packages, and out of scope for this fix-round — tracking as a real follow-up rather than blocking here.

packages/platform/src/editor/Editor.tsx
```diff
@@ -156,10 +156,44 @@ const Editor = forwardRef(function Editor<TLogger extends LoggerBasic>(
...
+  // such re-renders while still producing a new one - correctly triggering a reload - when a view
+  // option genuinely changes.
+  //
+  // `nodeOptions` and `contextMenuOptions`, destructured just below, don't need this treatment:
+  // `nodeOptions` is only a dependency of `LoadStatePlugin`'s separate adaptor-*initialize* effect,
+  // not its reload effect, and `contextMenuOptions` isn't passed to `LoadStatePlugin` at all - so
+  // of the three, only `viewOptions`'s identity can trigger the spurious reload this fix addresses.
+  const resolvedViewOptions = view ?? defaultViewOptions;
+  const viewOptionsRef = useRef(resolvedViewOptions);
+  if (!deepEqual(viewOptionsRef.current, resolvedViewOptions)) {
```
captaincrazybro, line 182: Skipped for now. Agreed the pattern is correct but could use the one-liner on why (idempotent, derived purely from this render's props, so StrictMode double-invoke and discarded concurrent renders are safe). Deferring alongside the `useDeepCompareMemo` extraction suggestion — now that `logger` uses the identical pattern (#8), a shared hook is more clearly worth it, but as its own follow-up rather than reshuffling this fix-round further.

packages/platform/src/editor/Editor.tsx
```diff
@@ -156,10 +156,44 @@ const Editor = forwardRef(function Editor<TLogger extends LoggerBasic>(
...
+  // risk applies here if a caller ever passes a fresh-but-equivalent logger object. Deep-equality
+  // is still the right comparison for an object whose properties are mostly methods: two
+  // genuinely different loggers won't have the same function references and will correctly be
+  // treated as different, while a caller that re-wraps the same underlying stable methods in a
+  // new object each render will correctly be treated as unchanged.
+  const loggerRef = useRef(logger);
+  if (!deepEqual(loggerRef.current, logger)) {
+    loggerRef.current = logger;
+  }
+  const stableLogger = loggerRef.current;
```
captaincrazybro, line 200: Replying here (re: your finding on the old line 192/`initialConfig` area — the diff shifted, so GitHub wouldn't let me anchor a new comment there) about scribe being out of scope. Skipped for now, but confirmed accurate: `packages/scribe/src/editor/Editor.tsx` and `NoteEditor.tsx` also consume `LoadStatePlugin` and pass `viewOptions` straight through with no local memoization — the identical exposure this PR fixes, just in a package this PR doesn't touch. Genuinely out of scope here (different package, no shared caller with `platform`'s `Editor.tsx`). Adding the boundary note to the PR description per your suggestion rather than leaving it implicit. Real fix is the same `LoadStatePlugin` follow-up as the finding above.

packages/platform/src/editor/Editor.test.tsx
```diff
@@ -270,3 +275,200 @@ describe("undo after a verse-spanning delete (PT-4102 regression)", () => {
...
+          <GrabEditor onEditor={(e) => (editor = e)} />
+        </Editor>,
+      );
+    });
+    await flushQueuedEvents();
+
+    expect(canUndo).toBe(true);
+  });
+
+  it("clears undo history across a re-render that passes a genuinely different `view` object", async () => {
```
captaincrazybro, line 504: Skipped for now. Agreed this is real duplication, but with a third test now added for finding #2, extracting shared setup touches three tests at once — would rather do that extraction as its own follow-up cleanup pass than fold it into this fix-round's diff.

katherinejensen00 (approved):
Thanks for making all of those fixes, Levi!

> @katherinejensen00 reviewed 3 files and all commit messages, made 1 comment, and resolved 15 discussions.
> *[Reviewable](https://reviewable.io/reviews/eten-tech-foundation/scripture-editors/526#-:-OzRquZf4431JemhLfe5:ba6bqxb)* status: all files reviewed, 1 unresolved discussion (waiting on captaincrazybro).

PR #527, opened 2026-08-06
Fix marker-dropdown verse-number inference (T-A items 2+3)

captaincrazybro:
# Code Review Summary

**Branch**: fix-verse-marker-restoration-and-inference

**Base**: origin/main

**Date**: 2026-08-05

**Review model**: Claude Sonnet 5

**Files changed**: 5

## Overview

This branch addresses T-A items 2+3 from the design spec (verse-marker restoration reliability and verse-number inference). Two distinct pieces:

1. **Undo reliability** — investigated whether a run of verse-marker deletions could lose data on Undo/Redo (Build 223's complaint). Reproduced the scenario against unfixed code and found nothing is ever permanently lost: each deletion is already independently undo/redo-able, and pressing Undo the matching number of times fully restores everything, including the first-deleted verse, with Redo mirroring it exactly. No production fix was made; a regression test (`StructureKeyboardPlugin.test.tsx`) pins this behavior so a future refactor can't silently reintroduce data loss.
2. **Verse-number inference** — the marker-dropdown's `v` action previously derived the inserted verse number from a caller-supplied `reference.verseNum`/`.verse`, which can drift from the actual document and was producing garbled sequences ("1114"/"1244" per Build 223). Reworked to scan the real `VerseNode` sequence via a new `$findNextVerseAfter` helper plus the existing `$findThisVerse`, applying the PRD's inference rules (no preceding → 1; gap → increment; adjacent → increment + highlight; no following → increment).

## API Changes

- `libs/shared-react` (internal workspace lib, not published to npm): added new export `$findNextVerseAfter(verseNode: SomeVerseNode): SomeVerseNode | undefined` — a forward-search counterpart to the existing `$findPreviousVerseInSiblings`.
- `packages/platform` (published as `@eten-tech-foundation/platform-editor`): in the internal (non-exported) `usj-marker-action.utils.ts`, the internal `UsjMarkerAction.action` return type changed from `MarkerContent[]` to `{ content: MarkerContent[]; highlightInserted?: boolean }`.
- Neither of the above is re-exported from `packages/platform/src/index.ts` or `packages/scribe/src/index.ts` — the published npm surfaces of `@eten-tech-foundation/platform-editor` and `@eten-tech-foundation/scribe-editor` are unchanged.
- No changes to `packages/utilities` (`@eten-tech-foundation/scripture-utilities`) or any `.d.ts` files.

- [x] **`libs/shared-react/src/nodes/usj/node-react-utils.test.ts:265` was missing the `?.` that the other three assertions in the same `describe("$findNextVerseAfter()", ...)` block have** (`$findNextVerseAfter(verse1).getNumber()` instead of `$findNextVerseAfter(verse1)?.getNumber()`), which fails `tsc` with `TS2532: Object is possibly 'undefined'` since the helper returns `SomeVerseNode | undefined`. This would have failed CI's typecheck step. _(fixed during review: added the missing `?.`)_

Author confirmed and asked me to fix it immediately — verified independently by running `tsc` before and after (`TS2532` reproduced, then cleared).

- [x] **`packages/platform/src/editor/adaptors/usj-marker-action.utils.ts`: `highlightInserted` was computed via `parseInt`-truncated preceding/following verse numbers and a numeric diff (`followingNum - precedingNum === 1`), rather than comparing the actual computed `nextVerseNumber` to the following verse's number.** This breaks for verse segments and bridges, which `getNextVerse` explicitly supports:
  - Segment false positive: preceding `"5b"`, following `"6"` → `nextVerseNumber` computes to `"5c"` (no collision), but the parseInt diff (`5` vs `6`, diff `1`) wrongly set `highlightInserted = true`.
  - Bridge false negative: preceding `"5-6"`, following `"7"` → `nextVerseNumber` computes to `"7"` (a real collision with the following verse), but `parseInt("5-6")` truncates to `5`, diff `2` → wrongly set `highlightInserted = false`, silently missing a real collision.
  _(fixed during review: `highlightInserted` now computed as `!!followingVerse && nextVerseNumber === followingVerse.getNumber()` — simpler and correct for plain numbers, segments, and bridges alike. Verified against all four existing "verse-number inference" test cases plus the two edge cases above; full test suite re-run, 30/30 passing.)_

  This finding was independently reported by two of the four analysis passes (API and compliance), which is a strong corroborating signal.

- [ ] ~~`libs/shared-react/src/nodes/usj/node-react.utils.ts`: the new `$findNextVerseAfter` duplicates a tree-walk already inlined in `$selectNextVerse` in the same file (forward-scan remaining siblings, then cross paragraphs via `$findNextVerseInNode` until a chapter boundary). `$selectNextVerse` is live code, called from `ArrowNavigationPlugin.tsx` — not dead code. Consider refactoring `$selectNextVerse` to call `$findNextVerseAfter` instead of maintaining two copies of the same traversal.~~ _(Author: skipped for now — real duplication, but it touches a separately-tested, live arrow-key navigation path that's outside this branch's stated scope. Left as a follow-up opportunity rather than refactored here.)_

Author response: fixed the highlight-logic bug immediately upon hearing it explained (with the concrete segment/bridge counterexamples); explicitly deferred the `$selectNextVerse` duplication as out of scope for this branch.

### Minor — Consider

- [ ] ~~`packages/scribe/src/editor/adaptors/usj-marker-action.utils.ts` (pre-existing, untouched by this branch): scribe's own copy of the verse marker action has the same code shape (derives the next verse number from `currentEditor.reference.verseNum`/`.verse` when `autoNumbering` is true) that this branch fixes in `packages/platform`.~~ _(Author: skipped. On investigation, this is not a confirmed reproduced bug — unlike platform's, which had concrete evidence (the "1114"/"1244" reproduction) — just a similar code shape; I haven't traced scribe's own callers (`Toolbar.tsx`, `Editor.tsx`) to confirm they feed a stale reference the same way. It's also not a quick fix: scribe's `getUsjMarkerAction` still returns a plain `MarkerContent[]`, not the `{content, highlightInserted}` shape introduced here, so porting the fix would mean reworking scribe's interface too, plus preserving its separate `autoNumbering`/`newVerseRChapterNum` explicit-number-entry mechanism. It's also a different package/UI surface (`packages/scribe`) that T-A's ticket doesn't reference at all. Left as a candidate follow-up ticket, not pulled into this branch.)_
- [x] **`parseInt(precedingVerseString, 10)` was computed twice in the `v` action** — once for `nextVerseNumber`, again for the highlight-gap check. _(fixed during review: resolved automatically as a side effect of the Important #1 fix above, which removed the redundant `precedingNum`/`followingNum` computation entirely — only one `parseInt` call remains.)_

Author response: declined to expand scope to the scribe package; the double-`parseInt` finding turned out to already be resolved by the Important #1 fix.

### Template Propagation

Not applicable. This repo (`eten-tech-foundation/scripture-editors`) has no `#region shared with <url>` marker convention and no `extensions/` directory / template-repo propagation concept — these are paranext-core-specific compliance checks with no equivalent structure here.

#### Shared Regions Modified

N/A — convention doesn't exist in this repo.

#### Extension Config Changes

N/A — this repo has no `extensions/` directory.

## Positive Observations

- The verse-number-inference fix correctly moves from a stale external `reference` to walking the live Lexical tree, matching the stated purpose.
- `$findNextVerseAfter` has solid, focused test coverage: same-parent sibling, cross-paragraph, chapter-boundary stop, and no-next-verse-anywhere cases — mirroring the existing pattern used for `$findPreviousVerseInSiblings`.
- The "Build 223 regression" test is a well-targeted regression test tied to a concretely described real-world bug (re-adding a verse into a gap without repeating/garbling numbers).
- The undo/redo investigation was handled well: rather than churning production code on a hunch, a targeted per-edit LIFO undo/redo test was added to pin the existing (already-correct) behavior — consistent with "investigated, no fix needed" rather than a speculative production change.
- No published package surface (`packages/platform`, `packages/scribe`, `packages/utilities` `index.ts` files) was touched — the changes are internal-only, keeping blast radius low.
- Existing tests were updated in place (not just added to) where the underlying expected verse number changed (`"2"` → `"3"`), each with an explanatory inline comment, making the diff's intent easy to verify against the code.
- The verse-highlight mechanism (`$createNodeSelection()` / `.add(key)` / `$setSelection()`) exactly mirrors the existing "arm" pattern in `StructureKeyboardPlugin.tsx`, reusing an established selection mechanism rather than inventing a new one.
- New/modified imports are correctly alphabetized within their `$`-prefixed and non-`$`-prefixed groups, consistent with the rest of the codebase.
- All four call sites that build a `UsjMarkerAction.action` return value (`c`, `v`, para-marker, char-marker) were updated consistently to the new `{ content, highlightInserted? }` shape — no call site was missed.

## Interview Notes

**Stated purpose**: Fix undo reliability (investigated, no fix needed) and verse-number inference in the marker dropdown for T-A items 2+3.

**Key decisions and reasoning**:
- The undo-reliability "fix" is a regression test, not a production change — confirmed via direct reproduction (3 deletions → 3 Undo presses → full restoration in LIFO order; 3 Redo presses → full re-deletion, symmetric) that nothing is lost under today's code.
- The Critical typecheck failure and the Important highlight-logic bug were both fixed immediately once explained, with concrete before/after verification (re-running `tsc` and the affected test suite).
- Two findings (the `$selectNextVerse`/`$findNextVerseAfter` duplication, and the scribe package's similar-shaped code) were deliberately left as follow-ups rather than folded into this branch, on the reasoning that both would expand scope beyond this branch's stated purpose — one into a separately-tested live navigation path, the other into a different package with its own interface shape that isn't referenced by the T-A ticket.
- No areas of uncertainty or AI-deferred understanding were raised — all findings were resolved with direct, specific responses.

**Unresolved items**: None.

## In-Review Quality Check

- **Typecheck**: initially found a **new regression** not flagged by any of the four analysis passes — `TS2454: Variable '...' is used before being assigned` on 5 lines in `usj-marker-action-utils.test.ts`. This was a latent consequence of an earlier (already-committed, pre-interview) lint-fix commit that removed non-null assertions (`!`) from test fixtures to satisfy `@typescript-eslint/no-non-null-assertion` — that fix satisfied ESLint but broke `tsc`'s definite-assignment analysis for `let` variables declared and assigned inside a nested closure within the same test function. (This differs from the file's pre-existing `secondVerseTextNode` pattern, which is module-scoped and not subject to the same flow-sensitive check — the two situations look similar but aren't, which is how this slipped through.) Fixed by hoisting the affected fixtures (`noVerseText`, `verse1Text`, `insertedVerse2Text`) to module scope, matching the file's own established convention. Re-verified: `tsc` clean, ESLint clean, 30/30 tests in the file passing.
- **Full repo-wide typecheck** (`nx run-many -t typecheck`, all 10 projects): clean.
- **Lint** (targeted at the 5 changed files' directories): clean, no errors.
- **Full test suite** (`vitest run`, repo-wide): 1627 passed, 6 skipped (pre-existing, unrelated), 45 files passed / 1 skipped.

## Suggested Review Focus

- [ ] Confirm the undo/redo investigation's conclusion (no production fix needed) matches the reviewer's own expectation from the original Build 223 report — the regression test in `StructureKeyboardPlugin.test.tsx` is the durable artifact of that investigation.
- [ ] Spot-check the corrected `highlightInserted` logic (`usj-marker-action.utils.ts`) against a real segment/bridge verse scenario in the live editor if one is easy to set up, since the fix was verified against unit tests but not manually in the running app.
- [ ] Decide whether the `$selectNextVerse`/`$findNextVerseAfter` duplication and the scribe package's similar-shaped stale-reference code are worth their own follow-up tickets, or can be left for whoever next touches those areas.

- - -
This change is [Reviewable](https://reviewable.io/reviews/eten-tech-foundation/scripture-editors/527)

### DIFF
--- a/libs/shared-react/src/nodes/usj/node-react-utils.test.ts
+++ b/libs/shared-react/src/nodes/usj/node-react-utils.test.ts
@@ -262,7 +262,7 @@ describe("$findNextVerseAfter()", () => {
     );
 
     editor.getEditorState().read(() => {
-      expect($findNextVerseAfter(verse1).getNumber()).toBe("2");
+      expect($findNextVerseAfter(verse1)?.getNumber()).toBe("2");
     });
   });
 
--- a/libs/shared-react/src/nodes/usj/node-react-utils.test.ts
+++ b/libs/shared-react/src/nodes/usj/node-react-utils.test.ts
@@ -5,6 +5,7 @@ import { ViewOptions } from "../../views/view-options.utils";
 import { $isImmutableNoteCallerNode, ImmutableNoteCallerNode } from "./ImmutableNoteCallerNode";
 import { ImmutableVerseNode, $createImmutableVerseNode } from "./ImmutableVerseNode";
 import {
+  $findNextVerseAfter,
   $findPreviousVerseInSiblings,
   $findVerseInNode,
   $findVerseOrPara,
@@ -242,6 +243,81 @@ describe("$findPreviousVerseInSiblings()", () => {
       const verseNode = $findPreviousVerseInSiblings(p, 2);
 
       expect(verseNode?.getNumber()).toBe("1");
+    });
+  });
+});
+
+describe("$findNextVerseAfter()", () => {
+  it("finds the next verse sibling within the same parent", () => {
+    let verse1: VerseNode;
+    const { editor } = createBasicTestEnvironment([ParaNode, VerseNode]);
+    editor.update(
+      () => {
+        verse1 = $createVerseNode("1");
+        $getRoot().append(
+          $createParaNode().append(verse1, $createTextNode("text1"), $createVerseNode("2")),
+        );
+      },
+      { discrete: true },
+    );
+
+    editor.getEditorState().read(() => {
+      expect($findNextVerseAfter(verse1).getNumber()).toBe("2");
+    });
+  });
+
+  it("finds the next verse in a following paragraph", () => {
+    let verse1: VerseNode;
+    const { editor } = createBasicTestEnvironment([ParaNode, VerseNode]);
+    editor.update(
+      () => {
+        verse1 = $createVerseNode("1");
+        $getRoot().append(
+          $createParaNode().append(verse1, $createTextNode("text1")),
+          $createParaNode().append($createVerseNode("2"), $createTextNode("text2")),
+        );
+      },
+      { discrete: true },
+    );
+
+    editor.getEditorState().read(() => {
+      expect($findNextVerseAfter(verse1)?.getNumber()).toBe("2");
+    });
+  });
+
+  it("returns undefined when a chapter boundary is reached before another verse", () => {
+    let verse1: VerseNode;
+    const { editor } = createBasicTestEnvironment([ParaNode, VerseNode, ImmutableChapterNode]);
+    editor.update(
+      () => {
+        verse1 = $createVerseNode("1");
+        $getRoot().append(
+          $createParaNode().append(verse1, $createTextNode("text1")),
+          $createImmutableChapterNode("2"),
+          $createParaNode().append($createVerseNode("1"), $createTextNode("text2")),
+        );
+      },
+      { discrete: true },
+    );
+
+    editor.getEditorState().read(() => {
+      expect($findNextVerseAfter(verse1)).toBeUndefined();
+    });
+  });
+
+  it("returns undefined when there is no next verse anywhere", () => {
+    let verse1: VerseNode;
+    const { editor } = createBasicTestEnvironment([ParaNode, VerseNode]);
+    editor.update(
+      () => {
+        verse1 = $createVerseNode("1");
+        $getRoot().append($createParaNode().append(verse1, $createTextNode("text1")));
+      },
+      { discrete: true },
+    );
+
+    editor.getEditorState().read(() => {
+      expect($findNextVerseAfter(verse1)).toBeUndefined();
     });
   });
 });
--- a/libs/shared-react/src/nodes/usj/node-react.utils.ts
+++ b/libs/shared-react/src/nodes/usj/node-react.utils.ts
@@ -426,6 +426,32 @@ export function $findPreviousVerseInSiblings(
 }
 
 /**
+ * Find the next verse node after `verseNode` in document order, stopping at the next chapter
+ * boundary (or the end of the document). Forward counterpart to `$findPreviousVerseInSiblings` /
+ * the backward walk in `$findThisVerse`.
+ * @param verseNode - The verse node to search forward from.
+ * @returns The next verse node, or `undefined` if none exists before the next chapter/document end.
+ */
+export function $findNextVerseAfter(verseNode: SomeVerseNode): SomeVerseNode | undefined {
+  const parent = verseNode.getParent();
+  if (parent && $isElementNode(parent)) {
+    const children = parent.getChildren();
+    for (let i = verseNode.getIndexWithinParent() + 1; i < children.length; i++) {
+      const child = children[i];
+      if ($isSomeVerseNode(child)) return child;
+    }
+  }
+
+  let nextPara = parent?.getNextSibling();
+  while (nextPara && !$isSomeChapterNode(nextPara)) {
+    const verse = $findNextVerseInNode(nextPara);
+    if (verse) return verse;
+    nextPara = nextPara.getNextSibling();
+  }
+  return undefined;
+}
+
+/**
  * Find the last verse in the children of the node.
  * @param node - Node with potential verses in children.
  * @returns the verse node if found, `undefined` otherwise.
--- a/libs/shared-react/src/plugins/usj/StructureKeyboardPlugin.test.tsx
+++ b/libs/shared-react/src/plugins/usj/StructureKeyboardPlugin.test.tsx
@@ -3,6 +3,7 @@
 
 import { $createImmutableVerseNode, ImmutableVerseNode } from "../../nodes/usj";
 import { StructureKeyboardPlugin } from "./StructureKeyboardPlugin";
+import { HistoryPlugin } from "../History/HistoryPlugin";
 import { baseTestEnvironment, pressKey, updateSelection } from "./react-test.utils";
 import { act } from "@testing-library/react";
 import {
@@ -19,8 +20,10 @@ import {
   DRAGSTART_COMMAND,
   DROP_COMMAND,
   PASTE_COMMAND,
+  UNDO_COMMAND,
+  REDO_COMMAND,
 } from "lexical";
-import { $createParaNode, $isParaNode, ParaNode } from "shared";
+import { $createParaNode, $isParaNode, createEmptyHistoryState, ParaNode } from "shared";
 
 // NOTE: jsdom cannot drive collapsed mid-text deletes (domSelection.modify) or printable-char
 // insertion via dispatchCommand, and IS_APPLE is false so Alt+Backspace is a no-op. Those cases
@@ -593,6 +596,89 @@ describe("StructureKeyboardPlugin — two-step delete for range selections with 
       expect(para.getChildren().some((n) => n instanceof ImmutableVerseNode)).toBe(true); // survived
       expect(para.getTextContent()).toBe("abcd");
     });
+  });
+});
+
+describe("StructureKeyboardPlugin — per-edit undo/redo for a run of verse deletions", () => {
+  it("restores and re-deletes verses one at a time via Undo/Redo, with no data loss", async () => {
+    let tailText: TextNode;
+    const historyState = createEmptyHistoryState();
+    const { editor } = await baseTestEnvironment(
+      () => {
+        tailText = $createTextNode("text");
+        $getRoot().append(
+          $createParaNode("p").append(
+            $createImmutableVerseNode("1"),
+            $createImmutableVerseNode("2"),
+            $createImmutableVerseNode("3"),
+            tailText,
+          ),
+        );
+      },
+      <>
+        <StructureKeyboardPlugin structureProtectionMode="guarded" />
+        <HistoryPlugin externalHistoryState={historyState} />
+      </>,
+    );
+    updateSelection(editor, tailText!, 0);
+
+    // Delete verse 3, then verse 2, then verse 1 - each via the two-step Backspace gesture.
+    await pressKey(editor, "Backspace", 0); // arm verse 3
+    await pressKey(editor, "Backspace", 0); // fire: verse 3 removed
+    await pressKey(editor, "Backspace", 0); // arm verse 2
+    await pressKey(editor, "Backspace", 0); // fire: verse 2 removed
+    await pressKey(editor, "Backspace", 0); // arm verse 1
+    await pressKey(editor, "Backspace", 0); // fire: verse 1 removed
+
+    const readVerseNumbers = () =>
+      editor.getEditorState().read(() => {
+        const para = $getRoot().getChildren()[0] as ParaNode;
+        return para
+          .getChildren()
+          .filter((n): n is ImmutableVerseNode => n instanceof ImmutableVerseNode)
+          .map((n) => n.getNumber());
+      });
+
+    expect(readVerseNumbers()).toEqual([]);
+
+    // Each Undo press restores exactly the deletion it corresponds to, LIFO - proving no deletion
+    // (including the FIRST one, verse 3) is ever permanently lost from the undo stack.
+    await act(async () => {
+      editor.dispatchCommand(UNDO_COMMAND, undefined);
+    });
+    expect(readVerseNumbers()).toEqual(["1"]);
+
+    await act(async () => {
+      editor.dispatchCommand(UNDO_COMMAND, undefined);
+    });
+    expect(readVerseNumbers()).toEqual(["1", "2"]);
+
+    await act(async () => {
+      editor.dispatchCommand(UNDO_COMMAND, undefined);
+    });
+    expect(readVerseNumbers()).toEqual(["1", "2", "3"]);
+
+    // A further Undo is a no-op - nothing left to undo.
+    await act(async () => {
+      editor.dispatchCommand(UNDO_COMMAND, undefined);
+    });
+    expect(readVerseNumbers()).toEqual(["1", "2", "3"]);
+
+    // Redo walks back through the same sequence in reverse, fully symmetric.
+    await act(async () => {
+      editor.dispatchCommand(REDO_COMMAND, undefined);
+    });
+    expect(readVerseNumbers()).toEqual(["1", "2"]);
+
+    await act(async () => {
+      editor.dispatchCommand(REDO_COMMAND, undefined);
+    });
+    expect(readVerseNumbers()).toEqual(["1"]);
+
+    await act(async () => {
+      editor.dispatchCommand(REDO_COMMAND, undefined);
+    });
+    expect(readVerseNumbers()).toEqual([]);
   });
 });
 
--- a/packages/platform/src/editor/adaptors/usj-marker-action-utils.test.ts
+++ b/packages/platform/src/editor/adaptors/usj-marker-action-utils.test.ts
@@ -29,6 +29,9 @@ const nodes = usjReactNodes;
 const reference = { book: "GEN", chapterNum: 1, verseNum: 1 };
 
 let secondVerseTextNode: TextNode;
+let noVerseText: TextNode;
+let verse1Text: TextNode;
+let insertedVerse2Text: TextNode;
 
 function $defaultInitialEditorState() {
   secondVerseTextNode = $createTextNode("second verse text ");
@@ -145,7 +148,6 @@ describe("USJ Marker Action Utils", () => {
 
   describe("verse-number inference", () => {
     it("inserts verse 1 with no highlight when no verse precedes the caret", () => {
-      let noVerseText: TextNode;
       const { editor } = createBasicTestEnvironment(nodes, () => {
         noVerseText = $createTextNode("no verse yet");
         $getRoot().append($createImmutableChapterNode("1"), $createParaNode().append(noVerseText));
@@ -173,7 +175,6 @@ describe("USJ Marker Action Utils", () => {
     });
 
     it("increments across a gap with no highlight", () => {
-      let verse1Text: TextNode;
       const { editor } = createBasicTestEnvironment(nodes, () => {
         verse1Text = $createTextNode("verse one ");
         $getRoot().append(
@@ -206,7 +207,6 @@ describe("USJ Marker Action Utils", () => {
     });
 
     it("increments and highlights when preceding and following verses are adjacent (no numeric slot)", () => {
-      let verse1Text: TextNode;
       const { editor } = createBasicTestEnvironment(nodes, () => {
         verse1Text = $createTextNode("first verse text ");
         $getRoot().append(
@@ -244,7 +244,6 @@ describe("USJ Marker Action Utils", () => {
     });
 
     it("does not repeat/garble verse numbers when re-adding into a gap (Build 223 regression)", () => {
-      let verse1Text: TextNode;
       const { editor } = createBasicTestEnvironment(nodes, () => {
         // "verse one " (10 chars) + "text" (4 chars): caret splits after the first part, leaving
         // "text" as a trailing node so a second insertion can be anchored right after the first.
@@ -266,7 +265,6 @@ describe("USJ Marker Action Utils", () => {
       updateSelection(editor, verse1Text, 10);
       markerAction.action({ editor, reference });
 
-      let insertedVerse2Text: TextNode;
       editor.getEditorState().read(() => {
         const para = $getRoot().getChildren()[1];
         if (!$isParaNode(para)) throw new Error("Expected a ParaNode");
--- a/packages/platform/src/editor/adaptors/usj-marker-action-utils.test.ts
+++ b/packages/platform/src/editor/adaptors/usj-marker-action-utils.test.ts
@@ -6,7 +6,14 @@ import {
   updateSelection,
 } from "../../../../../libs/shared/src/nodes/usj/test.utils";
 import { getUsjMarkerAction, isUsjMarkerSupported } from "./usj-marker-action.utils";
-import { $createTextNode, $getRoot, $isTextNode, TextNode } from "lexical";
+import {
+  $createTextNode,
+  $getRoot,
+  $getSelection,
+  $isNodeSelection,
+  $isTextNode,
+  TextNode,
+} from "lexical";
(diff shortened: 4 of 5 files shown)


katherinejensen00 (changes requested):
Thanks for all your work on this, Levi! Definitely a helpful fix.

NIT: **The summary is ~1,900 words** with fully-quoted resolved findings and four N/A template blocks — collapse resolved items and drop the N/A sections so the open items stand out.

> @katherinejensen00 made 19 comments.
> *[Reviewable](https://reviewable.io/reviews/eten-tech-foundation/scripture-editors/527#-:-OzOKnk55VM8rbC_sFgL:b-v1po3d)* status: 0 of 5 files reviewed, 18 unresolved discussions (waiting on captaincrazybro).

___
*[`libs/shared-react/src/nodes/usj/node-react-utils.test.ts` line 250 at r1](https://reviewable.io/reviews/eten-tech-foundation/scripture-editors/527#-OzOKG3p-_LbLCk8cJF1:-OzOKG3p-_LbLCk8cJF2:b-b1nlti) ([raw file](https://github.com/eten-tech-foundation/scripture-editors/blob/dc6a1616f6dbb6760a68b020aeedd7426b31e136/libs/shared-react/src/nodes/usj/node-react-utils.test.ts#L250)):*
> ```XML
> });
> 
> describe("$findNextVerseAfter()", () => {
> ```

NIT

**No test for a verse-less intermediate paragraph** between the anchor and the next verse — the skip-and-continue branch is unguarded.
___
*[`libs/shared-react/src/nodes/usj/node-react.utils.ts` line 435 at r1](https://reviewable.io/reviews/eten-tech-foundation/scripture-editors/527#-OzOIQzM5mbv_ibJFMyk:-OzOIQzM5mbv_ibJFMyl:bo0xpb6) ([raw file](https://github.com/eten-tech-foundation/scripture-editors/blob/dc6a1616f6dbb6760a68b020aeedd7426b31e136/libs/shared-react/src/nodes/usj/node-react.utils.ts#L435)):*
> ```XML
>  * @returns The next verse node, or `undefined` if none exists before the next chapter/document end.
>  */
> export function $findNextVerseAfter(verseNode: SomeVerseNode): SomeVerseNode | undefined {
> ```

Low

**`   $findNextVerseAfter` duplicates the walk still inlined in `$selectNextVerse`.** Deferred is fine; note `$selectPreviousVerse` already delegates to its shared helper, so the forward path is the lone holdout — keep both copies in sync until unified.
___
*[`libs/shared-react/src/nodes/usj/node-react.utils.ts` line 445 at r1](https://reviewable.io/reviews/eten-tech-foundation/scripture-editors/527#-OzOJteo5Mk15qkaNeEL:-OzOJteo5Mk15qkaNeEM:b-l52wru) ([raw file](https://github.com/eten-tech-foundation/scripture-editors/blob/dc6a1616f6dbb6760a68b020aeedd7426b31e136/libs/shared-react/src/nodes/usj/node-react.utils.ts#L445)):*
> ```XML
>   }
> 
>   let nextPara = parent?.getNextSibling();
> ```

NIT

**`   $findNextVerseAfter` assumes verses are direct children of top-level blocks** (flat scan + immediate-parent sibling walk); a nested verse is missed. Consistent with existing helpers — noted only for completeness.
___
*[`libs/shared-react/src/plugins/usj/StructureKeyboardPlugin.test.tsx` line 603 at r1](https://reviewable.io/reviews/eten-tech-foundation/scripture-editors/527#-OzO4jC3BSlsXot-0FZH:-OzO4jC3BSlsXot-0FZI:b-rlcufh) ([raw file](https://github.com/eten-tech-foundation/scripture-editors/blob/dc6a1616f6dbb6760a68b020aeedd7426b31e136/libs/shared-react/src/plugins/usj/StructureKeyboardPlugin.test.tsx#L603)):*
> ```XML
> 
> describe("StructureKeyboardPlugin — per-edit undo/redo for a run of verse deletions", () => {
>   it("restores and re-deletes verses one at a time via Undo/Redo, with no data loss", async () => {
> ```

**Low**  **The undo/redo "no data loss" test only checks verse-marker numbers, not that the text survives.** Every assertion filters to `ImmutableVerseNode`; a dropped/duplicated tail `TextNode` would pass unnoticed. Add a `tailText` content assertion at each undo/redo step to actually pin "no data loss." 
___
*[`packages/platform/src/editor/adaptors/usj-marker-action-utils.test.ts` line 214 at r1](https://reviewable.io/reviews/eten-tech-foundation/scripture-editors/527#-OzO5UtoBiTFhCVes0So:-OzO5UtpBW9N9pd9wFDb:bu0d9fs) ([raw file](https://github.com/eten-tech-foundation/scripture-editors/blob/dc6a1616f6dbb6760a68b020aeedd7426b31e136/packages/platform/src/editor/adaptors/usj-marker-action-utils.test.ts#L214)):*
> ```XML
>         $getRoot().append(
>           $createImmutableChapterNode("1"),
>           $createParaNode().append($createImmutableVerseNode("1"), verse1Text),
> ```

**The negative-highlight tests can't distinguish "no collision" from "collision detection is dead," and the bridge case from #1 is untested.** They assert only `$isNodeSelection === false`; a `$findNextVerseAfter` that always returned `undefined` would pass all of them. Add a negative-highlight test with a bridge/segment follower.
___
*[`packages/platform/src/editor/adaptors/usj-marker-action-utils.test.ts` line 246 at r1](https://reviewable.io/reviews/eten-tech-foundation/scripture-editors/527#-OzOKSKIA4mOXm5Kmn5z:-OzOKSKIA4mOXm5Kmn6-:b-q34hgg) ([raw file](https://github.com/eten-tech-foundation/scripture-editors/blob/dc6a1616f6dbb6760a68b020aeedd7426b31e136/packages/platform/src/editor/adaptors/usj-marker-action-utils.test.ts#L246)):*
> ```XML
>     });
> 
>     it("does not repeat/garble verse numbers when re-adding into a gap (Build 223 regression)", () => {
> ```

NIT

**Test name embeds "Build 223"** — prefer forward-facing naming (behavior + rationale, not build history).
___
*[`packages/platform/src/editor/adaptors/usj-marker-action.utils.ts` line 4 at r1](https://reviewable.io/reviews/eten-tech-foundation/scripture-editors/527#-OzO6JJT8LhlUaPgeRWa:-OzO6JJT8LhlUaPgeRWb:b29gqvi) ([raw file](https://github.com/eten-tech-foundation/scripture-editors/blob/dc6a1616f6dbb6760a68b020aeedd7426b31e136/packages/platform/src/editor/adaptors/usj-marker-action.utils.ts#L4)):*
> ```XML
> import { SerializedVerseRef } from "@sillsdev/scripture";
> import {
>   $createNodeSelection,
> ```

Low

**Scribe's copy still ships the stale-reference logic** you fixed here. You deferred this deliberately; noting the shared seam already exists in `shared-react`, so a unifying follow-up ticket would stop the drift.
___
*[`packages/platform/src/editor/adaptors/usj-marker-action.utils.ts` line 20 at r1](https://reviewable.io/reviews/eten-tech-foundation/scripture-editors/527#-OzOI6y43F7-fGZZqKzo:-OzOI6y43F7-fGZZqKzp:b3doxn8) ([raw file](https://github.com/eten-tech-foundation/scripture-editors/blob/dc6a1616f6dbb6760a68b020aeedd7426b31e136/packages/platform/src/editor/adaptors/usj-marker-action.utils.ts#L20)):*
> ```XML
>   $createNodeFromSerializedNode,
>   $isMarkerNode,
>   $isNoteNode,
> ```

Low

**`highlightInserted` is honored on only one insertion branch;** wrap-text/paragraph/note/non-range paths drop it, and NodesMenu's auto-close ignores the new `NodeSelection`. Benign today — add a comment marking the scope as deliberate.
___
*[`packages/platform/src/editor/adaptors/usj-marker-action.utils.ts` line 47 at r1](https://reviewable.io/reviews/eten-tech-foundation/scripture-editors/527#-OzOJ9aGATAtNh-bf2z7:-OzOJ9aGATAtNh-bf2z8:b-vv793y) ([raw file](https://github.com/eten-tech-foundation/scripture-editors/blob/dc6a1616f6dbb6760a68b020aeedd7426b31e136/packages/platform/src/editor/adaptors/usj-marker-action.utils.ts#L47)):*
> ```XML
> import usjEditorAdaptor from "./usj-editor.adaptor";
> 
> interface UsjMarkerActionResult {
> ```

NIT

**The result envelope threads a boolean only `v` uses through all four branches** — acceptable, but it's this diff's churn source.
___
*[`packages/platform/src/editor/adaptors/usj-marker-action.utils.ts` line 59 at r1](https://reviewable.io/reviews/eten-tech-foundation/scripture-editors/527#-OzOIp9W-KlHBeRh1-dZ:-OzOIp9W-KlHBeRh1-d_:b-p48pz4) ([raw file](https://github.com/eten-tech-foundation/scripture-editors/blob/dc6a1616f6dbb6760a68b020aeedd7426b31e136/packages/platform/src/editor/adaptors/usj-marker-action.utils.ts#L59)):*
> ```XML
>     editor: LexicalEditor;
>     reference: ScriptureReference;
>     autoNumbering?: boolean;
> ```

NIT

**`autoNumbering`/`newVerseRChapterNum` are declared but unread** by the new `v` action — dead param surface that a future caller could pass expecting an effect.
___
*[`packages/platform/src/editor/adaptors/usj-marker-action.utils.ts` line 68 at r1](https://reviewable.io/reviews/eten-tech-foundation/scripture-editors/527#-OzO69tfAJJB1wrsDEMK:-OzO69tfAJJB1wrsDEML:bso8shf) ([raw file](https://github.com/eten-tech-foundation/scripture-editors/blob/dc6a1616f6dbb6760a68b020aeedd7426b31e136/packages/platform/src/editor/adaptors/usj-marker-action.utils.ts#L68)):*
> ```XML
>   c: {
>     action: (currentEditor) => {
>       const { chapterNum } = currentEditor.reference;
> ```

Low

**The `c` action still trusts `reference.chapterNum` — the input `v` was rewritten to distrust.** Not a proven bug, but leave a comment that `c` intentionally still uses the reference, or fix alongside.
___
*[`packages/platform/src/editor/adaptors/usj-marker-action.utils.ts` line 79 at r1](https://reviewable.io/reviews/eten-tech-foundation/scripture-editors/527#-OzO5z1B2yM-IKi9sqYX:-OzO5z1B2yM-IKi9sqYY:b-bk76sh) ([raw file](https://github.com/eten-tech-foundation/scripture-editors/blob/dc6a1616f6dbb6760a68b020aeedd7426b31e136/packages/platform/src/editor/adaptors/usj-marker-action.utils.ts#L79)):*
> ```XML
>   },
>   v: {
>     action: () => {
> ```

Low

**`   v` now ignores `reference` entirely; with no selection it inserts verse "1" regardless of `scrRef`.** Intended, but a behavior change for any `insertMarker("v")` caller without an active selection. Worth one line in the PR body flagging the changed path. 
___
*[`packages/platform/src/editor/adaptors/usj-marker-action.utils.ts` line 84 at r1](https://reviewable.io/reviews/eten-tech-foundation/scripture-editors/527#-OzOJ3um0A2uByW-rfzm:-OzOJ3um0A2uByW-rfzn:b-x9gdjf) ([raw file](https://github.com/eten-tech-foundation/scripture-editors/blob/dc6a1616f6dbb6760a68b020aeedd7426b31e136/packages/platform/src/editor/adaptors/usj-marker-action.utils.ts#L84)):*
> ```XML
>       const precedingVerse = $findThisVerse(anchorNode);
> 
>       let nextVerseNumber: string;
> ```

NIT

**he two mutable `let`s + else-only branch flatten** to a const/helper returning `{ number, highlightInserted }`.
___
*[`packages/platform/src/editor/adaptors/usj-marker-action.utils.ts` line 90 at r1](https://reviewable.io/reviews/eten-tech-foundation/scripture-editors/527#-OzOIyS0ByS8JnLXvhAr:-OzOIyS0ByS8JnLXvhAs:b43dhju) ([raw file](https://github.com/eten-tech-foundation/scripture-editors/blob/dc6a1616f6dbb6760a68b020aeedd7426b31e136/packages/platform/src/editor/adaptors/usj-marker-action.utils.ts#L90)):*
> ```XML
>       } else {
>         const precedingVerseString = precedingVerse.getNumber();
>         nextVerseNumber = getNextVerse(parseInt(precedingVerseString, 10), precedingVerseString);
> ```

NIT

**`parseInt(precedingVerseString, 10)` is dead work** — `getNextVerse` ignores its first arg when the verse string is present. The `,10` also gives false reassurance since `getNextVerse`'s own `parseInt`s are radix-less.
___
*[`packages/platform/src/editor/adaptors/usj-marker-action.utils.ts` line 90 at r1](https://reviewable.io/reviews/eten-tech-foundation/scripture-editors/527#-OzOK6LC-J9GHcWDqTVe:-OzOK6LC-J9GHcWDqTVf:bvb3krm) ([raw file](https://github.com/eten-tech-foundation/scripture-editors/blob/dc6a1616f6dbb6760a68b020aeedd7426b31e136/packages/platform/src/editor/adaptors/usj-marker-action.utils.ts#L90)):*
> ```XML
>       } else {
>         const precedingVerseString = precedingVerse.getNumber();
>         nextVerseNumber = getNextVerse(parseInt(precedingVerseString, 10), precedingVerseString);
> ```

NIT
**Pre-existing `getNextVerse` fragility** (empty/exotic numbers → `"NaN"`/`"1e+21"`) is now sourced from `getNumber()`. Out of scope, but the input source changed.
___
*[`packages/platform/src/editor/adaptors/usj-marker-action.utils.ts` line 92 at r1](https://reviewable.io/reviews/eten-tech-foundation/scripture-editors/527#-OzOKfJp5v6l5Roah00n:-OzOKfJp5v6l5Roah00o:bx0ao1a) ([raw file](https://github.com/eten-tech-foundation/scripture-editors/blob/dc6a1616f6dbb6760a68b020aeedd7426b31e136/packages/platform/src/editor/adaptors/usj-marker-action.utils.ts#L92)):*
> ```XML
>         nextVerseNumber = getNextVerse(parseInt(precedingVerseString, 10), precedingVerseString);
>         const followingVerse = $findNextVerseAfter(precedingVerse);
>         // Compare the actual inserted number against the following verse's number directly
> ```

NIT
 **   Comment overclaims ("truly collide")** — pairs with #1; reword to reflect exact-string semantics. 
___
*[`packages/platform/src/editor/adaptors/usj-marker-action.utils.ts` line 95 at r1](https://reviewable.io/reviews/eten-tech-foundation/scripture-editors/527#-OzO3fxvAYjh4FFAvxWb:-OzO3fxvAYjh4FFAvxWc:b-tan68l) ([raw file](https://github.com/eten-tech-foundation/scripture-editors/blob/dc6a1616f6dbb6760a68b020aeedd7426b31e136/packages/platform/src/editor/adaptors/usj-marker-action.utils.ts#L95)):*
> ```XML
>         // (not a parsed-integer diff), so segments ("5b" -> "5c") and bridges ("5-6" -> "7")
>         // are judged by whether they truly collide, not by their leading digit's distance.
>         highlightInserted = !!followingVerse && nextVerseNumber === followingVerse.getNumber();
> ```

Low-Medium

**The "no numeric slot" highlight can silently fail to fire when the next verse is a bridge or segment.** The check is exact string equality against only the immediate next verse, so preceding `"4"` \+ following `"5-6"` compares `"5" === "5-6"` → false, and no highlight appears even though 5 is inside the bridge. `isVerseInRange(5,"5-6")` (node.utils.ts:716) already returns true and would catch this. Also soften the comment at lines 92–94 — "truly collide" overstates what exact equality guarantees.
___
*[`packages/platform/src/editor/adaptors/usj-marker-action.utils.ts` line 207 at r1](https://reviewable.io/reviews/eten-tech-foundation/scripture-editors/527#-OzOJbNpER6SfrtCHdkV:-OzOJbNpER6SfrtCHdkW:b-blkyhv) ([raw file](https://github.com/eten-tech-foundation/scripture-editors/blob/dc6a1616f6dbb6760a68b020aeedd7426b31e136/packages/platform/src/editor/adaptors/usj-marker-action.utils.ts#L207)):*
> ```XML
>           selection.insertNodes([nodeToInsert]);
>           $moveVerseFollowingSpaceToPreviousNode(nodeToInsert);
>           if (highlightInserted) {
> ```

NIT

**The highlight `NodeSelection` has no "still-attached" fallback** (the else-branch does), so it's fragile if a future refactor detaches the node during space-move.

captaincrazybro:
Addressed the review feedback below (referencing your file:line locations). Note: since this review came through Reviewable.io as one bundled review body rather than native GitHub inline comments, I'm replying here as a single top-level comment rather than threaded per-item.

**Fixed:**

- `usj-marker-action.utils.ts:95` (Low-Medium, bridge-highlight gap) — Confirmed. `highlightInserted` now also checks `isVerseRange(followingVerse) && isVerseInRange(nextVerseNumber, followingVerse)` alongside the exact-string match, gated on the following verse actually being a range so it doesn't false-positive against an unrelated segment that merely shares a leading digit (e.g. "5c" vs "5d"). Added a positive test (bridge "5-6" collides with inserted "5") and reworded the "truly collide" comment.
- `usj-marker-action.utils.ts:90` (dead `parseInt`) — Removed; `getNextVerse` ignores its first argument whenever `verse` is given.
- `usj-marker-action.utils.ts:20` (`highlightInserted` scope) — Added a comment noting it's deliberately only honored on the collapsed-caret insert branch.
- `usj-marker-action.utils.ts:68` (`c` still trusts `reference.chapterNum`) — Added a comment noting this is owned by the separate chapter-number-reinstatement branch/PR, not left inconsistent by accident.
- `node-react-utils.test.ts:250` (missing intermediate-paragraph test) — Added a test where the skip-and-continue loop walks past a verse-less paragraph before finding the next verse.
- `StructureKeyboardPlugin.test.tsx:603` (undo/redo test doesn't check text) — Added a `para.getTextContent()` assertion at every undo/redo step, so a dropped/duplicated tail `TextNode` would actually be caught.
- `usj-marker-action-utils.test.ts:214` (negative-highlight tests can't distinguish dead detection) — Added both a positive bridge-collision test and a negative bridge-non-collision test (bridge "6-7" does *not* collide with inserted "5"), so the lookup itself is exercised, not just its absence.
- `usj-marker-action-utils.test.ts:246` (test name embeds "Build 223") — Renamed to describe the behavior.

**Evaluated, not changing for this PR:**

- `usj-marker-action.utils.ts:79` (`v` ignores `reference` entirely with no selection, always inserts "1") — Confirmed as a real behavior change from the old code, which always used `reference.verseNum`/`.verse` unconditionally regardless of selection state. After discussion, decided not to add a reference-based fallback in this change — flagging here for visibility in case it needs follow-up.
- `usj-marker-action.utils.ts:84` (flatten the two `let`s into a helper) — Style preference; leaving the current shape as-is for this PR.

All other points (the `$selectNextVerse` duplication, scribe's stale-reference copy, the nested-verse limitation, the result-envelope churn, `autoNumbering`/`newVerseRChapterNum` being pre-existing dead surface, `getNextVerse`'s pre-existing fragility, and the `NodeSelection` fallback) were already noted in your review as deferred/informational/out-of-scope, so no action taken there.

Full repo typecheck, lint, and test suite (1630 passing) all re-verified green after these changes.

PR #528, opened 2026-08-06
feat(platform): ship a usable editor stylesheet

henrique221:
Closes part of #516 — the CSS half. Bundle-size trimming (the `@xmldom`/`@lexical/table`/yjs cuts) is not in here.

## Demo

A short screencast (2 min, with narration) is attached in the comments. It uses a **fresh consumer app** — a new Vite project that installs the package from this branch and does exactly what the updated README says: one import for the component, one for the stylesheet, nothing copied out of the repo. It renders a real Gujarati chapter, then toggles only the packaged stylesheet off and on, so you can see precisely what the stylesheet contributes: superscript verse numbers, the chapter heading, spacing, and the toolbar icons (which travel inlined inside it).

https://github.com/user-attachments/assets/00b14767-c689-47f9-81d2-04a79ba96a41

## The problem

The package ships no importable CSS. `src/usj-nodes.css`, `src/editor/editor.css` and `libs/shared/src/styles/nodes-menu.css` are inside the tarball only because `files` includes `src`, but the `exports` map exposes only `.` and `./package.json`, so consumers can't reach them. `dist/index.css` (11.9 kB) contains just the comments-modal styles — they land there because `marginal/comments/*.tsx` do `import "./X.css"`. The 33 icon SVGs aren't published at all, and `editor.css` references them by absolute URL. That's why the README asks consumers to copy the CSS and assets out of the repo by hand — which is what we ended up doing in Fluent: 3,269 vendored lines that will silently drift.

## The change

A second, CSS-only Rollup input pulls the three stylesheets through the same Vite build and emits `dist/styles.css` (51.3 kB, 9.0 kB gzip), with the 15 toolbar icons inlined as data URIs. Consumers then write one line:

```ts
import "@eten-tech-foundation/platform-editor/styles.css";
```

`./comments.css` additionally exposes the existing `dist/index.css` for the deprecated `<Marginal />`.

## Nothing reachable through `"."` moves

This is the part that matters for Platform.Bible, and it's measured rather than argued. Against a pre-change build:

```
index.css   IDENTICAL   (11,869 B)
index.js    IDENTICAL
index.js.map IDENTICAL
index.d.ts  IDENTICAL
```

No CSS is auto-injected (`grep -cE 'index\.css|styles\.css' dist/index.js` → 0; `formats: ["es"]` structurally prevents it), and the source stylesheets are not edited by a single byte, so any vendored copies stay in sync. The new stylesheet is a file you choose to import — importing the component never pulls it in.

## Three things worth knowing before reviewing the config

- **`cssCodeSplit` must be on.** Vite defaults it to `false` whenever `build.lib` is set, which concatenates every entry's CSS into one asset — that alone would change `dist/index.css`. Safe here because the package has no dynamic imports, so each entry is a single chunk.
- **`lib.entry` stays a string** even though `rollupOptions.input` now drives entry resolution. `vite-plugin-dts` derives its entry list from `lib.entry`; making it an object flips it to multi-entry mode and api-extractor then fails with `Unable to determine module for: dist/styles.d.ts` on the export-less CSS entry. There's a comment saying so, because the duplication looks removable.
- **`rollupOptions.input` uses absolute paths.** Vite passes that through to Rollup verbatim (only the `lib.entry` fallback is resolved against `root`), so relative ids resolve against the cwd — it would work under `nx build` today and break under a plain `vite build --config`.

## Icons

Inlined, not shipped as files. Vite's lib mode inlines assets unconditionally, which is already why the four `CommentPlugin.css` icons are data URIs in today's `dist/index.css` — this just extends the same behaviour. Cost is ~2.3 kB gzip, paid also by `hasExternalUI: true` consumers who never render a toolbar; that seemed a fair trade against making everyone copy 144 kB of assets to their web root. The source `url()` paths are deliberately left absolute so the existing copy-to-web-root workflow keeps working. Bootstrap Icons' MIT license now travels with the package.

One consequence for the README: a host CSP whose `img-src` omits `data:` loses the icons silently, so that's called out.

## Verification

```bash
pnpm nx build @eten-tech-foundation/platform-editor
cmp <baseline>/index.css dist/index.css          # identical
grep -o 'data:image/svg' dist/styles.css | wc -l # 15
grep -c '/assets/'        dist/styles.css        # 0, nothing unresolved
grep -c 'annot_spellingerror'        dist/styles.css  # usj-nodes.css present
grep -c 'toolbar-item'               dist/styles.css  # editor.css present
grep -c 'autocomplete-menu-container' dist/styles.css # nodes-menu.css present
pnpm nx extract-api platform-editor              # API report unchanged, no styles.d.ts
npm pack --dry-run                               # ships dist/styles.css + icon LICENSE
```

A root-level `styles.css` shim (`@import "./dist/styles.css"`) is included so resolvers that ignore `exports` (webpack 4, older Metro/Jest) get a working path too.

Happy to adjust if this collides with the editor changes your team has planned — it's deliberately additive and confined to packaging.

- - -
This change is [Reviewable](https://reviewable.io/reviews/eten-tech-foundation/scripture-editors/528)

### DIFF
--- a/packages/platform/README.md
+++ b/packages/platform/README.md
@@ -140,17 +140,38 @@ export default function App() {
 
 ## Styling
 
-This npm package does not include styling so you need to style the editor component to suit your application. A good place to start is to copy the **CSS** from this repo:
+The package ships a bundled stylesheet covering the Scripture nodes, the editor chrome and the marker menu. Import it once, wherever your app imports CSS:
+
+```ts
+import "@eten-tech-foundation/platform-editor/styles.css";
+```
+
+The toolbar icons `editor.css` references are inlined into that stylesheet as data URIs, so **no asset copying is required**. Two things worth knowing:
+
+- Importing it from a **bundler-processed stylesheet** works too (`@import "@eten-tech-foundation/platform-editor/styles.css";`), but a browser loading a plain `<link>`ed file cannot resolve a bare package specifier.
+- If your host page sets a Content Security Policy, `img-src` must allow `data:` or the icons silently disappear — which looks like the stylesheet failed to load.
+
+If using the **commenting features** in the deprecated `<Marginal />` component, also import:
+
+```ts
+import "@eten-tech-foundation/platform-editor/comments.css";
+```
+
+Consumers doing a bare `tsc` (no bundler types) may need `declare module "*.css";` for the import to type-check; anything using `vite/client` types already has it.
+
+Icons are [Bootstrap Icons](https://icons.getbootstrap.com/) (MIT); the license travels with the package at `assets/images/icons/LICENSE.md`.
+
+### Overriding or forking the stylesheets
+
+To restyle beyond CSS overrides, copy the sources out of this repo instead:
 
 - Scripture Nodes [/packages/platform/src/usj-nodes.css](/packages/platform/src/usj-nodes.css)
 - Editor [/packages/platform/src/editor/editor.css](/packages/platform/src/editor/editor.css)
-- Marker Menu [/libs/shared/styles/nodes-menu.css](/libs/shared/styles/nodes-menu.css)
+- Marker Menu [/libs/shared/src/styles/nodes-menu.css](/libs/shared/src/styles/nodes-menu.css)
 
-For **icon assets** for the editor referenced in `editor.css` (the license file is included):
+`editor.css` references its icons from [/packages/platform/assets](/packages/platform/assets) by absolute URL, so a hand-copied stylesheet also needs those assets served from your web root. Swap the vendored imports for `styles.css` rather than stacking both, or the rules double up.
 
-- [/packages/platform/assets](/packages/platform/assets)
-
-If using the **commenting features** in the `<Marginal />` component:
+Comment styles, if using `<Marginal />`:
 
 - [/packages/platform/src/marginal/comments/ui/Button.css](/packages/platform/src/marginal/comments/ui/Button.css)
 - [/packages/platform/src/marginal/comments/ui/ContentEditable.css](/packages/platform/src/marginal/comments/ui/ContentEditable.css)
--- a/packages/platform/package.json
+++ b/packages/platform/package.json
@@ -21,13 +21,19 @@
       "types": "./dist/index.d.ts",
       "import": "./dist/index.js",
       "default": "./dist/index.js"
-    }
+    },
+    "./styles.css": "./dist/styles.css",
+    "./comments.css": "./dist/index.css"
   },
   "files": [
     "dist",
     "src",
     "!src/**/*.test.ts",
-    "!src/**/*.test.tsx"
+    "!src/**/*.test.tsx",
+    "!src/styles.css",
+    "!src/styles.ts",
+    "styles.css",
+    "assets/images/icons/LICENSE.md"
   ],
   "scripts": {
     "generate:test-data": "cross-env GENERATE_TEST_DATA=1 pnpm vitest run generate-2sa-lexical-states",
--- a/packages/platform/src/styles.css
+++ b/packages/platform/src/styles.css
@@ -1,0 +1,13 @@
+/*
+ * Bundled stylesheet for @eten-tech-foundation/platform-editor.
+ *
+ * Built to dist/styles.css and exposed to consumers as
+ * `@eten-tech-foundation/platform-editor/styles.css`. Icon SVGs referenced by
+ * editor.css are inlined as data URIs at build time, so no assets need copying.
+ *
+ * Import order is the documented cascade: Scripture nodes, then the editor
+ * chrome, then the marker menu.
+ */
+@import "./usj-nodes.css";
+@import "./editor/editor.css";
+@import "../../../libs/shared/src/styles/nodes-menu.css";
--- a/packages/platform/src/styles.ts
+++ b/packages/platform/src/styles.ts
@@ -1,0 +1,4 @@
+// CSS-only build entry: emits dist/styles.css. The companion dist/styles.js is
+// empty by design and is not part of the public API — consumers import the
+// stylesheet, never this module.
+import "./styles.css";
--- a/packages/platform/styles.css
+++ b/packages/platform/styles.css
@@ -1,0 +1,6 @@
+/*
+ * Stable path for resolvers that don't honour the "exports" map (webpack 4,
+ * older Metro/Jest). Bundlers that do honour it get dist/styles.css directly
+ * through the "./styles.css" subpath.
+ */
+@import "./dist/styles.css";
--- a/packages/platform/vite.config.mts
+++ b/packages/platform/vite.config.mts
@@ -34,19 +34,40 @@ export default defineConfig({
     emptyOutDir: true,
     sourcemap: true,
     reportCompressedSize: true,
+    // Emit one stylesheet per entry. Vite defaults this to false whenever
+    // build.lib is set, which would concatenate every entry's CSS into a single
+    // asset; keeping it per-entry is what leaves dist/index.css (the comment
+    // styles reachable through ".") byte-for-byte unchanged while the bundled
+    // editor stylesheet lands in dist/styles.css. Safe here because this package
+    // has no dynamic imports, so each entry is a single chunk, and formats:["es"]
+    // means Vite never emits style-injection code into the JS.
+    cssCodeSplit: true,
     commonjsOptions: {
       transformMixedEsModules: true,
     },
     lib: {
-      // Could also be a dictionary or array of multiple entry points.
+      // Entry resolution comes from rollupOptions.input below; this stays a
+      // STRING only to keep vite-plugin-dts in single-entry mode (it derives its
+      // entry list from lib.entry, and an object here makes api-extractor fail on
+      // the export-less CSS entry: "Unable to determine module for styles.d.ts").
+      // Don't "clean up" the apparent duplication.
       entry: "src/index.ts",
       name: "@eten-tech-foundation/platform-editor",
-      fileName: "index",
+      // Two inputs, so the name must be derived per entry — a fixed "index"
+      // makes Rollup dedup into index.js + index2.js.
+      fileName: (_format: string, entryName: string) => `${entryName}.js`,
       // Change this to the formats you want to support.
       // Don't forget to update your package.json as well.
       formats: ["es" as const],
     },
     rollupOptions: {
+      // Absolute paths are required: Vite passes rollupOptions.input to Rollup
+      // verbatim (only the lib.entry fallback is resolved against `root`), so
+      // relative ids would resolve against the current working directory.
+      input: {
+        index: path.resolve(__dirname, "src/index.ts"),
+        styles: path.resolve(__dirname, "src/styles.ts"),
+      },
       external: [
         "react/jsx-runtime",
         // Also externalize the dev JSX runtime so a dev-mode build can never bundle a


henrique221:
Reworked to match your review on #516.

`usj-nodes.css` is out of the bundle, and `editor.css` is split along the boundaries it already labelled: core, built-in toolbar (`hasExternalUI: false`), ContextMenuPlugin, and the debug-only TreeView. Only the core plus the marker menu land in `styles.css`.

The one judgement call is in the PR description: I kept the toolbar and context menu shipping as their own entries rather than dropping them, since `hasExternalUI` defaults to `false` so most consumers still need them. Folding them back into the bundle is a one-line change if you prefer that.

The split is rule-for-rule lossless (88 selectors before and after) and the repo's platform demo imports the new files, so `nx dev platform` is unaffected.

packages/platform/src/styles.css
```diff
@@ -0,0 +1,18 @@
+/*
+ * Bundled stylesheet for @eten-tech-foundation/platform-editor.
+ *
+ * Built to dist/styles.css and exposed to consumers as
+ * `@eten-tech-foundation/platform-editor/styles.css`. Icon SVGs referenced by
+ * editor.css are inlined as data URIs at build time, so no assets need copying.
```
irahopkinson: This is the wrong way round. After the split, `editor.css` has **zero** `url()` references — all 15 icon URLs are in `toolbar.css`, which this file deliberately excludes. `dist/styles.css` contains no data URIs at all.

Worth moving the sentence to `src/editor/toolbar.css`, where it's both true and useful.

packages/platform/src/styles.css
```diff
@@ -0,0 +1,18 @@
...
+ *
+ * Scope: what every editor needs, whatever its configuration. Three things are
+ * deliberately NOT here:
```
irahopkinson: Four things, not three — `debug-tree-view.css` is also absent and isn't mentioned. It's the one split file with no build entry and no `exports` subpath, so it's worth naming here explicitly so the omission reads as deliberate rather than forgotten. See my comment on `debug-tree-view.css`.

packages/platform/src/styles.css
```diff
@@ -0,0 +1,18 @@
...
+ *   - the built-in toolbar (`hasExternalUI: false`) and the context menu, which
+ *     ship as their own entries so a consumer supplying its own UI doesn't pay
+ *     for them: `platform-editor/toolbar.css`, `platform-editor/context-menu.css`;
+ *   - `usj-nodes.css`, which Platform generates per project — it has to respond to
+ *     project stylesheets, `custom.sty` markers and view modes, so a fixed copy in
+ *     the package would be wrong. Keep copying it (see README).
+ */
+@import "./editor/editor.css";
+@import "../../../libs/shared/src/styles/nodes-menu.css";
```
irahopkinson: **Question — why is this in the core bundle?**

The PR description says "only the core plus the marker menu land in `styles.css`", but not why the marker menu is core. As far as I can tell it's inherited from the README you're replacing, which listed three files to copy (`usj-nodes.css`, `editor.css`, `nodes-menu.css`) — so this rode along on the copy list rather than on any analysis of when it's needed.

What the code says: the only mount is `<UsjNodesMenuPlugin>` at `Editor.tsx:452-453`, gated on `scrRef && !hasExternalUI`. Its selectors (`.autocomplete-menu-container`, `.autocomplete-menu-options`, and the `.floating-box:has(> .autocomplete-menu-container)` rules) are used only by `NodeSelectionMenu.tsx`, reached only through that plugin. So it's gated on the *same* `!hasExternalUI` condition as the toolbar, plus a `scrRef` requirement — strictly more optional than the toolbar, and it's the one thing that ended up in core. About 1.2 KB of the 3.6 KB built `styles.css`, dead for any external-UI consumer.

By the boundary this PR sets, it belongs in its own entry alongside `toolbar.css`. Happy to hear the counter-argument if there is one.

Separately, minor: this reaches out of the package by relative path while TS in the same package uses the workspace alias (`ActiveTextPlugin.tsx:18` does `from "shared"`). Use the alias if the CSS resolver supports it, otherwise a one-line note saying why not.

packages/platform/vite.config.mts
```diff
@@ -34,19 +34,44 @@ export default defineConfig({
...
+      fileName: (_format: string, entryName: string) => `${entryName}.js`,
       // Change this to the formats you want to support.
       // Don't forget to update your package.json as well.
       formats: ["es" as const],
     },
     rollupOptions: {
+      // Absolute paths are required: Vite passes rollupOptions.input to Rollup
+      // verbatim (only the lib.entry fallback is resolved against `root`), so
+      // relative ids would resolve against the current working directory.
+      input: {
```
irahopkinson, line 67: On the open question from #516 — I'm happy with the toolbar and context menu shipping as their own entries rather than folding back into the bundle, so no change needed there.

For the context menu specifically: it currently renders unconditionally (`Editor.tsx:484`, and `options` is optional since `ContextMenuPlugin.tsx:135` supplies defaults), so today every consumer needs `context-menu.css`. I'm fine with it split out anyway, because it will eventually be gated on `hasExternalUI` like the toolbar. Flagging it so the split is a deliberate bet on that change rather than an accident — and it's worth a word in the README that this one is currently needed by everyone.

The entry that does look misplaced is `nodes-menu.css` — see my comment on `src/styles.css`.

packages/platform/package.json
```diff
@@ -21,13 +21,25 @@
       "types": "./dist/index.d.ts",
       "import": "./dist/index.js",
       "default": "./dist/index.js"
-    }
+    },
+    "./styles.css": "./dist/styles.css",
+    "./comments.css": "./dist/index.css",
```
irahopkinson: Two things here.

**The name doesn't describe the target.** This points at the `.` entry's auto-generated CSS, not a comments artifact. Any future CSS anywhere in the index graph would silently ship as "comments.css". Give comments its own `src/comments.ts` entry the way the toolbar has one, and the name becomes true.

**Question:** should we add a supported subpath for `<Marginal />` at all? It's deprecated and I'd take a PR removing it, so a public `./comments.css` buys us a breaking removal later. Nothing in #516 asked for it. I'd lean toward dropping this entry and leaving the comment styles on the copy list until Marginal's future is settled — say if that breaks something for you.

packages/platform/package.json
```diff
@@ -21,13 +21,25 @@
...
-    "!src/**/*.test.tsx"
+    "!src/**/*.test.tsx",
+    "!src/context-menu.ts",
+    "!src/context-menu.css",
+    "!src/toolbar.ts",
+    "!src/toolbar.css",
+    "!src/styles.ts",
+    "!src/styles.css",
+    "styles.css",
+    "assets/images/icons/LICENSE.md"
```
irahopkinson, line 53: Good catch shipping the license — but I think it now needs to be the license, not a pointer to it.

Until this PR the SVGs were never published. Now the artwork is redistributed inlined in `dist/toolbar.css` and `dist/index.css`, and `assets/images/icons/LICENSE.md` is 123 bytes: the project name and two URLs. MIT wants "the above copyright notice and this permission notice ... included in all copies", so I think we need the full Bootstrap Icons MIT text plus the upstream copyright line travelling with the package.

I'll check whether this routes past anyone on our side before it merges — flagging rather than asking you to guess at the right wording.

packages/platform/styles.css
```diff
@@ -0,0 +1,6 @@
+/*
+ * Stable path for resolvers that don't honour the "exports" map (webpack 4,
```
irahopkinson: The rationale is right, but it's applied to one subpath of four. A resolver that ignores `exports` also can't reach `toolbar.css`, `context-menu.css` or `comments.css` — so exactly the consumers this file exists for get core styles and an unresolvable toolbar, which is the default configuration.

Either add the other three root shims or drop this one and note the limitation in the README.

packages/platform/styles.css
```diff
@@ -0,0 +1,6 @@
...
+ * older Metro/Jest). Bundlers that do honour it get dist/styles.css directly
+ * through the "./styles.css" subpath.
+ */
+@import "./dist/styles.css";
```
irahopkinson, line 9: Minor: `dist/` is gitignored, so in a fresh clone this committed file points at nothing — it only resolves inside a published tarball. Worth a clause in the comment above saying so, otherwise it reads as broken to anyone who opens it in the repo.

packages/platform/src/toolbar.css
```diff
@@ -0,0 +1 @@
+@import "./editor/toolbar.css";
```
irahopkinson: This file (and `src/context-menu.css`) is a single `@import` of its `editor/` sibling. `src/toolbar.ts` could `import "./editor/toolbar.css"` directly, which removes two files and two `files` negations, and takes the per-stylesheet edit count from six to four.

Is there a resolver reason for the indirection I'm missing? If not, I'd drop both shims.

packages/platform/src/styles.ts
```diff
@@ -0,0 +1,4 @@
+// CSS-only build entry: emits dist/styles.css. The companion dist/styles.js is
+// empty by design and is not part of the public API — consumers import the
```
irahopkinson: The comment is honest, but the three empty `dist/{styles,toolbar,context-menu}.js` files plus their `.js.map` siblings still get published — six files that are in the tarball, not in `exports`, and do nothing. Add `"!dist/styles.js"` and friends to `files`.

packages/platform/README.md
```diff
@@ -140,17 +140,58 @@ export default function App() {
...
+If you use the **built-in UI** (`hasExternalUI` is `false`, the default), also import the toolbar, and the context menu if you use it:
+
+```ts
+import "@eten-tech-foundation/platform-editor/toolbar.css";
+import "@eten-tech-foundation/platform-editor/context-menu.css";
+```
+
+They are separate entries so an app supplying its own toolbar or context menu doesn't ship styles it will override anyway.
+
+The toolbar icons are inlined as data URIs, so **no asset copying is required**. Two things worth knowing:
```
irahopkinson: This paragraph sits under the `styles.css` import block, but `styles.css` has no icons in it — the data URIs and the `img-src data:` concern both belong to `toolbar.css` (and `index.css`). As written it reads as contradicting line 192, which says toolbar icons are referenced by absolute URL. Both statements are true — built vs source — but a reader hits them as a contradiction.

Scope this paragraph to the toolbar/context-menu imports above it.

packages/platform/src/editor/debug-tree-view.css
```diff
@@ -0,0 +1,67 @@
+/* TreeView styles, only needed when the editor is created with `debug=true`.
```
irahopkinson, line 1: I only asked for the two labelled sections, so this is a third split I didn't request — but I want to keep it, for a better reason than "debug only".

This file carries two **unscoped** element selectors:

```css
pre::-webkit-scrollbar
pre::-webkit-scrollbar-thumb
```

Those hit every `<pre>` on the host page, not just the tree view. Had they stayed in `editor.css`, the new bundled `styles.css` would silently restyle scrollbars on every consumer's code blocks the moment they import it. The split prevents a real regression — worth saying so in the header comment, because "TreeView, debug only" undersells it.

What's missing is a consumption path: `debug` is a public prop (`editor.model.ts:260`, `Editor.tsx:498`), so someone can turn it on and get an unstyled tree view. Rather than add a fourth public subpath for an experimental dev-only flag, let's give this the same treatment as `usj-nodes.css`: source-only, on the copy list (it already is, at README:190), and named in the `src/styles.css` header so the omission is visibly deliberate. No code change beyond the two comments.

demos/platform/src/styles.css
```diff
@@ -6,6 +6,11 @@
 @import "../../../packages/platform/src/marginal/comments/comment-editor.theme";
 @import "../../../packages/platform/src/marginal/comments/CommentPlugin";
 @import "../../../packages/platform/src/editor/editor";
+/* Split out of editor.css (#516): the built-in toolbar, the context menu and the debug TreeView
+   are their own entries so consumers with external UI don't ship them. The demo uses all three. */
+@import "../../../packages/platform/src/editor/toolbar";
```
irahopkinson, line 14: The demo re-enumerates `editor` + `nodes-menu` rather than importing `packages/platform/src/styles.css`, so nothing in the repo actually exercises the composition we ship. If the demo imported `src/styles.css` plus the optional files, `nx dev platform` would become a regression test for the bundle instead of a parallel arrangement that can drift from it.

packages/platform/src/styles.css
henrique221: Moved it to `src/editor/toolbar.css` in c528e98, where it's actually true, along with why it works: Vite resolves the root-absolute urls against `config.root`, and the icons are 258-720 B so they all land under the 4096 B inline limit.

packages/platform/src/styles.css
henrique221: Fixed in c528e98. It's five now, since the marker menu moved out too, and `debug-tree-view.css` is named explicitly.

packages/platform/src/styles.css
henrique221: No counter-argument, you're right. It's its own entry now, `platform-editor/nodes-menu.css` (d2216d0): built `styles.css` goes from 3.63 kB to 2.46 kB, `nodes-menu.css` is 1.17 kB, and the 35 rule blocks are conserved exactly across the two.

On the alias, it didn't resolve. `shared` exported only `.` and `./package.json`, so `shared/src/styles/nodes-menu.css` failed with `Missing "./src/styles/nodes-menu.css" specifier in "shared" package`. And now that the import sits in TS rather than CSS, a relative path out of the project is an `@nx/enforce-module-boundaries` error, so a note wasn't an option either. I added `"./styles/*"` to `shared`'s exports (it's `private: true`, so nothing public widens) and used the alias.

packages/platform/vite.config.mts
henrique221: Good, thanks. Added the README note in c528e98 saying everyone needs `context-menu.css` today because `ContextMenuPlugin` renders unconditionally, and that the split is a deliberate bet on it getting gated later rather than an accident. `nodes-menu.css` moved out, see that thread.

packages/platform/package.json
henrique221: Dropped it in d2216d0. Nothing on my side uses `<Marginal />`, so there's no reason to buy a breaking removal later. The comment styles stay on the README copy list.

packages/platform/package.json
henrique221: Agreed. 9b81a77 replaces the pointer with the verbatim upstream text from twbs/icons.

One thing for your check: the 4 icons in `dist/index.css` come from `CommentPlugin.css`, which this PR doesn't touch, so they already ship inlined on main today. The redistribution predates this PR rather than starting with it, which probably makes the fix more urgent, not less. Leaving this open for you.

packages/platform/styles.css
henrique221: Added the other three in d2216d0, one per published stylesheet, so an exports-blind resolver reaches all of them and not just the core one.

packages/platform/styles.css
henrique221: Added, in all four shims (d2216d0).

packages/platform/src/toolbar.css
henrique221: No resolver reason, dropped both in d2216d0. The `.ts` entries import their `editor/` sibling directly and the built `toolbar.css` and `context-menu.css` come out byte-identical (`cmp`).

I kept `src/styles.css` though. With the marker menu gone it's down to a single `@import` like the other two were, but it's the one place that says what the bundle contains, and it's what the demo imports now for your other comment. Happy to inline it into `styles.ts` if you'd rather have the consistency.

packages/platform/src/styles.ts
henrique221: Done in d2216d0, eight negations counting the `.js.map` siblings. `npm pack --dry-run` on a clean build is 78 files with none of them in it.

packages/platform/README.md
henrique221: Scoped it to the toolbar in c528e98. It also now says `styles.css` references no images at all, so the two statements stop reading as a contradiction.

packages/platform/src/editor/debug-tree-view.css
henrique221: Hadn't thought about the `pre` selectors, only about it being debug-only. The header says so now, and explains why this is the one stylesheet with no build entry and no subpath. `src/styles.css` names it too, so the omission reads as deliberate (c528e98).

demos/platform/src/styles.css
henrique221: Done in 0b24ad3, the demo imports `src/styles.css` plus the optional files. The built demo stylesheet has the same 760 rule blocks, none added, none missing. The marker menu moves earlier in the cascade, but its 9 selectors appear in neither of the files it now precedes, so nothing resolves differently.

henrique221:
The two items in your review body that had no line to anchor to:

**Orphaned stylelint directive.** Moved in c528e98. `editor.css` has no qualifying-type selectors left at all now, and the directive sits on `toolbar.css` and `context-menu.css`, which do.

**Six edits per stylesheet.** This did not drop to four, and I want to flag that rather than let you find it. Dropping the `src/*.css` shims took one edit off, but adding the three missing root shims put one back, so it is still six: `demos/platform/src/styles.css`, `src/<n>.ts`, the `vite.config.mts` input, `exports`, `files`, and the root `<n>.css`. Agreed it is not this PR's problem, but the arithmetic did not work out the way we both expected.

On your `nodes-menu.css` question I had no counter-argument, so it moved to its own entry. That changed the shape of what ships, so the description is updated. Numbers I re-derived rather than reused: built `styles.css` 3.63 kB to 2.46 kB, `nodes-menu.css` 1.17 kB, 35 rule blocks conserved across the two; toolbar and context-menu output byte-identical after the shims went; the demo's built stylesheet holds the same 760 rule blocks; `npm pack --dry-run` on a clean build is 78 files with the eight empty JS chunks gone. And you were right that my "88 selectors" figure did not reproduce, sorry, I have dropped it.

packages/platform/src/styles.css
irahopkinson: Confirmed at 777cf8ed: the `editor.css` header is clean and the explanation now sits in `toolbar.css` with the mechanism spelled out, which is more than I asked for and the right place for it. Your 258–720 B figure is exact for the 15 referenced icons — I re-measured (`type-h1.svg` 258, `3-square.svg` 720).

packages/platform/src/styles.css
irahopkinson: Reads right now — five, with `debug-tree-view.css` named and its reason given.

packages/platform/src/styles.css
irahopkinson: Thanks for taking it. Verified on a clean build: `dist/nodes-menu.css` is 1168 B (your 1.17 kB, exact) and `dist/styles.css` drops to 2488 B. Mine reads 2.49 kB rather than your 2.46 — that is the main merge touching `editor.css` after you measured, not a discrepancy in your work.

One note on the block count, since you asked me to check the numbers. I make it **34** conserved, not 35: 25 in `editor.css` plus 9 in `nodes-menu.css`. The difference is counting convention — `@media (prefers-reduced-motion: reduce)` is one block the way I count and two the way you presumably did. Conservation is exact either way, so there is nothing to fix here; I mention it only because it is the second counting figure I have queried and this one holds up.

The `shared` exports addition is the right call, and I checked it is `private: true`, so nothing public widens.

packages/platform/vite.config.mts
irahopkinson: Good — the README note lands the point that everyone needs `context-menu.css` today and why it is still split. No further change wanted here.

packages/platform/package.json
irahopkinson: Agreed, and dropping it entirely is the better outcome. Worth noting for the record that `dist/index.css` still ships inside the tarball — it is under `files: ["dist"]` — it just is not reachable through `exports` any more. That is fine, and the README copy list covers the `<Marginal />` case.

packages/platform/package.json
irahopkinson: You are right and my premise was wrong — thank you for pushing back rather than just doing what I asked.

I checked it: `CommentPlugin.css` is byte-identical on `origin/main` and untouched by this PR, and `dist/` is published wholesale under `files`, so `dist/index.css` with those 4 inlined icons already ships on npm today. Bootstrap Icons artwork has been redistributed for some releases already. My comment said "until this PR the SVGs were never published" — true of the `.svg` files, wrong about the artwork, which is the thing the licence actually attaches to. So this PR closes a pre-existing gap rather than opening one, which as you say makes it more urgent, not less.

`LICENSE.md` now carries the verbatim upstream text and the `Copyright (c) 2019-2024 The Bootstrap Authors` line, which is what was missing. I am picking up whether this needs a sign-off on our side before merge; that is on me, not a change request for you.

packages/platform/styles.css
irahopkinson: All four are there and I confirmed them in `npm pack --dry-run`: `styles.css`, `toolbar.css`, `nodes-menu.css`, `context-menu.css` at the package root, each pointing at its own `dist/` file.

packages/platform/styles.css
irahopkinson: Added in all four, thanks.

packages/platform/src/toolbar.css
irahopkinson: Both shims gone, and I confirmed the built `toolbar.css` (11804 B) and `context-menu.css` (1336 B) are unchanged to the byte from before the change — as expected, since you only added header comments and those minify away.

On your question: **keep `src/styles.css`.** The Middle Man objection applied to the two shims because each had exactly one consumer. This one has two — the `styles.ts` build entry and the demo — and inlining it into `styles.ts` would leave the demo importing a `.ts` file to get CSS. It is also the one place that states what the bundle contains, which is worth a file on its own.

packages/platform/src/styles.ts
irahopkinson: Verified: a clean build plus `npm pack --dry-run` shows none of `dist/{styles,toolbar,nodes-menu,context-menu}.js` or their `.map` siblings in the tarball. I get 79 files rather than your 78; the extra one arrives with the main merge, not from you.

packages/platform/README.md
irahopkinson: Scoped correctly now, and adding "`styles.css`, `context-menu.css` and `nodes-menu.css` reference no images at all" removes the apparent contradiction with the forking section rather than just moving it.

packages/platform/src/editor/debug-tree-view.css
irahopkinson: This is the shape I wanted. The header, `src/styles.css` and the README entry now each say why it is source-only, and the README spelling out the unscoped `pre` leak is the part a consumer actually needs. Nothing further.

demos/platform/src/styles.css
irahopkinson: Verified — the demo pulls `src/styles.css` itself, so `nx dev platform` now renders the composition we publish rather than a parallel one. The reorder is safe: the marker menu's 9 selectors collide with nothing in the files it now precedes.

packages/platform/README.md
```diff
@@ -140,17 +140,62 @@ export default function App() {
...
+- Editor [/packages/platform/src/editor/editor.css](/packages/platform/src/editor/editor.css)
+- Built-in toolbar [/packages/platform/src/editor/toolbar.css](/packages/platform/src/editor/toolbar.css)
+- Context menu [/packages/platform/src/editor/context-menu.css](/packages/platform/src/editor/context-menu.css)
+- Marker Menu [/libs/shared/src/styles/nodes-menu.css](/libs/shared/src/styles/nodes-menu.css)
+- TreeView, `debug` only [/packages/platform/src/editor/debug-tree-view.css](/packages/platform/src/editor/debug-tree-view.css) — the only stylesheet with no published subpath at all, because `debug` is an experimental dev-only prop and the file carries unscoped `pre` selectors that would restyle every `<pre>` on your page. Copy it if you turn `debug` on.
 
-- [/packages/platform/assets](/packages/platform/assets)
+`toolbar.css` references its icons from [/packages/platform/assets](/packages/platform/assets) by absolute URL, so a hand-copied stylesheet also needs those assets served from your web root. Swap the vendored imports for `styles.css` rather than stacking both, or the rules double up.
```
irahopkinson: Nit, last one: "Swap the vendored imports for `styles.css`" now under-specifies the set. A consumer forking the stylesheets needs `styles.css` **plus** `toolbar.css`, `context-menu.css` and `nodes-menu.css` to replace what they were vendoring — swapping in `styles.css` alone would silently drop the toolbar, context menu and marker menu.

irahopkinson (commented):
Re-verified the whole thing at 777cf8ed in a clean worktree — fresh install, fresh build, `npm pack --dry-run`, lint, typecheck and prettier. Since the branch has merged main in the meantime I diffed only your four commits (`c8d54f00..9b81a778`) so the review surface stayed the 21 files you actually touched.

Everything I could re-derive holds:

| claim | result |
|---|---|
| `nodes-menu.css` its own entry | `dist/nodes-menu.css` 1168 B — your 1.17 kB exact |
| built `styles.css` shrinks | 2488 B (you said 2.46 kB; mine reads 2.49 — main's merge touched `editor.css` after you measured) |
| toolbar / context-menu unchanged | 11804 B and 1336 B, identical to the byte |
| icons still inline | 15/15 `data:` URIs in `toolbar.css`; zero external `url()` in any built stylesheet |
| the 15 icons are 258–720 B | exact |
| empty JS chunks gone | zero in the tarball |
| `exports` subpaths | all four resolve to emitted files |
| lint / typecheck / prettier | green (one `no-loop-func` warning in `verse-block.utils.ts`, from main, not from you) |

Two figures differ from yours and neither is a problem: 79 packed files rather than 78, and 34 conserved rule blocks rather than 35. Both are explained in the threads — the first is the main merge, the second is a counting convention around `@media`.

Three answers you were waiting on, all in the threads: **keep `src/styles.css`** (it has two consumers now, unlike the shims); **you were right about the licensing premise and I was wrong** — the artwork already ships on main via `dist/index.css`, so this fixes a pre-existing gap rather than opening one; and thanks for flagging the six-edits arithmetic yourself rather than letting me find it — you are right that adding the three root shims put back what dropping the `src/*.css` shims took off.

One nit inline below, and that is the last thing from me. I will take a final pass in Reviewable.

irahopkinson (commented):
> @irahopkinson+AI resolved 1 discussion.
> *[Reviewable](https://reviewable.io/reviews/eten-tech-foundation/scripture-editors/528)* status: 0 of 21 files reviewed, 13 unresolved discussions (waiting on henrique221 and irahopkinson).

packages/platform/README.md
henrique221: listed all four

kaseywright (approved):

kaseywright (approved):
> @kaseywright reviewed 21 files and all commit messages.
> *[Reviewable](https://reviewable.io/reviews/eten-tech-foundation/scripture-editors/528)* status: all files reviewed, 13 unresolved discussions (waiting on henrique221, irahopkinson, and irahopkinson+AI).

kaseywright (approved):
> *[Reviewable](https://reviewable.io/reviews/eten-tech-foundation/scripture-editors/528)* status: all files reviewed, 13 unresolved discussions (waiting on henrique221, irahopkinson, and irahopkinson+AI).

henrique221:
Replaced by #560 with the same commits and changes. Review is requested from Kasey Wright on the new PR.

Issue #26, opened 2024-02-12
`ImmutableNoteCallerNode` convert to vanilla JS


irahopkinson:
Can we convert `ImmutableNoteCallerNode` to vanilla JS and it still work in React?

**Background:** `ImmutableNoteCallerNode` is a decorator node simply decorated with an HTML anchor element with the `onClick` function as a passed-in property. It was pretty simple to implement it as it is currently as a TSX node.

abelpz:
I'll take a look

irahopkinson:
This is not worth doing. If we need a plain vanilla JS version just copy this React node and convert it to vanilla TS.

Issue #18, opened 2024-02-06
Move shared plugin and node to `shared-react`


irahopkinson:
- git ignore NX cache
- also fix NoteNode onClick
- change NoteNode to JSX
- fix adapter onClick
- move NoteNode to `shared-react`
- add View Mode to adaptor

Note that the `EditorAdaptor.initialize()` and `EditorAdaptor.reset()` methods are optional - you don't have to define them if they aren't used. **UpdateStatePlugin** will only call them if they are defined.

irahopkinson:
**Note:** recreates Pull Request history from previous repository.

Issue #10, opened 2024-02-06
Platform: convert USJ to USX


irahopkinson:
Convert USJ back to USX. See [App.tsx](../blob/main/packages/platform/src/App.tsx#L10) for how we currently go in the other direction.

This requires #9 and #37 in order to complete saving files.

Remember to update the mermaid diagram in the [README](../blob/main/packages/platform/README.md).

PR #529, opened 2026-08-06
Add character marker removal to the editor ref

jolierabideau:
# Code Review Summary

**Branch**: character-marker-remove

**Base**: origin/main

**Date**: 2026-08-06

**Review model**: Claude Opus 5 (1M context)

**Files changed**: 11

## Overview

Adds a new `removeCharacterMarker(marker?)` method to `EditorRef`. It removes a USFM character
marker (e.g. `\nd`, `\wj`) from the current editor selection while preserving every character of the
text inside it — deleting the marker, never the content. This is the editor-side half of giving users
a discoverable way to add, change, and remove character markers; the UI that will call it lives in
paranext-core.

Removal is the only operation added here. Querying which marker is at the cursor needs no new API,
because `StateChangeSnapshot.contextMarker` already reports it and paranext-core already reads it.

The implementation composes helpers that already existed in `usj-marker-action.utils.ts` rather than
adding new algorithms or a new module: it calls `$unwrapNode` from `@lexical/utils` directly, leaves
`$unwrapTypedMarkNode` and every annotation and comment path untouched, and adds no new package-level
exports. It handles a selection that covers only part of a marked run by splitting the `CharNode`
around the selection, strips the marker's own visible representation and its NBSP prefix in the views
that display markers, skips selections inside a footnote, and restores the user's selection
afterwards — including collapsed carets and backward selections.

The review made seven substantive changes. The most significant is a **behavioral change**: when the
selection covers only part of a *nested* marked run, removing the outer marker is now refused instead
of removing that marker from the whole nested run. See Interview Notes for the reasoning — this is
the one item that rests on an unresolved product question.

## API Changes

- `@eten-tech-foundation/platform-editor` (`EditorRef`): added
  `removeCharacterMarker(marker?: string): void`. Purely additive — no existing member's signature,
  documentation, or behavior changed. Implemented at `Editor.tsx:316` and mirrored in the
  `Marginal.tsx:186` pass-through (mandatory, since `MarginalRef extends EditorRef`). `Editorial`
  forwards its ref straight to `Editor`, so it needed no change.
- `packages/platform/etc/platform-editor.api.md`: regenerated with the single new
  `removeCharacterMarker(marker?: string): void` line, alphabetized between `removeAnnotation` and
  `replaceEmbedUpdate`. Confirmed generated rather than hand-edited — `nx extract-api` produces an
  empty diff.
- Internal, not public API: `$removeCharacterMarkerAtSelection` and the new
  `isCharacterMarkerSupported` are exported from `usj-marker-action.utils.ts`, which is not
  re-exported from `packages/platform/src/index.ts`; neither appears in the API report.
- `libs/shared`: `$isParaMarkerPrefix` renamed to `$isSynthesizedMarkerNode` (2 call sites).
  `libs/shared` has no API report, so this is source-only and not a published-surface change.
- No new or changed exported types. `packages/scribe` has its own unrelated `EditorRef` and is
  unaffected.

Downstream cross-check in paranext-core (branch `character-marker-control`):
`use-character-marker-state.hook.ts:33` declares the consumer seam as
`removeCharacterMarker?: (marker: string) => void`. The name here now matches it exactly.

- [x] **`$splitCharNodeAroundTargets` silently removed the outer marker from a nested marked run's
      entire span when the selection covered only part of it** — changing the saved USJ for text the
      user never selected. It was documented and pinned by a "known limitation" test, but a caller had
      no way to detect it. _(fixed during review: `$splitCharNodeAroundTargets` now returns
      `CharNode | undefined` and returns `undefined` when a nested element child that the selection
      reaches into is not fully covered. The check runs during the read-only coverage scan, before any
      mutation, so the document is left untouched. Added an `$isFullyCoveredByTargets` helper;
      synthesized marker children are excluded from the coverage requirement. Public TSDoc reframed
      around the guarantee that removal never alters unselected text, and the pinning test flipped
      from asserting the marker was dropped to asserting nothing changed. Removing either marker of a
      fully covered nested pair is unaffected.)_

Author response: Author asked for a reasoned conclusion drawn from the product requirements and the
implementation plan rather than accepting or dismissing the finding, then approved the recommendation
to refuse. Full reasoning is in Interview Notes; the decisive point is that the product decision was
to act on the innermost marker only, which makes this code path unreachable from the shipping UI.

- [x] **A call that matched no marker still mutated the document before knowing there was anything
      to remove** — the loop ran `splitText` over every selected text node before any matching
      `CharNode` was resolved, contradicting the documented "Does nothing, without throwing."
      _(fixed during review: added a read-only `$hasCharNodeToRemove` pre-pass with an early return,
      and extracted `$getTargetNode`'s skip rule into a shared `$isSkippedByMarkerAction` so the
      pre-pass cannot drift from the loop it guards. The pre-pass deliberately omits
      `handleTextNode`'s zero-width filter, so it can only ever be more permissive — it can never
      wrongly refuse a removal that would otherwise have happened.)_
- [x] **`$createCharNodeLike` hand-rolled Lexical's `$copyNode`** — five manual property copies plus
      a node-state copy. _(fixed during review: replaced with `$copyNode`. Verified against the
      Lexical 0.43.0 source rather than taken on trust: `$copyNode` assigns a fresh key before
      `afterCloneFrom`, whose children copy is gated on matching keys, so the copy is childless; it
      unconditionally copies indent, format, style, direction and both text-style members; it carries
      node state via `getWritable`; and `resetOnCopyNode()` only clears state configs that opt in,
      which `charIdState` does not — so the char id survives. `CharNode.clone` carries the marker and
      unknown attributes. `$copyNode` skips `$applyNodeReplacement`, which is harmless here: the only
      registered replacement is `ParagraphNode → ImpliedParaNode`. Four imports became unused and
      were dropped. The pre-existing char-id/unknown-attributes identity test still passes, so this is
      confirmed by a falsifiable test and not only by reading source.)_
- [x] **A local predicate was a byte-identical fork of an exported shared one** — the new
      `$isSynthesizedMarkerNode` duplicated `$isParaMarkerPrefix` from `libs/shared` across a package
      boundary. _(fixed during review: renamed the shared function to `$isSynthesizedMarkerNode` —
      its docstring was already generic, only the name said "para" — generalized that docstring to
      cover character markers, deleted the local fork, and updated both call sites.
      `ParaMarkerPrefixGuardPlugin` keeps its name, since that concept genuinely is about
      paragraphs.)_
- [x] **Marker validation lived in the imperative handle and diverged from `insertMarker` on
      `extraValidMarkers`** — removal validated inline with `CharNode.isValidMarker(marker,
      extraValidMarkers)` while `insertMarker` delegates to `isUsjMarkerSupported`, which ignores the
      extras. Net effect: a project-supplied extra marker was removable but not insertable.
      _(fixed during review: added an exported `isCharacterMarkerSupported(marker,
      extraValidMarkers?)` beside `isUsjMarkerSupported`, documenting both deliberate divergences —
      stricter, because removal only ever targets a `CharNode`, and extras-aware. `Editor.tsx` now
      calls it and no longer imports `CharNode` at all. Nine new tests, including one that pins the
      asymmetry explicitly. No public API change.)_
- [ ] ~~The collaborative-editing path at `delta-apply-update.utils.ts:398,406` calls bare
      `$unwrapNode` on a `CharNode` with no marker stripping and no NBSP trim — the exact gap
      `$removeCharNodeKeepingContent` exists to close. Its placement in `packages/platform` means
      `libs/shared-react` cannot share it.~~ _(Pre-existing defect, neither introduced by nor
      reachable from this branch's code. Author agreed to record it for the reviewer rather than widen
      scope. Logged in Suggested Review Focus.)_

Author response: Author requested fixes for all four in-scope items and agreed the
collaborative-editing finding should be recorded as pre-existing rather than pulled into this branch.

### Minor — Consider

- [x] **Naming: the method was `removeCharMarker`**, which contradicted the project's preferred
      terminology ("character marker" over "char marker") and the seam paranext-core had already
      declared as `removeCharacterMarker`. Free to change now; breaking once paranext-core consumes
      it. _(fixed during review: renamed to `removeCharacterMarker` across `editor.model.ts`,
      `Editor.tsx`, `Marginal.tsx`, tests, and the regenerated API report;
      `$removeCharMarkerAtSelection` → `$removeCharacterMarkerAtSelection`; error messages and the
      test helper and describe block renamed to match. Pre-existing "char marker" strings in
      `usj-editor.adaptor.ts` and the scribe adaptors were deliberately left alone, to keep the diff
      to this change.)_
- [x] **Dead optional chain** at `Editor.tsx:318` — `nodeOptions?.extraValidMarkers`, where
      `nodeOptions` is `useMemo(() => nodes ?? defaultNodeOptions, [nodes])` and is never
      `undefined`; every other use in the file passes it without `?.`. _(fixed during review: folded
      into the `isCharacterMarkerSupported` change above.)_
- [x] **The exported entry point sat inside the private-helpers `#region`**, whose header had been
      widened to accommodate it. _(fixed during review, but NOT by hoisting the export as suggested —
      that would have broken the file's actual convention, which is "primary function immediately
      above its own helpers" (`getUsjMarkerAction` → `getMarkerAction`;
      `$wrapTextSelectionInInlineNode` → its helper region). Instead the wrap-helpers region is now
      closed before `$removeCharacterMarkerAtSelection`, and a second region opened for the removal
      helpers, so no exported function sits inside a block labelled "Helper functions". The widened
      first-region label was kept because it is now literally accurate — four of its helpers are
      genuinely shared by both the wrap and the unwrap paths.)_
- [x] **Over-long comment blocks** — a 12-line note justifying a guard the note itself argued was
      currently unreachable, and a ~24-line selection-restore note above a 9-line block. _(fixed
      during review: trimmed to 3 and 14 lines. The selection-restore note was restructured into
      three named traps — the `setTextContent` offset drift, capturing selection direction before
      mutation, and the stale-selection-object trap — each mapping to a line that looks removable but
      isn't. No documented behavior was dropped; all of it remains asserted by tests.)_
- [x] **Eight module-scope mutable test fixtures** reassigned across roughly 18 tests. _(fixed during
      review: the five used by a single test each were converted to locals, taking module scope from
      nine declarations down to four. The genuinely multi-test fixtures were kept. **Caveat flagged
      to the author:** this required introducing `let x!: TextNode;`, a definite-assignment assertion
      that appears nowhere else in this repo's tests — a plain local fails with `TS2454`, because the
      assignment happens inside a callback TypeScript cannot see through. Most of the benefit is
      readability rather than safety: the cross-test leakage hazard was already near-zero for
      single-use variables.)_
- [x] **No end-to-end test of the `Editor.tsx` wiring**, specifically that it forwards its view
      options as the third argument — a wrong or omitted argument would leave every adaptor-level test
      green while breaking marker stripping in the views that display markers. _(fixed during review:
      added `describe("removeCharacterMarker through the editor ref")` in `Editor.test.tsx`, driving
      the public ref end to end with markers displayed and asserting the text reads back with no
      leftover NBSP. Verified falsifiable: stubbing the view options to `undefined` makes it fail on
      the leftover NBSP while all 187 adaptor tests stay green. It also asserts a precondition that
      the NBSP was actually present, so it cannot pass vacuously.)_
- [x] **The README `EditorRef` table had no row for the new method.** _(fixed during review: added
      between `insertMarker` and `insertNote`, matching the interface's own grouping.)_
- [ ] ~~The `Marginal.tsx` pass-through is untested — typecheck proves the member exists, not that it
      delegates.~~ _(Pre-existing gap: no test file exists anywhere under
      `packages/platform/src/marginal/`, and `insertMarker`'s pass-through is equally untested. Author
      agreed that creating the directory's first-ever test file to assert one delegation is
      disproportionate, and that it is better logged for the reviewer.)_

Author response: Author requested fixes for every minor finding except the two recorded as
pre-existing, and asked that the region-placement item follow the file's own conventions rather than
the analysis agent's suggestion — which changed the fix that was made.

#### Shared Regions Modified

None. No `#region shared with` markers exist in any changed file.

#### Extension Config Changes

None. This repo has no `extensions/` directory, so extension-config propagation does not apply.

## Positive Observations

- The API change is strictly additive by construction, verifiable straight from the diff:
  `insertMarker`, `formatPara`, `applyUpdate`, `$unwrapTypedMarkNode`, and every annotation and
  comment path are untouched. That was an explicit acceptance criterion, since one editor package
  serves more than one product surface.
- The three binding design decisions made up front were all honored and are confirmable from the
  diff: call `$unwrapNode` directly rather than generalizing `$unwrapTypedMarkNode`; put the code in
  the existing `usj-marker-action.utils.ts` rather than a new `libs/shared-react` module; add zero
  new package-level exports.
- Test coverage substantially exceeds the required list: every required case is present and asserts
  content preservation, plus leading-only and trailing-only clone splits, identity copying,
  backward-selection direction preservation, empty-marker-placeholder handling, and both
  marker-display modes.
- The remaining documented limitation is pinned by a test asserting today's actual output with an
  explanatory comment, so a future fix trips the test rather than silently changing behavior.
- The marker-display test fixtures were built to mirror `usj-editor.adaptor.ts`'s `createChar` rather
  than guessed at, each with a comment tying it back to that function.
- Caret and selection restoration — the main user-visible risk in an operation like this — is handled
  deliberately: the collapsed branch compensates for the NBSP trim and clamps to range, the range
  branch preserves selection direction instead of normalizing to forward, and the selection is
  re-fetched to avoid a `TextNode.replace()` stale-selection trap. Restoring the range over the same
  characters means a toolbar caller can re-toggle without the user having to re-select.
- A real pre-existing bug was spotted and deliberately not copied: `CharNode.insertNewAfter` and
  `ParaNode.insertNewAfter` both pair `setStyle` with `getTextStyle()` — two different members —
  making their style copies silently inert. The comment recording this survives in
  `$createCharNodeLike`.
- Selections inside a footnote are skipped rather than partially handled, avoiding a confusing
  half-applied edit inside note content.
- No localization work needed: `packages/platform/src` has no i18n machinery and the only new strings
  are developer-facing `Error` messages. No new React components, so no Storybook stories apply. No
  build config, dependency, or workflow changes.

## Interview Notes

**Stated purpose**: add the editor-side character-marker removal capability described in the
character-level markers implementation plan, cross-referenced against paranext-core as the consumer.

**The author did not defer to AI on any finding, and did not express uncertainty about their own
code.** On the one genuinely ambiguous item — the nested-marker Critical — they explicitly declined to
either accept or dismiss it, and instead asked for a conclusion derived from the product requirements
and the implementation plan. That is the right instinct for a question the plan itself flags as
needing product sign-off, and is noted here as a positive rather than a gap.

**The nested-marker decision, and why it is the one item a reviewer should scrutinize.** The
recommendation to refuse rather than over-remove rests on four points, three of them independently
verifiable:

1. Nesting was deliberately taken out of scope. The product decision was to act on the **innermost
   marker only**; the richer "show every marker touching the selection" option — the one that would
   make nested markers individually addressable — was closed rather than deferred. Recursive
   splitting is exactly the work that decision excluded, and the appetite for this epic fixes time
   and flexes scope.
2. Shipping the over-removal was not sanctioned either. The plan's test list covers nested removal
   only for *fully covered* markers. It separately records that semantics for nested and
   partially-covered markers are still undecided and need a product decision, and names USFM
   semantics as an explicit stop-and-ask area requiring product sign-off rather than an engineering
   guess.
3. **Decisive, and verified in paranext-core rather than assumed**: the affected path is unreachable
   from the shipping UI. `resolveCurrentMarker` returns `undefined` whenever two or more markers cover
   the selection (`use-character-marker-state.hook.ts:79-80`), and the remove row renders only when a
   current marker is set (`character-marker-menu.utils.ts:125`). Every nested selection yields two or
   more covering markers — proven by that repo's own test "reports nested markers independently"
   (`character-marker-coverage.utils.test.ts:173-183`, which asserts `wj: 'all'` and `nd: 'partial'`).
   So the UI offers no removal at all in nested situations, and this behavioral change costs the
   feature nothing.
4. Refusing keeps the eventual recursive fix strictly additive. Shipping the over-removal would have
   made that later fix a behavior change on documents users had already edited.

**This remains a reading of an open product question, not a resolved one.** It is a one-line question
to the product owner, and refusing is the safe default while waiting. It should be confirmed in the
review meeting.

**Corrections made during the review, recorded because they changed the work:**

- The "a no-match call leaves the document split" framing used by three analysis agents was wrong in
  one respect, discovered while writing the test for it: Lexical's reconciliation re-merges adjacent
  simple text nodes, so the split was never observable in the resulting tree. The real cost was
  dirtying a node, which is what creates the undo entry and the collaborative-editing delta. The test
  asserts `dirtyLeaves.size === 0` accordingly. The first attempt at that test passed both with and
  without the fix — it was worthless — which is how the re-merging behavior surfaced.
- The region-placement suggestion was not followed, because scanning the file showed its convention is
  "primary function, then its own helpers", not "all exports at the top". Following the suggestion
  would have broken the pattern rather than restored it.

**Incidental discovery, out of scope but worth a look:** the first version of the end-to-end test used
the public `setSelection()` with a USJ json path covering a marked run fully, and got back
`["the Lor", {char nd content:["d"]}, " said"]` — three of four characters. With markers displayed,
the USJ-to-editor range mapping appears to shift the start offset past the NBSP but not the end. That
lives in `$getRangeFromUsjSelection`, which this branch does not touch, so it is a **possible
pre-existing off-by-one** rather than a regression. The test was rewritten to select through the
Lexical editor instead. Not investigated further.

## In-Review Quality Check

All checks were run after the in-review changes, across the whole workspace rather than only the
touched projects, because the changes reached `libs/shared`:

- `nx run-many -t typecheck lint test` — **passed for all 10 projects.** 1,633 tests passing
  (`platform-editor` 188, `shared-react` 1,322, `shared` 123), 3 skipped.
- `nx run-many -t build extract-api` — **passed for all 10 projects.** Two pre-existing
  `Unexpected ")"` CSS warnings, unrelated to these files.
- API report drift — **none.** Running `nx extract-api` after all the renames leaves `git diff` empty
  on `packages/*/etc/`, confirming the report is generated rather than hand-edited. This also closes
  the compliance analysis pass's one caveat, which had verified the report by inspection only.
- `prettier --check` on every changed file — **clean.**

Two issues were caught by the checks rather than by review, and fixed:

- `typecheck` flagged `.select()` on an un-narrowed `.find()` result in the new end-to-end test —
  invisible to the test run, since vitest does not typecheck. Fixed with an explicit type guard.
- The new test's node lookup initially matched the opening `\nd` marker node instead of the content,
  because `MarkerNode extends TextNode`. Fixed using the newly generalized
  `$isSynthesizedMarkerNode` — incidental confirmation that promoting that predicate to `libs/shared`
  was the right call.

No unfixable failures.

## Suggested Review Focus

- [ ] **The nested partial-coverage decision (highest priority).** Removing an outer marker across a
      partially covered nested marked run is now refused. This is a reasoned reading of an open
      product question about nested and partially-covered marker semantics, which the plan itself says
      needs product sign-off. Confirm with the product owner. Alternatives considered and rejected:
      over-remove (silently edits unselected text) and recursive splitting (outside the appetite, and
      the scope the innermost-marker-only decision excluded).
- [ ] **Whether "refuse" should be silent or throw.** It is currently a silent no-op, consistent with
      the method's existing behavior when the requested marker is not present, on the grounds that an
      unsupported document shape is not a programmer error. A `@throws` would be louder and
      caller-detectable. Unreachable from today's UI either way.
- [ ] **The same rule will need to bind the follow-on replace and extend work**, which share this file
      and the same splitting path. Settling it now gives those a consistent contract.
- [ ] **Pre-existing collaborative-editing gap** (`delta-apply-update.utils.ts:398,406`): bare
      `$unwrapNode` on a `CharNode`, with no marker stripping or NBSP trim. In the views that display
      markers, a remotely-applied marker removal leaves a literal `\nd`/`\nd*` and a stray NBSP.
      Out of scope here; worth its own ticket, and worth deciding whether
      `$removeCharNodeKeepingContent` should move to `libs/shared` so both paths share one
      implementation.
- [ ] **Possible pre-existing off-by-one in the USJ-to-editor range mapping** when markers are
      displayed (see Interview Notes). A full-coverage USJ range over a marked run resolved to three
      of four characters. Not investigated; may warrant its own ticket.
- [ ] **The remaining documented limitation**: in the views that display markers, a selection strictly
      inside a marked run leaves an unpaired `\nd`/`\nd*` visible, because only the covered clone is
      stripped. Presentation-only — it is excluded from USJ export and self-corrects on the next load
      — but user-visible, and the upcoming UI work will expose it. Decide whether it needs a tracked
      follow-up referenced from the docstring.
- [ ] **The `Marginal.tsx` pass-through and `insertMarker` are both untested** (no test file exists
      under `src/marginal/`). Pre-existing; decide whether that directory should get its first test
      file.
- [ ] **New idiom in tests**: `let x!: TextNode;` was introduced in five places to convert single-use
      module-scope fixtures into locals. It appears nowhere else in this repo's tests. Confirm the
      style is acceptable, or revert that one change.

AI-assisted — session: <session URL>

---

This change is [Reviewable](https://reviewable.io/reviews/eten-tech-foundation/scripture-editors/529)

### DIFF
--- a/libs/shared-react/src/plugins/usj/TextSpacingPlugin.tsx
+++ b/libs/shared-react/src/plugins/usj/TextSpacingPlugin.tsx
@@ -13,7 +13,7 @@ import {
   $isCharNode,
   $isNoteNode,
   $isParaLikeNode,
-  $isParaMarkerPrefix,
+  $isSynthesizedMarkerNode,
   $isTypedMarkNode,
   $isUnknownNode,
   CharNode,
@@ -127,7 +127,7 @@ function $verseNodeTransform(node: SomeVerseNode): void {
     // space belongs after them — and an inserted plain " " would be exporter-visible USJ
     // content that shifts every content index in the paragraph (see PT-3835). Their visual
     // separation comes from the prefix nodes' own text.
-    !$isParaMarkerPrefix(previousSibling) &&
+    !$isSynthesizedMarkerNode(previousSibling) &&
     // Bare text before a verse gets its structural space from $textNodeTrailingSpaceTransform;
     // text inside an annotation wrapper can't (that transform skips TypedMarkNode parents), so
     // the space is inserted here instead and coalesces onto the same USJ text run.
--- a/libs/shared/src/nodes/usj/node.utils.ts
+++ b/libs/shared/src/nodes/usj/node.utils.ts
@@ -563,15 +563,16 @@ export function $isVisibleMarkerNode(
 }
 
 /**
- * True for either flavor of paragraph-marker node: a `MarkerNode` (markerMode "editable") or an
+ * True for either flavor of synthesized marker node: a `MarkerNode` (markerMode "editable") or an
  * `ImmutableTypedTextNode` with `textType: "marker"` (markerMode "visible" or gutter views).
- * These are the two node shapes used to render a paragraph's USFM marker (e.g. `\p`, `\s2`,
- * `\q1`) as the visible first child of its paragraph.
+ * These are the two node shapes used to render a USFM marker as visible content — a paragraph's
+ * marker (e.g. `\p`, `\s2`, `\q1`) as the first child of its `ParaNode`, or a character marker's
+ * opening and closing markers inside its `CharNode`.
  *
  * @param node - The node to check.
  * @returns `true` if the node is a `MarkerNode` or a visible marker node.
  */
-export function $isParaMarkerPrefix(node: LexicalNode | null | undefined): boolean {
+export function $isSynthesizedMarkerNode(node: LexicalNode | null | undefined): boolean {
   return $isMarkerNode(node) || $isVisibleMarkerNode(node);
 }
 
--- a/packages/platform/README.md
+++ b/packages/platform/README.md
@@ -191,15 +191,15 @@ Programmatic control over the editor. Self-evident methods (`focus`, `undo`, `re
 `paste`, `pastePlainText`, `getUsj`, `setUsj`, `formatPara`, `getElementByKey`, `selectNote`,
 `getNoteOps`) are omitted — see the full reference below.
 
-| Method                               | Note                                                                                        |
-| ------------------------------------ | ------------------------------------------------------------------------------------------- |
-| `getSelection` / `setSelection`      | Uses json-path; assumes no comment Milestone nodes in the USJ                               |
-| `setAnnotation` / `removeAnnotation` | Ephemeral — not persisted; same json-path caveat                                            |
-| `applyUpdate` / `replaceEmbedUpdate` | EXPERIMENTAL: real-time collaborative editing                                               |
-| `insertMarker`                       | Replicates marker menu; throws if readonly, `scrRef` not provided, or marker is unsupported |
-| `removeCharacterMarker`              | Removes a character marker, keeping its text; throws if readonly or marker is unsupported   |
-| `insertNote`                         | Deprecated — use `insertMarker`                                                             |
-| `toolbarEndRef`                      | Internal use only: for dynamically adding toolbar controls                                  |
+| Method                               | Note                                                                                                                                                                                                                             |
+| ------------------------------------ | -------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
+| `getSelection` / `setSelection`      | Uses json-path; assumes no comment Milestone nodes in the USJ                                                                                                                                                                    |
+| `setAnnotation` / `removeAnnotation` | Ephemeral — not persisted; same json-path caveat                                                                                                                                                                                 |
+| `applyUpdate` / `replaceEmbedUpdate` | EXPERIMENTAL: real-time collaborative editing                                                                                                                                                                                    |
+| `insertMarker`                       | Replicates marker menu; throws if readonly, `scrRef` not provided, or marker is unsupported                                                                                                                                      |
+| `removeCharacterMarker`              | Removes a character marker, keeping its text. Returns `false` and changes nothing when no marker matches, the selection is in a note, or removal can't be confined to the selection; throws if readonly or marker is unsupported |
+| `insertNote`                         | Deprecated — use `insertMarker`                                                                                                                                                                                                  |
+| `toolbarEndRef`                      | Internal use only: for dynamically adding toolbar controls                                                                                                                                                                       |
 
 > Full API reference: [platform-editor.api.md](etc/platform-editor.api.md)
 
--- a/packages/platform/README.md
+++ b/packages/platform/README.md
@@ -197,6 +197,7 @@ Programmatic control over the editor. Self-evident methods (`focus`, `undo`, `re
 | `setAnnotation` / `removeAnnotation` | Ephemeral — not persisted; same json-path caveat                                            |
 | `applyUpdate` / `replaceEmbedUpdate` | EXPERIMENTAL: real-time collaborative editing                                               |
 | `insertMarker`                       | Replicates marker menu; throws if readonly, `scrRef` not provided, or marker is unsupported |
+| `removeCharacterMarker`              | Removes a character marker, keeping its text; throws if readonly or marker is unsupported   |
 | `insertNote`                         | Deprecated — use `insertMarker`                                                             |
 | `toolbarEndRef`                      | Internal use only: for dynamically adding toolbar controls                                  |
 
--- a/packages/platform/etc/platform-editor.api.md
+++ b/packages/platform/etc/platform-editor.api.md
@@ -110,6 +110,7 @@ export interface EditorRef {
     pastePlainText(): void;
     redo(): void;
     removeAnnotation(type: string, id: string): void;
+    removeCharMarker(marker?: string): void;
     replaceEmbedUpdate(embedNodeKey: string, insertEmbedOps: DeltaOp[]): void;
     selectNote(noteKeyOrIndex: string | number): void;
     setAnnotation(selection: AnnotationRange, type: string, id: string, callbacks?: {
--- a/packages/platform/etc/platform-editor.api.md
+++ b/packages/platform/etc/platform-editor.api.md
@@ -110,7 +110,7 @@ export interface EditorRef {
     pastePlainText(): void;
     redo(): void;
     removeAnnotation(type: string, id: string): void;
-    removeCharacterMarker(marker?: string): void;
+    removeCharacterMarker(marker?: string): boolean;
     replaceEmbedUpdate(embedNodeKey: string, insertEmbedOps: DeltaOp[]): void;
     selectNote(noteKeyOrIndex: string | number): void;
     setAnnotation(selection: AnnotationRange, type: string, id: string, callbacks?: {
--- a/packages/platform/etc/platform-editor.api.md
+++ b/packages/platform/etc/platform-editor.api.md
@@ -110,7 +110,7 @@ export interface EditorRef {
     pastePlainText(): void;
     redo(): void;
     removeAnnotation(type: string, id: string): void;
-    removeCharMarker(marker?: string): void;
+    removeCharacterMarker(marker?: string): void;
     replaceEmbedUpdate(embedNodeKey: string, insertEmbedOps: DeltaOp[]): void;
     selectNote(noteKeyOrIndex: string | number): void;
     setAnnotation(selection: AnnotationRange, type: string, id: string, callbacks?: {
--- a/packages/platform/src/editor/Editor.test.tsx
+++ b/packages/platform/src/editor/Editor.test.tsx
@@ -220,12 +220,30 @@ describe("removeCharacterMarker guards", () => {
     expect(() => editor.removeCharacterMarker("zzz")).toThrow("Unsupported character marker 'zzz'");
   });
 
-  it("does not throw when the marker is omitted and there is no selection", async () => {
+  it.each(["ft", "xt"])(
+    "throws for the note-only character marker '%s', which removal always skips",
+    async (marker) => {
+      const ref = await createEditorRefForTesting();
+      const editor = ref.current;
+      if (!editor) throw new Error("Editor not mounted");
+
+      // CharNode.isValidMarker accepts these — VALID_CHAR_MARKERS spreads in the footnote and
+      // cross-reference markers — but they only ever occur inside a NoteNode, which
+      // $getCharNodeToRemove skips. Throwing beats accepting the call and silently doing nothing.
+      expect(() => editor.removeCharacterMarker(marker)).toThrow(
+        `Unsupported character marker '${marker}'`,
+      );
+    },
+  );
+
+  it("returns false without throwing when the marker is omitted and there is no selection", async () => {
     const ref = await createEditorRefForTesting();
     const editor = ref.current;
     if (!editor) throw new Error("Editor not mounted");
 
-    expect(() => editor.removeCharacterMarker()).not.toThrow();
+    // The return value is what makes this more than a smoke test: a fresh editor has no selection,
+    // so the call must report that it removed nothing rather than merely not crashing.
+    expect(editor.removeCharacterMarker()).toBe(false);
   });
 });
 
--- a/packages/platform/src/editor/Editor.test.tsx
+++ b/packages/platform/src/editor/Editor.test.tsx
@@ -182,6 +182,52 @@ describe("setAnnotation overload", () => {
   });
 });
 
+describe("removeCharMarker guards", () => {
+  async function createReadonlyEditorRefForTesting(): Promise<RefObject<EditorRef | null>> {
+    const ref = createRef<EditorRef>();
+    await act(async () => {
+      render(<Editor ref={ref} defaultUsj={sampleUsj} options={{ isReadonly: true }} />);
+    });
+    if (!ref.current) throw new Error("EditorRef did not mount");
+    return ref;
+  }
+
+  it("throws in readonly mode", async () => {
+    const ref = await createReadonlyEditorRefForTesting();
+    const editor = ref.current;
+    if (!editor) throw new Error("Editor not mounted");
+
+    expect(() => editor.removeCharMarker("nd")).toThrow(
+      "Cannot remove char marker in readonly mode",
+    );
+  });
+
+  it("throws for a para marker, which removal can never act on", async () => {
+    const ref = await createEditorRefForTesting();
+    const editor = ref.current;
+    if (!editor) throw new Error("Editor not mounted");
+
+    // Note this is stricter than insertMarker's isUsjMarkerSupported, which accepts "p".
+    expect(() => editor.removeCharMarker("p")).toThrow("Unsupported char marker 'p'");
+  });
+
+  it("throws for an unknown marker", async () => {
+    const ref = await createEditorRefForTesting();
+    const editor = ref.current;
+    if (!editor) throw new Error("Editor not mounted");
+
+    expect(() => editor.removeCharMarker("zzz")).toThrow("Unsupported char marker 'zzz'");
+  });
+
+  it("does not throw when the marker is omitted and there is no selection", async () => {
+    const ref = await createEditorRefForTesting();
+    const editor = ref.current;
+    if (!editor) throw new Error("Editor not mounted");
+
+    expect(() => editor.removeCharMarker()).not.toThrow();
+  });
+});
+
 /** Grabs the underlying Lexical editor so tests can dispatch commands the public ref doesn't expose. */
 function GrabEditor({ onEditor }: { onEditor: (editor: LexicalEditor) => void }): null {
   const [editor] = useLexicalComposerContext();
--- a/packages/platform/src/editor/Editor.test.tsx
+++ b/packages/platform/src/editor/Editor.test.tsx
@@ -8,8 +8,9 @@ import { flushQueuedEvents } from "./editor-test.utils";
 import { ContentJsonPath, Usj } from "@eten-tech-foundation/scripture-utilities";
 import { useLexicalComposerContext } from "@lexical/react/LexicalComposerContext";
 import { act, render } from "@testing-library/react";
-import { KEY_DOWN_COMMAND, LexicalEditor } from "lexical";
+import { $getRoot, $isTextNode, KEY_DOWN_COMMAND, LexicalEditor, TextNode } from "lexical";
 import { createRef, RefObject, useEffect } from "react";
+import { $isCharNode, $isSomeParaNode, $isSynthesizedMarkerNode, NBSP } from "shared";
 import { vi } from "vitest";
 
 /** USJ with book PSA for Editor sync effect test (clone of usjGen1v1 with book code changed) */
@@ -182,7 +183,7 @@ describe("setAnnotation overload", () => {
   });
 });
 
-describe("removeCharMarker guards", () => {
+describe("removeCharacterMarker guards", () => {
   async function createReadonlyEditorRefForTesting(): Promise<RefObject<EditorRef | null>> {
     const ref = createRef<EditorRef>();
     await act(async () => {
@@ -197,8 +198,8 @@ describe("removeCharMarker guards", () => {
     const editor = ref.current;
     if (!editor) throw new Error("Editor not mounted");
 
-    expect(() => editor.removeCharMarker("nd")).toThrow(
-      "Cannot remove char marker in readonly mode",
+    expect(() => editor.removeCharacterMarker("nd")).toThrow(
+      "Cannot remove character marker in readonly mode",
     );
   });
 
@@ -208,7 +209,7 @@ describe("removeCharMarker guards", () => {
     if (!editor) throw new Error("Editor not mounted");
 
     // Note this is stricter than insertMarker's isUsjMarkerSupported, which accepts "p".
-    expect(() => editor.removeCharMarker("p")).toThrow("Unsupported char marker 'p'");
+    expect(() => editor.removeCharacterMarker("p")).toThrow("Unsupported character marker 'p'");
   });
 
   it("throws for an unknown marker", async () => {
@@ -216,7 +217,7 @@ describe("removeCharMarker guards", () => {
     const editor = ref.current;
     if (!editor) throw new Error("Editor not mounted");
 
-    expect(() => editor.removeCharMarker("zzz")).toThrow("Unsupported char marker 'zzz'");
+    expect(() => editor.removeCharacterMarker("zzz")).toThrow("Unsupported character marker 'zzz'");
   });
 
   it("does not throw when the marker is omitted and there is no selection", async () => {
@@ -224,7 +225,7 @@ describe("removeCharMarker guards", () => {
     const editor = ref.current;
     if (!editor) throw new Error("Editor not mounted");
 
-    expect(() => editor.removeCharMarker()).not.toThrow();
+    expect(() => editor.removeCharacterMarker()).not.toThrow();
   });
 });
 
@@ -314,5 +315,89 @@ describe("undo after a verse-spanning delete (PT-4102 regression)", () => {
     expect(afterUndo).toContain("Alpha ");
     expect(afterUndo).toContain("Bravo");
     expect(afterUndo).toContain('"number":"2"');
+  });
+});
+
+const usjWithCharMarker: Usj = {
+  type: "USJ",
+  version: "3.1",
+  content: [
+    { type: "book", marker: "id", code: "GEN", content: ["Test Book"] },
+    { type: "chapter", marker: "c", number: "1" },
+    {
+      type: "para",
+      marker: "p",
+      content: ["the ", { type: "char", marker: "nd", content: ["Lord"] }, " said"],
+    },
+  ],
+};
+
+/** Selects the whole content of the first `CharNode` in the document. */
+async function selectCharNodeContent(editor: LexicalEditor): Promise<void> {
+  await act(async () => {
+    editor.update(
+      () => {
+        const para = $getRoot().getChildren().find($isSomeParaNode);
+        const charNode = para?.getChildren().find($isCharNode);
+        // Not just `$isTextNode`: `MarkerNode` extends `TextNode`, so under
+        // `markerMode: "editable"` the first match would be the opening `\nd` marker.
+        const textNode = charNode
+          ?.getChildren()
+          .find(
+            (child): child is TextNode => $isTextNode(child) && !$isSynthesizedMarkerNode(child),
+          );
+        if (!textNode) throw new Error("Expected a text node inside a CharNode");
+        textNode.select(0, textNode.getTextContentSize());
+      },
+      { discrete: true },
+    );
+  });
+}
+
+describe("removeCharacterMarker through the editor ref", () => {
+  // The guards above only prove the method throws when it should. This drives it end to end so
+  // `Editor.tsx`'s wiring is covered too - in particular that it forwards its `viewOptions` as the
+  // third argument. `viewOptions` is what enables the NBSP trim, and every adaptor-level test
+  // passes its own view options directly, so a wiring that dropped that argument would leave all
+  // of them green while leaving an NBSP in the text of the real editor.
+  it("removes the marker and its NBSP under markerMode 'editable'", async () => {
+    const ref = createRef<EditorRef>();
+    let editor: LexicalEditor | undefined;
+    await act(async () => {
+      render(
+        <Editor
+          ref={ref}
+          defaultUsj={usjWithCharMarker}
+          options={{
+            view: {
+              markerMode: "editable",
+              noteMode: "expanded",
+              hasSpacing: false,
+              isFormattedFont: false,
+            },
+          }}
+        >
+          <GrabEditor onEditor={(e) => (editor = e)} />
+        </Editor>,
+      );
+    });
+    await flushQueuedEvents();
+    if (!ref.current || !editor) throw new Error("EditorRef did not mount");
+    const editorRef = ref.current;
+
+    // Precondition: the adaptor really did prepend the NBSP, so the trim below has something to do.
+    expect(editor.getEditorState().read(() => $getRoot().getTextContent())).toContain(NBSP);
+
+    await selectCharNodeContent(editor);
+    await act(async () => {
+      editorRef.removeCharacterMarker("nd");
+    });
+    await flushQueuedEvents();
+
+    const para = editorRef.getUsj()?.content[2];
+    if (typeof para !== "object" || !("content" in para))
+      throw new Error("para is not a USJ para node");
+    expect(JSON.stringify(para.content)).not.toContain('"char"');
+    expect(para.content?.join("")).toBe("the Lord said");
   });
 });
--- a/packages/platform/src/editor/Editor.tsx
+++ b/packages/platform/src/editor/Editor.tsx
@@ -321,11 +321,19 @@ const Editor = forwardRef(function Editor<TLogger extends LoggerBasic>(
       )
         throw new Error(`Unsupported character marker '${marker}'`);
 
-      editorRef.current?.update(() => {
-        const selection = $getSelection();
-        if ($isRangeSelection(selection))
-          $removeCharacterMarkerAtSelection(selection, marker, viewOptions);
-      });
+      // `discrete` so the update runs now rather than being deferred behind an in-progress one,
+      // which would leave `didRemove` reporting `false` for a removal that did happen. Same reason
+      // `applyUpdate` above uses it.
+      let didRemove = false;
+      editorRef.current?.update(
+        () => {
+          const selection = $getSelection();
+          if ($isRangeSelection(selection))
+            didRemove = $removeCharacterMarkerAtSelection(selection, marker, viewOptions);
+        },
+        { discrete: true },
+      );
+      return didRemove;
     },
     insertMarker(marker) {
       if (isReadonly) throw new Error("Cannot insert marker in readonly mode");
--- a/packages/platform/src/editor/Editor.tsx
+++ b/packages/platform/src/editor/Editor.tsx
@@ -1,8 +1,9 @@
 import editorUsjAdaptor from "./adaptors/editor-usj.adaptor";
 import usjEditorAdaptor from "./adaptors/usj-editor.adaptor";
 import {
-  $removeCharMarkerAtSelection,
+  $removeCharacterMarkerAtSelection,
   getUsjMarkerAction,
+  isCharacterMarkerSupported,
   isUsjMarkerSupported,
 } from "./adaptors/usj-marker-action.utils";
 import { EditorOptions, EditorProps, EditorRef } from "./editor.model";
@@ -47,7 +48,6 @@ import {
 import {
   $createParaNode,
   blackListedChangeTags,
-  CharNode,
   DELTA_CHANGE_TAG,
   externalTypedMarkType,
   LoggerBasic,
@@ -313,15 +313,18 @@ const Editor = forwardRef(function Editor<TLogger extends LoggerBasic>(
         () => editorRef.current?.getElementByKey(nodeKey) ?? undefined,
       );
     },
-    removeCharMarker(marker) {
-      if (isReadonly) throw new Error("Cannot remove char marker in readonly mode");
-      if (marker !== undefined && !CharNode.isValidMarker(marker, nodeOptions?.extraValidMarkers))
-        throw new Error(`Unsupported char marker '${marker}'`);
+    removeCharacterMarker(marker) {
+      if (isReadonly) throw new Error("Cannot remove character marker in readonly mode");
+      if (
+        marker !== undefined &&
+        !isCharacterMarkerSupported(marker, nodeOptions.extraValidMarkers)
+      )
+        throw new Error(`Unsupported character marker '${marker}'`);
 
       editorRef.current?.update(() => {
         const selection = $getSelection();
         if ($isRangeSelection(selection))
-          $removeCharMarkerAtSelection(selection, marker, viewOptions);
+          $removeCharacterMarkerAtSelection(selection, marker, viewOptions);
       });
     },
     insertMarker(marker) {
--- a/packages/platform/src/editor/Editor.tsx
+++ b/packages/platform/src/editor/Editor.tsx
@@ -1,6 +1,10 @@
 import editorUsjAdaptor from "./adaptors/editor-usj.adaptor";
 import usjEditorAdaptor from "./adaptors/usj-editor.adaptor";
-import { getUsjMarkerAction, isUsjMarkerSupported } from "./adaptors/usj-marker-action.utils";
+import {
+  $removeCharMarkerAtSelection,
+  getUsjMarkerAction,
+  isUsjMarkerSupported,
+} from "./adaptors/usj-marker-action.utils";
 import { EditorOptions, EditorProps, EditorRef } from "./editor.model";
 import editorTheme from "./editor.theme";
 import { ActiveTextPlugin } from "./ActiveTextPlugin";
@@ -43,6 +47,7 @@ import {
 import {
   $createParaNode,
   blackListedChangeTags,
+  CharNode,
   DELTA_CHANGE_TAG,
   externalTypedMarkType,
   LoggerBasic,
@@ -307,6 +312,17 @@ const Editor = forwardRef(function Editor<TLogger extends LoggerBasic>(
       return editorRef.current?.read(
         () => editorRef.current?.getElementByKey(nodeKey) ?? undefined,
       );
+    },
+    removeCharMarker(marker) {
+      if (isReadonly) throw new Error("Cannot remove char marker in readonly mode");
+      if (marker !== undefined && !CharNode.isValidMarker(marker, nodeOptions?.extraValidMarkers))
+        throw new Error(`Unsupported char marker '${marker}'`);
+
+      editorRef.current?.update(() => {
+        const selection = $getSelection();
+        if ($isRangeSelection(selection))
+          $removeCharMarkerAtSelection(selection, marker, viewOptions);
+      });
     },
     insertMarker(marker) {
       if (isReadonly) throw new Error("Cannot insert marker in readonly mode");
--- a/packages/platform/src/editor/ParaMarkerPrefixGuardPlugin.tsx
+++ b/packages/platform/src/editor/ParaMarkerPrefixGuardPlugin.tsx
@@ -1,6 +1,6 @@
 import { useLexicalComposerContext } from "@lexical/react/LexicalComposerContext";
 import { useEffect } from "react";
-import { $isParaMarkerPrefix, LoggerBasic, ParaNode, PARA_MARKER_DEFAULT } from "shared";
+import { $isSynthesizedMarkerNode, LoggerBasic, ParaNode, PARA_MARKER_DEFAULT } from "shared";
 import { ViewOptions } from "shared-react";
 
 /**
@@ -58,7 +58,7 @@ export function ParaMarkerPrefixGuardPlugin({
 export function $resetMarkerIfPrefixDeleted(para: ParaNode, logger?: LoggerBasic): void {
   if (para.getMarker() === PARA_MARKER_DEFAULT) return;
   if (para.isEmpty()) return;
-  if ($isParaMarkerPrefix(para.getFirstChild())) return;
+  if ($isSynthesizedMarkerNode(para.getFirstChild())) return;
   logger?.debug(
     `[ParaMarkerPrefixGuard] Resetting paragraph "${para.getMarker()}" → "${PARA_MARKER_DEFAULT}" (key ${para.getKey()})`,
   );
--- a/packages/platform/src/editor/adaptors/usj-marker-action-utils.test.ts
+++ b/packages/platform/src/editor/adaptors/usj-marker-action-utils.test.ts
@@ -51,6 +51,7 @@ const reference = { book: "GEN", chapterNum: 1, verseNum: 1 };
 let secondVerseTextNode: TextNode;
 let charTextNode: TextNode;
 let placeholderTextNode: TextNode;
+let targetTextNode: TextNode;
 let noteTextNode: TextNode;
 let tailTextNode: TextNode;
 let firstCharTextNode: TextNode;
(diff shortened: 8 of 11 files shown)


katherinejensen00 (changes requested):
Thanks for your hard work on this, Jolie! Despite all of the comments, only one is actually worth blocking this PR over. This code could honestly merge as is and be great. Don't feel like you need to do all of the lows and nits. Just giving you the info/options.

> @katherinejensen00 reviewed 11 files and all commit messages, and made 28 comments.
> *[Reviewable](https://reviewable.io/reviews/eten-tech-foundation/scripture-editors/529#-:-OzRqDYY6No6YYD2I1v-:b-dvtig3)* status: all files reviewed, 27 unresolved discussions (waiting on jolierabideau).

___
*[`packages/platform/README.md` line 200 at r1](https://reviewable.io/reviews/eten-tech-foundation/scripture-editors/529#-OzRq7dnC1_DOifVCJ90:-OzRq7dnC1_DOifVCJ91:bbae524) ([raw file](https://github.com/eten-tech-foundation/scripture-editors/blob/92ce3882a5832721da34d9dd7383a64759c0323d/packages/platform/README.md#L200)):*
> ```Markdown
> | `applyUpdate` / `replaceEmbedUpdate` | EXPERIMENTAL: real-time collaborative editing                                               |
> | `insertMarker`                       | Replicates marker menu; throws if readonly, `scrRef` not provided, or marker is unsupported |
> | `removeCharacterMarker`              | Removes a character marker, keeping its text; throws if readonly or marker is unsupported   |
> ```

NIT

README documents only the throw behavior, omitting the silent-refuse/no-op semantics
___
*[`packages/platform/src/editor/Editor.test.tsx` line 128 at r1](https://reviewable.io/reviews/eten-tech-foundation/scripture-editors/529#-OzRp2r45BWxA-HxsU4o:-OzRp2r45BWxA-HxsU4p:b-z7k9h0) ([raw file](https://github.com/eten-tech-foundation/scripture-editors/blob/92ce3882a5832721da34d9dd7383a64759c0323d/packages/platform/src/editor/Editor.test.tsx#L128)):*
> ```XML
> describe("setAnnotation overload", () => {
>   it("accepts the deprecated positional form (onClick, onRemove)", async () => {
>     const ref = await createEditorRefForTesting();
> ```

NIT

Near-vacuous "no selection" guard test (`.not.toThrow()` on a fresh editor)
___
*[`packages/platform/src/editor/editor.model.ts` line 127 at r1](https://reviewable.io/reviews/eten-tech-foundation/scripture-editors/529#-OzRpu-v4LtJLRq-HfsQ:-OzRpu-v4LtJLRq-HfsR:b-ibx4fs) ([raw file](https://github.com/eten-tech-foundation/scripture-editors/blob/92ce3882a5832721da34d9dd7383a64759c0323d/packages/platform/src/editor/editor.model.ts#L127)):*
> ```XML
>    * marker, leaving the uncovered text marked) selections.
>    *
>    * Does nothing, without throwing, when there is no matching character marker enclosing the
> ```

NIT

Public `removeCharacterMarker` `@remarks` reads dense in IntelliSense — optional one-line summary first.
___
*[`packages/platform/src/editor/editor.model.ts` line 151 at r1](https://reviewable.io/reviews/eten-tech-foundation/scripture-editors/529#-OzRmWWPBUsztljt2wiT:-OzRmWWPBUsztljt2wiU:b-mhs3v3) ([raw file](https://github.com/eten-tech-foundation/scripture-editors/blob/92ce3882a5832721da34d9dd7383a64759c0323d/packages/platform/src/editor/editor.model.ts#L151)):*
> ```XML
>    * @param marker - A USFM character marker string, e.g. `"nd"`, `"wj"`. Omit to remove the
>    *   innermost character marker enclosing the selection.
>    * @throws Will throw an error if the editor is in readonly mode.
> ```

NIT

**Severity: Low–Medium** 

The method returns `void`, so a caller can't tell "removed" from "refused/nothing matched."_ The paranext-core seam gets no signal; a toolbar could show success while the marker stays. You flagged silent-vs-throw as open — worth settling (a `boolean` return is the cheapest observable) before the consumer wires it, since the signature is free to change now and breaking once consumed.
___
*[`packages/platform/src/editor/adaptors/usj-marker-action-utils.test.ts` line 340 at r1](https://reviewable.io/reviews/eten-tech-foundation/scripture-editors/529#-OzRobzV8BNpkj8aM8CO:-OzRobzV8BNpkj8aM8CP:b-dljyfp) ([raw file](https://github.com/eten-tech-foundation/scripture-editors/blob/92ce3882a5832721da34d9dd7383a64759c0323d/packages/platform/src/editor/adaptors/usj-marker-action-utils.test.ts#L340)):*
> ```XML
> 
>   describe("should remove a character marker", () => {
>     it("when the cursor is collapsed inside a CharNode", () => {
> ```

NIT 

Non-NBSP collapsed-caret restore offset unasserted.
___
*[`packages/platform/src/editor/adaptors/usj-marker-action-utils.test.ts` line 356 at r1](https://reviewable.io/reviews/eten-tech-foundation/scripture-editors/529#-OzRo3ne17Vt67ml8aVT:-OzRo3ne17Vt67ml8aVU:b-qfhbvk) ([raw file](https://github.com/eten-tech-foundation/scripture-editors/blob/92ce3882a5832721da34d9dd7383a64759c0323d/packages/platform/src/editor/adaptors/usj-marker-action-utils.test.ts#L356)):*
> ```XML
> 
>       editor.getEditorState().read(() => {
>         const para = $getRoot().getFirstChild();
> ```

NIT

Test-fixture duplication (~13 paragraph fixtures, 23 unwrap-and-assert prologues).
___
*[`packages/platform/src/editor/adaptors/usj-marker-action-utils.test.ts` line 427 at r1](https://reviewable.io/reviews/eten-tech-foundation/scripture-editors/529#-OzRoS383bvfq44nDzw8:-OzRoS383bvfq44nDzw9:b-9b46l4) ([raw file](https://github.com/eten-tech-foundation/scripture-editors/blob/92ce3882a5832721da34d9dd7383a64759c0323d/packages/platform/src/editor/adaptors/usj-marker-action-utils.test.ts#L427)):*
> ```XML
>       });
>     });
> ```

NIT **Severity: Low** 

The "dirties no node" test only counts `dirtyLeaves`._ That's exactly the blind spot for #1 — an element-level dirty (the CharNode) wouldn't be caught. Add `dirtyElements` to the assertion, and once #1 is fixed, extend this test to the refuse path.
___
*[`packages/platform/src/editor/adaptors/usj-marker-action-utils.test.ts` line 493 at r1](https://reviewable.io/reviews/eten-tech-foundation/scripture-editors/529#-OzRoJhk1B1tJ65S4Kv5:-OzRoJhk1B1tJ65S4Kv6:bssauul) ([raw file](https://github.com/eten-tech-foundation/scripture-editors/blob/92ce3882a5832721da34d9dd7383a64759c0323d/packages/platform/src/editor/adaptors/usj-marker-action-utils.test.ts#L493)):*
> ```XML
>     });
> 
>     it("splits off only a trailing clone when the selection covers the leading text", () => {
> ```

NIT **Severity: Low** 

The cid copy is tested, but the re-merge it exists to enable is not._ `$charNodeTransform` lives in `CharNodePlugin` (shared-react), which `createBasicTestEnvironment` doesn't install — so nothing exercises the split halves actually re-merging. Worth one test in an environment that registers the plugin.
___
*[`packages/platform/src/editor/adaptors/usj-marker-action-utils.test.ts` line 953 at r1](https://reviewable.io/reviews/eten-tech-foundation/scripture-editors/529#-OzRon9FE58QmkCHe1p-:-OzRon9FE58QmkCHe1p0:bdhj405) ([raw file](https://github.com/eten-tech-foundation/scripture-editors/blob/92ce3882a5832721da34d9dd7383a64759c0323d/packages/platform/src/editor/adaptors/usj-marker-action-utils.test.ts#L953)):*
> ```XML
>     });
> 
>     it("restores the selection after the NBSP trim shifts content length in markerMode 'editable'", () => {
> ```

NIT

Backward × NBSP-trim restore never co-tested; visible-mode selection restore unasserted.
___
*[`packages/platform/src/editor/adaptors/usj-marker-action-utils.test.ts` line 1072 at r1](https://reviewable.io/reviews/eten-tech-foundation/scripture-editors/529#-OzRot7W6Mk8kRWqxKAI:-OzRot7W6Mk8kRWqxKAJ:b8kkugk) ([raw file](https://github.com/eten-tech-foundation/scripture-editors/blob/92ce3882a5832721da34d9dd7383a64759c0323d/packages/platform/src/editor/adaptors/usj-marker-action-utils.test.ts#L1072)):*
> ```XML
>         // leading/trailing clone that $removeCharNodeKeepingContent never sees — the marker is
>         // never stripped. Asserting today's actual (broken) output so a future fix to this
>         // limitation is noticed here rather than silently reintroduced.
> ```

NIT

Known-limitation visible-mode test may not trip an adaptor-level fix (pins in-editor text)
___
*[`packages/platform/src/editor/adaptors/usj-marker-action.utils.ts` line 88 at r1](https://reviewable.io/reviews/eten-tech-foundation/scripture-editors/529#-OzRm7-E5Q5ebN9jF5y2:-OzRm7-E5Q5ebN9jF5y3:b-vuesz2) ([raw file](https://github.com/eten-tech-foundation/scripture-editors/blob/92ce3882a5832721da34d9dd7383a64759c0323d/packages/platform/src/editor/adaptors/usj-marker-action.utils.ts#L88)):*
> ```XML
> 
> /** Returns whether the given USFM marker is supported by {@link getUsjMarkerAction}. */
> export function isUsjMarkerSupported(marker: string): boolean {
> ```

NIT **Severity: Low** 

A project's configured extra char marker is removable but not insertable._ `isUsjMarkerSupported` (used by `insertMarker`) calls `CharNode.isValidMarker(marker)` without `extraValidMarkers`, while your new `isCharacterMarkerSupported` honors them. You documented the asymmetry, but the real inconsistency is on the insert side — worth threading extras into `isUsjMarkerSupported` in a follow-up so the two agree.
___
*[`packages/platform/src/editor/adaptors/usj-marker-action.utils.ts` line 111 at r1](https://reviewable.io/reviews/eten-tech-foundation/scripture-editors/529#-OzRlczj5na4djAPKVXk:-OzRlczj5na4djAPKVXl:b-v1gwk6) ([raw file](https://github.com/eten-tech-foundation/scripture-editors/blob/92ce3882a5832721da34d9dd7383a64759c0323d/packages/platform/src/editor/adaptors/usj-marker-action.utils.ts#L111)):*
> ```XML
>  * @returns `true` if the marker is a character marker.
>  */
> export function isCharacterMarkerSupported(
> ```

NIT **Severity: Low** 

Validation admits footnote/cross-reference markers that removal can never act on._ `VALID_CHAR_MARKERS` spreads in `VALID_CHAR_FOOTNOTE_MARKERS`/`VALID_CHAR_CROSS_REFERENCE_MARKERS`, so `removeCharacterMarker('ft')`/`('xt')` passes the guard and then silently no-ops (those live in notes, which `$getCharNodeToRemove` skips). That contradicts the docstring's "removal only ever targets a CharNode, so `p` must be rejected" strictness. Either exclude the note markers or soften the docstring claim.
___
*[`packages/platform/src/editor/adaptors/usj-marker-action.utils.ts` line 394 at r1](https://reviewable.io/reviews/eten-tech-foundation/scripture-editors/529#-OzRnVAuAdsTc9cli6X4:-OzRnVAuAdsTc9cli6X5:b-oyhycn) ([raw file](https://github.com/eten-tech-foundation/scripture-editors/blob/92ce3882a5832721da34d9dd7383a64759c0323d/packages/platform/src/editor/adaptors/usj-marker-action.utils.ts#L394)):*
> ```XML
>  * @param viewOptions - View options, used to strip synthesized marker content.
>  */
> export function $removeCharacterMarkerAtSelection(
> ```

NIT 

Duplicated collapsed-vs-range restore blocks + multi-paragraph comments.
___
*[`packages/platform/src/editor/adaptors/usj-marker-action.utils.ts` line 430 at r1](https://reviewable.io/reviews/eten-tech-foundation/scripture-editors/529#-OzRnhVrB1mHtJgcTMUU:-OzRnhVrB1mHtJgcTMUV:b-ihhkwz) ([raw file](https://github.com/eten-tech-foundation/scripture-editors/blob/92ce3882a5832721da34d9dd7383a64759c0323d/packages/platform/src/editor/adaptors/usj-marker-action.utils.ts#L430)):*
> ```XML
>   // Check there is something to remove before the loop below starts splitting text nodes, so a
>   // no-match request mutates nothing at all. See `$hasCharNodeToRemove`.
>   if (!$hasCharNodeToRemove(nodes, marker)) return;
> ```

NIT 

Redundant double ancestor-walk (`$hasCharNodeToRemove` then the loop re-walks)
___
*[`packages/platform/src/editor/adaptors/usj-marker-action.utils.ts` line 469 at r1](https://reviewable.io/reviews/eten-tech-foundation/scripture-editors/529#-OzRpEdo8pCd7GC6gSBx:-OzRpEdo8pCd7GC6gSBy:b-qv9ocn) ([raw file](https://github.com/eten-tech-foundation/scripture-editors/blob/92ce3882a5832721da34d9dd7383a64759c0323d/packages/platform/src/editor/adaptors/usj-marker-action.utils.ts#L469)):*
> ```XML
>   // - `isBackward` is captured before the loop: a backward range's anchor is the *last* target, so
>   //   swapping the roles keeps it backward instead of normalizing to forward.
>   // - `$getSelection()` is re-fetched rather than reusing `selection`: `TextNode.replace()` (via
> ```

NIT **Severity: Nit (please verify)** 

Please confirm the `$unwrapNode` → `TextNode.replace()` wording against the pinned Lexical 0.43.0 source._ One finder believed 0.43.0's `$unwrapNode` is caret-splice-based and never calls `replace()`; I couldn't verify locally (no `node_modules`). The manual `$getSelection()` re-fetch is correct regardless (the NBSP `setTextContent` trim genuinely needs it), so this is only about the comment's stated mechanism.
___
*[`packages/platform/src/editor/adaptors/usj-marker-action.utils.ts` line 476 at r1](https://reviewable.io/reviews/eten-tech-foundation/scripture-editors/529#-OzRi-80EEh6HkYiUUYL:-OzRi-80EEh6HkYiUUYM:btm6bag) ([raw file](https://github.com/eten-tech-foundation/scripture-editors/blob/92ce3882a5832721da34d9dd7383a64759c0323d/packages/platform/src/editor/adaptors/usj-marker-action.utils.ts#L476)):*
> ```XML
>   // was removed outright — in which case Lexical's own selection repair applies instead.
>   const currentSelection = $getSelection();
>   const firstTargetNode = targetNodes[0];
> ```

NIT **Severity: Low** 

A fully-refused call still overwrites the user's selection._ Execution falls through to the restore block and calls `setTextNodeRange` even when nothing was removed. Benign in the single-node case (the range equals the original), but on a documented no-op the selection shouldn't be touched. Resolved for free by the early-return in #1.
___
*[`packages/platform/src/editor/adaptors/usj-marker-action.utils.ts` line 560 at r1](https://reviewable.io/reviews/eten-tech-foundation/scripture-editors/529#-OzRpS4d7dnzeEySurfu:-OzRpS4d7dnzeEySurfv:bjemtw9) ([raw file](https://github.com/eten-tech-foundation/scripture-editors/blob/92ce3882a5832721da34d9dd7383a64759c0323d/packages/platform/src/editor/adaptors/usj-marker-action.utils.ts#L560)):*
> ```XML
>  * @returns `true` if the selection covers all of the element's content.
>  */
> function $isFullyCoveredByTargets(element: ElementNode, targetKeys: Set<string>): boolean {
> ```

NIT

`$isFullyCoveredByTargets` docstring implies a uniform "synthesized markers don't count" mechanism, but in visible mode `ImmutableTypedTextNode extends DecoratorNode` so it never appears in `getAllTextNodes()`; the clause only fires for editable-mode `MarkerNode`s.
___
*[`packages/platform/src/editor/adaptors/usj-marker-action.utils.ts` line 588 at r1](https://reviewable.io/reviews/eten-tech-foundation/scripture-editors/529#-OzRmm4_2-N_e-Jkyo_M:-OzRmm4_2-N_e-Jkyo_N:bvk3nuu) ([raw file](https://github.com/eten-tech-foundation/scripture-editors/blob/92ce3882a5832721da34d9dd7383a64759c0323d/packages/platform/src/editor/adaptors/usj-marker-action.utils.ts#L588)):*
> ```XML
>  * covered.
>  *
>  * Known limitation — interior partial coverage under marker mode: the fold above only reaches a
> ```

NIT **Severity: Low (documented)** 

Interior partial coverage strands an unpaired marker half in marker-display views._ PresentaNIT

tion-only and self-corrects on reload; the real fix (regenerate paired open/close markers per surviving clone) is adaptor-level. Fine to keep as the documented limitation; consider a tracked follow-up referenced from the docstring.
___
*[`packages/platform/src/editor/adaptors/usj-marker-action.utils.ts` line 588 at r1](https://reviewable.io/reviews/eten-tech-foundation/scripture-editors/529#-OzRpmJICri-_Sa7Ppbn:-OzRpmJICri-_Sa7Ppbo:bf1bfhz) ([raw file](https://github.com/eten-tech-foundation/scripture-editors/blob/92ce3882a5832721da34d9dd7383a64759c0323d/packages/platform/src/editor/adaptors/usj-marker-action.utils.ts#L588)):*
> ```XML
>  * covered.
>  *
>  * Known limitation — interior partial coverage under marker mode: the fold above only reaches a
> ```

NIT

`$splitCharNodeAroundTargets` ~60-line docstring is thorough but heavy — **optional** trim, not a defect (each block maps to a distinct tested behavior).
___
*[`packages/platform/src/editor/adaptors/usj-marker-action.utils.ts` line 600 at r1](https://reviewable.io/reviews/eten-tech-foundation/scripture-editors/529#-OzRmrbJBQ0VJbnO7nJ1:-OzRmrbJBQ0VJbnO7nJ2:blgv9bg) ([raw file](https://github.com/eten-tech-foundation/scripture-editors/blob/92ce3882a5832721da34d9dd7383a64759c0323d/packages/platform/src/editor/adaptors/usj-marker-action.utils.ts#L600)):*
> ```XML
>  * is left as a documented limitation rather than attempted here.
>  *
>  * Refuses — partial coverage of a nested CharNode: transitive coverage (above) only decides whether
> ```

NIT **Severity: Low** 

The nested partial-coverage refuse is a silent no-op with no user feedback_ — compounds #8. Reasonable while the product question is open; revisit alongside the return-value decision.
___
*[`packages/platform/src/editor/adaptors/usj-marker-action.utils.ts` line 631 at r1](https://reviewable.io/reviews/eten-tech-foundation/scripture-editors/529#-OzRhZPtEo7ljQOVi011:-OzRhZPtEo7ljQOVi012:b-4ow7g4) ([raw file](https://github.com/eten-tech-foundation/scripture-editors/blob/92ce3882a5832721da34d9dd7383a64759c0323d/packages/platform/src/editor/adaptors/usj-marker-action.utils.ts#L631)):*
> ```XML
>       coveredIndexes.push(index);
>     } else if ($isElementNode(child) && targetNodes.some((target) => child.isParentOf(target))) {
>       // Nothing has been mutated yet, so refusing here leaves the document exactly as it was.
> ```

NIT **Severity: Low** 

_The comment here is inaccurate and will mislead the next maintainer._ "Nothing has been mutated yet, so refusing here leaves the document exactly as it was" is only true within this function — the caller already ran `splitText` before calling it (see #1). Fix the comment as part of #1.
___
*[`packages/platform/src/editor/adaptors/usj-marker-action.utils.ts` line 632 at r1](https://reviewable.io/reviews/eten-tech-foundation/scripture-editors/529#-OzRhS134F1--dCGyAb9:-OzRhS134F1--dCGyAbA:bdqfgny) ([raw file](https://github.com/eten-tech-foundation/scripture-editors/blob/92ce3882a5832721da34d9dd7383a64759c0323d/packages/platform/src/editor/adaptors/usj-marker-action.utils.ts#L632)):*
> ```XML
>     } else if ($isElementNode(child) && targetNodes.some((target) => child.isParentOf(target))) {
>       // Nothing has been mutated yet, so refusing here leaves the document exactly as it was.
>       if (!$isFullyCoveredByTargets(child, targetKeys)) return undefined;
> ```

**Severity: Medium**

 _The "refuse" path isn't actually a no-op — it dirties the document and moves the selection before it decides to refuse._ When a range selection partially covers a nested CharNode and the outer marker is targeted, `$hasCharNodeToRemove` returns true, so the caller loop runs `splitText` (dirtying leaves → a spurious undo entry and an empty collab delta) and then `setTextNodeRange`, and only _after_ that does `$splitCharNodeAroundTargets` return `undefined`. Lexical re-merges so the final tree looks unchanged, which is why the "refuses…" test (which asserts only final structure) stays green. This is exactly the defect class you fixed for the no-match path with `$hasCharNodeToRemove`; it just isn't applied to the refuse case. Suggest deciding coverage in a read-only pre-pass (or extending the existing guard) so no `splitText`/restore runs when the answer is "refuse." It's reachable from a direct API call even though today's UI avoids it.
___
*[`packages/platform/src/editor/adaptors/usj-marker-action.utils.ts` line 632 at r1](https://reviewable.io/reviews/eten-tech-foundation/scripture-editors/529#-OzRpbwW4s3L1sQkXJpo:-OzRpbwW4s3L1sQkXJpp:b-7yenf) ([raw file](https://github.com/eten-tech-foundation/scripture-editors/blob/92ce3882a5832721da34d9dd7383a64759c0323d/packages/platform/src/editor/adaptors/usj-marker-action.utils.ts#L632)):*
> ```XML
>     } else if ($isElementNode(child) && targetNodes.some((target) => child.isParentOf(target))) {
>       // Nothing has been mutated yet, so refusing here leaves the document exactly as it was.
>       if (!$isFullyCoveredByTargets(child, targetKeys)) return undefined;
> ```

NIT

`$isFullyCoveredByTargets` is vacuously `true` for an empty nested element (`getAllTextNodes()` empty → `.every()` true); unreachable today (placeholder + `isParentOf` guard)
___
*[`packages/platform/src/editor/adaptors/usj-marker-action.utils.ts` line 636 at r1](https://reviewable.io/reviews/eten-tech-foundation/scripture-editors/529#-OzRmyH4BmqcmvctqV2I:-OzRmyH4BmqcmvctqV2J:b-olkzjw) ([raw file](https://github.com/eten-tech-foundation/scripture-editors/blob/92ce3882a5832721da34d9dd7383a64759c0323d/packages/platform/src/editor/adaptors/usj-marker-action.utils.ts#L636)):*
> ```XML
>     }
>   }
>   if (coveredIndexes.length === 0) return charNode;
> ```

NIT **Severity: Nit** 

The `coveredIndexes.length === 0` fallback returns `charNode` (full removal), the opposite of this function's "refuse rather than over-remove" contract._ It's unreachable today (the caller guarantees ≥1 covered index), but the follow-on replace/extend work shares this path — prefer `return undefined` so the safe behavior is the default if it ever becomes reachable.
___
*[`packages/platform/src/editor/adaptors/usj-marker-action.utils.ts` line 710 at r1](https://reviewable.io/reviews/eten-tech-foundation/scripture-editors/529#-OzRmOPK4X6me_HKHIAJ:-OzRmOPK4X6me_HKHIAK:bq26vt3) ([raw file](https://github.com/eten-tech-foundation/scripture-editors/blob/92ce3882a5832721da34d9dd7383a64759c0323d/packages/platform/src/editor/adaptors/usj-marker-action.utils.ts#L710)):*
> ```XML
>   // prepend a second one to it.
>   const remainingChildren = charNode.getChildren();
>   if (remainingChildren.length === 0 || charNode.getTextContent() === EMPTY_CHAR_PLACEHOLDER_TEXT) {
> ```

NIT **Severity: Nit** 

In visible mode, a CharNode whose real content is a lone NBSP is indistinguishable from the empty placeholder and gets deleted._ Pre-existing representational ambiguity (placeholder === NBSP, no prepend in visible mode); extremely unlikely in real text. Note-only, no action needed unless you want a guard.
___
*[`packages/platform/src/editor/adaptors/usj-marker-action.utils.ts` line 715 at r1](https://reviewable.io/reviews/eten-tech-foundation/scripture-editors/529#-OzRmHVRDp9kfABWsF-R:-OzRmHVRDp9kfABWsF-S:bawqain) ([raw file](https://github.com/eten-tech-foundation/scripture-editors/blob/92ce3882a5832721da34d9dd7383a64759c0323d/packages/platform/src/editor/adaptors/usj-marker-action.utils.ts#L715)):*
> ```XML
>   }
> 
>   if (viewOptions?.markerMode === "editable")
> ```

NIT **Severity: Low (trigger unconfirmed)** 

The NBSP trim removes only one leading NBSP, which could strand a second one._ Under editable mode each text child gets one NBSP prepended; if Lexical reconciliation merges two adjacent prefixed runs into a single node, `slice(NBSP.length)` leaves one NBSP behind as exporter-visible text. I couldn't confirm a CharNode ever carries two mergeable text children — worth a quick check, and a test if it can.
___
*[`packages/platform/src/editor/adaptors/usj-marker-action.utils.ts` line 715 at r1](https://reviewable.io/reviews/eten-tech-foundation/scripture-editors/529#-OzRnp1F74ZfszKT1b35:-OzRnp1F74ZfszKT1b36:b-i29vjw) ([raw file](https://github.com/eten-tech-foundation/scripture-editors/blob/92ce3882a5832721da34d9dd7383a64759c0323d/packages/platform/src/editor/adaptors/usj-marker-action.utils.ts#L715)):*
> ```XML
>   }
> 
>   if (viewOptions?.markerMode === "editable")
> ```

NIT

NBSP-strip logic duplicated in `editor-usj.adaptor.ts` `createCharMarker`

jolierabideau (commented):
> @jolierabideau made 16 comments and resolved 11 discussions.
> *[Reviewable](https://reviewable.io/reviews/eten-tech-foundation/scripture-editors/529)* status: 10 of 11 files reviewed, 16 unresolved discussions (waiting on katherinejensen00).

___
*[`packages/platform/src/editor/adaptors/usj-marker-action-utils.test.ts` line 340 at r1](https://reviewable.io/reviews/eten-tech-foundation/scripture-editors/529#-OzRobzV8BNpkj8aM8CO:-OzSE6kfDwuUz6BeBY0p:bqd1olj) ([raw file](https://github.com/eten-tech-foundation/scripture-editors/blob/92ce3882a5832721da34d9dd7383a64759c0323d/packages/platform/src/editor/adaptors/usj-marker-action-utils.test.ts#L340)):*

NIT 

Non-NBSP collapsed-caret restore offset unasserted.
</blockquote></details>

Deferred. Real gap — the test asserts structure but not the restored offset. I did add `expect(sutRemoveCharacterMarker(editor)).toBe(true)` here for the new return value, but not the offset assertion.
___
*[`packages/platform/src/editor/adaptors/usj-marker-action-utils.test.ts` line 356 at r1](https://reviewable.io/reviews/eten-tech-foundation/scripture-editors/529#-OzRo3ne17Vt67ml8aVT:-OzSFA3J1XDyM32NKHUt:bcatrqc) ([raw file](https://github.com/eten-tech-foundation/scripture-editors/blob/92ce3882a5832721da34d9dd7383a64759c0323d/packages/platform/src/editor/adaptors/usj-marker-action-utils.test.ts#L356)):*

NIT

Test-fixture duplication (~13 paragraph fixtures, 23 unwrap-and-assert prologues).
</blockquote></details>

Deferred. Agreed the ~13 near-identical paragraph fixtures and unwrap-and-assert prologues want extracting into helpers. Pure test refactor, no behavior at stake.
___
*[`packages/platform/src/editor/adaptors/usj-marker-action-utils.test.ts` line 493 at r1](https://reviewable.io/reviews/eten-tech-foundation/scripture-editors/529#-OzRoJhk1B1tJ65S4Kv5:-OzSFIDD1Gq5wkgKpkMo:b-vb1xgm) ([raw file](https://github.com/eten-tech-foundation/scripture-editors/blob/92ce3882a5832721da34d9dd7383a64759c0323d/packages/platform/src/editor/adaptors/usj-marker-action-utils.test.ts#L493)):*

NIT **Severity: Low** 

The cid copy is tested, but the re-merge it exists to enable is not._ `$charNodeTransform` lives in `CharNodePlugin` (shared-react), which `createBasicTestEnvironment` doesn't install — so nothing exercises the split halves actually re-merging. Worth one test in an environment that registers the plugin.
</blockquote></details>

Deferred. Correct that `createBasicTestEnvironment` never installs `CharNodePlugin`, so `$charNodeTransform` never runs and nothing proves the halves actually re-merge. Wants one test in an environment that registers the plugin.
___
*[`packages/platform/src/editor/adaptors/usj-marker-action-utils.test.ts` line 953 at r1](https://reviewable.io/reviews/eten-tech-foundation/scripture-editors/529#-OzRon9FE58QmkCHe1p-:-OzSFLG7EqM9zPTh2A63:bo5a8xq) ([raw file](https://github.com/eten-tech-foundation/scripture-editors/blob/92ce3882a5832721da34d9dd7383a64759c0323d/packages/platform/src/editor/adaptors/usj-marker-action-utils.test.ts#L953)):*

NIT

Backward × NBSP-trim restore never co-tested; visible-mode selection restore unasserted.
</blockquote></details>

Deferred. Both dimensions are covered individually but never together, and visible-mode selection restore is unasserted.
___
*[`packages/platform/src/editor/adaptors/usj-marker-action-utils.test.ts` line 1072 at r1](https://reviewable.io/reviews/eten-tech-foundation/scripture-editors/529#-OzRot7W6Mk8kRWqxKAI:-OzSFNmW64X23uWEB1xt:b-i0rlcw) ([raw file](https://github.com/eten-tech-foundation/scripture-editors/blob/92ce3882a5832721da34d9dd7383a64759c0323d/packages/platform/src/editor/adaptors/usj-marker-action-utils.test.ts#L1072)):*

NIT

Known-limitation visible-mode test may not trip an adaptor-level fix (pins in-editor text)
</blockquote></details>

Deferred. Agreed the assertion may not trip if the real fix lands at the adaptor level. Related to #18 — probably worth fixing together with the tracked follow-up.
___
*[`packages/platform/src/editor/adaptors/usj-marker-action.utils.ts` line 88 at r1](https://reviewable.io/reviews/eten-tech-foundation/scripture-editors/529#-OzRm7-E5Q5ebN9jF5y2:-OzSFa6P0i_zODPvc1Qe:bb44nq5) ([raw file](https://github.com/eten-tech-foundation/scripture-editors/blob/92ce3882a5832721da34d9dd7383a64759c0323d/packages/platform/src/editor/adaptors/usj-marker-action.utils.ts#L88)):*

NIT **Severity: Low** 

A project's configured extra char marker is removable but not insertable._ `isUsjMarkerSupported` (used by `insertMarker`) calls `CharNode.isValidMarker(marker)` without `extraValidMarkers`, while your new `isCharacterMarkerSupported` honors them. You documented the asymmetry, but the real inconsistency is on the insert side — worth threading extras into `isUsjMarkerSupported` in a follow-up so the two agree.
</blockquote></details>

Deferred as a follow-up, as you suggested. Confirmed the asymmetry: a project's configured extra char marker is removable but not insertable, and the inconsistency is genuinely on the insert side. This is one of the three deferred items I'd actually track.
___
*[`packages/platform/src/editor/adaptors/usj-marker-action.utils.ts` line 394 at r1](https://reviewable.io/reviews/eten-tech-foundation/scripture-editors/529#-OzRnVAuAdsTc9cli6X4:-OzSFjzr2HLu3Gtg4rve:b-aar9l6) ([raw file](https://github.com/eten-tech-foundation/scripture-editors/blob/92ce3882a5832721da34d9dd7383a64759c0323d/packages/platform/src/editor/adaptors/usj-marker-action.utils.ts#L394)):*

NIT 

Duplicated collapsed-vs-range restore blocks + multi-paragraph comments.
</blockquote></details>

Deferred. Didn't want to refactor the restore blocks in the same change as the refuse-path fix.
___
*[`packages/platform/src/editor/adaptors/usj-marker-action.utils.ts` line 430 at r1](https://reviewable.io/reviews/eten-tech-foundation/scripture-editors/529#-OzRnhVrB1mHtJgcTMUU:-OzSFpsB-m8iTWAW20Kc:ba2eig2) ([raw file](https://github.com/eten-tech-foundation/scripture-editors/blob/92ce3882a5832721da34d9dd7383a64759c0323d/packages/platform/src/editor/adaptors/usj-marker-action.utils.ts#L430)):*

NIT 

Redundant double ancestor-walk (`$hasCharNodeToRemove` then the loop re-walks)
</blockquote></details>

Deferred, and worth flagging: the #22 fix makes the pre-pass do _more_ work, not less, since it now also evaluates nested coverage. I judged a correct no-op worth the second walk — selections are small and this is not on a hot path — but the duplication is real and grew.
___
*[`packages/platform/src/editor/adaptors/usj-marker-action.utils.ts` line 469 at r1](https://reviewable.io/reviews/eten-tech-foundation/scripture-editors/529#-OzRpEdo8pCd7GC6gSBx:-OzSFt7UCzizrV3zXSqL:b-7or5mz) ([raw file](https://github.com/eten-tech-foundation/scripture-editors/blob/92ce3882a5832721da34d9dd7383a64759c0323d/packages/platform/src/editor/adaptors/usj-marker-action.utils.ts#L469)):*

NIT **Severity: Nit (please verify)** 

Please confirm the `$unwrapNode` → `TextNode.replace()` wording against the pinned Lexical 0.43.0 source._ One finder believed 0.43.0's `$unwrapNode` is caret-splice-based and never calls `replace()`; I couldn't verify locally (no `node_modules`). The manual `$getSelection()` re-fetch is correct regardless (the NBSP `setTextContent` trim genuinely needs it), so this is only about the comment's stated mechanism.
</blockquote></details>

Verified against the pinned 0.43.0 source. `$unwrapNode` is `$rewindSiblingCaret($getSiblingCaret(node,'next')).splice(1, node.getChildren())` — so caret-based, as the finder said. But `NodeCaret.splice` then calls `target.replace(node)` (`Lexical.dev.mjs:12858`) where `target` is the CharNode, and `LexicalNode.replace` (`:4037-4080`) does clone the active selection and `$setSelection` the clone. So `replace()` _is_ reached and the re-fetch _is_ required — it's the CharNode's `ElementNode.replace()`, not `TextNode.replace()`. Comment corrected to name the actual path.
___
*[`packages/platform/src/editor/adaptors/usj-marker-action.utils.ts` line 588 at r1](https://reviewable.io/reviews/eten-tech-foundation/scripture-editors/529#-OzRmm4_2-N_e-Jkyo_M:-OzSFzQb7-w4I0ocg8BW:b-jqynyp) ([raw file](https://github.com/eten-tech-foundation/scripture-editors/blob/92ce3882a5832721da34d9dd7383a64759c0323d/packages/platform/src/editor/adaptors/usj-marker-action.utils.ts#L588)):*

NIT **Severity: Low (documented)** 

Interior partial coverage strands an unpaired marker half in marker-display views._ PresentaNIT

tion-only and self-corrects on reload; the real fix (regenerate paired open/close markers per surviving clone) is adaptor-level. Fine to keep as the documented limitation; consider a tracked follow-up referenced from the docstring.
</blockquote></details>

Deferred, keeping it as the documented limitation as you suggested. Agreed the real fix is adaptor-level (regenerating paired open/close markers per surviving clone). Second of the three I'd actually track, together with #10.
___
*[`packages/platform/src/editor/adaptors/usj-marker-action.utils.ts` line 588 at r1](https://reviewable.io/reviews/eten-tech-foundation/scripture-editors/529#-OzRpmJICri-_Sa7Ppbn:-OzSG19aBFUIbwW8crJ2:bhduini) ([raw file](https://github.com/eten-tech-foundation/scripture-editors/blob/92ce3882a5832721da34d9dd7383a64759c0323d/packages/platform/src/editor/adaptors/usj-marker-action.utils.ts#L588)):*

NIT

`$splitCharNodeAroundTargets` ~60-line docstring is thorough but heavy — **optional** trim, not a defect (each block maps to a distinct tested behavior).
</blockquote></details>

Deferred. Taking the "optional, not a defect" framing — and this change added to it rather than trimming it.
___
*[`packages/platform/src/editor/adaptors/usj-marker-action.utils.ts` line 632 at r1](https://reviewable.io/reviews/eten-tech-foundation/scripture-editors/529#-OzRhS134F1--dCGyAb9:-OzSG5q4EXEyMebc-RhV:b-896fix) ([raw file](https://github.com/eten-tech-foundation/scripture-editors/blob/92ce3882a5832721da34d9dd7383a64759c0323d/packages/platform/src/editor/adaptors/usj-marker-action.utils.ts#L632)):*

**Severity: Medium**

 _The "refuse" path isn't actually a no-op — it dirties the document and moves the selection before it decides to refuse._ When a range selection partially covers a nested CharNode and the outer marker is targeted, `$hasCharNodeToRemove` returns true, so the caller loop runs `splitText` (dirtying leaves → a spurious undo entry and an empty collab delta) and then `setTextNodeRange`, and only _after_ that does `$splitCharNodeAroundTargets` return `undefined`. Lexical re-merges so the final tree looks unchanged, which is why the "refuses…" test (which asserts only final structure) stays green. This is exactly the defect class you fixed for the no-match path with `$hasCharNodeToRemove`; it just isn't applied to the refuse case. Suggest deciding coverage in a read-only pre-pass (or extending the existing guard) so no `splitText`/restore runs when the answer is "refuse." It's reachable from a direct API call even though today's UI avoids it.
</blockquote></details>

Done.
___
*[`packages/platform/src/editor/adaptors/usj-marker-action.utils.ts` line 632 at r1](https://reviewable.io/reviews/eten-tech-foundation/scripture-editors/529#-OzRpbwW4s3L1sQkXJpo:-OzSG9PQCa7yIYvGC896:b-ykrft6) ([raw file](https://github.com/eten-tech-foundation/scripture-editors/blob/92ce3882a5832721da34d9dd7383a64759c0323d/packages/platform/src/editor/adaptors/usj-marker-action.utils.ts#L632)):*

NIT

`$isFullyCoveredByTargets` is vacuously `true` for an empty nested element (`getAllTextNodes()` empty → `.every()` true); unreachable today (placeholder + `isParentOf` guard)
</blockquote></details>

Deferred as note-only, agreeing it's unreachable today (placeholder + `isParentOf` guard).
___
*[`packages/platform/src/editor/adaptors/usj-marker-action.utils.ts` line 710 at r1](https://reviewable.io/reviews/eten-tech-foundation/scripture-editors/529#-OzRmOPK4X6me_HKHIAJ:-OzSGPJU-OJKUmCqQJ0R:bh8trzc) ([raw file](https://github.com/eten-tech-foundation/scripture-editors/blob/92ce3882a5832721da34d9dd7383a64759c0323d/packages/platform/src/editor/adaptors/usj-marker-action.utils.ts#L710)):*

NIT **Severity: Nit** 

In visible mode, a CharNode whose real content is a lone NBSP is indistinguishable from the empty placeholder and gets deleted._ Pre-existing representational ambiguity (placeholder === NBSP, no prepend in visible mode); extremely unlikely in real text. Note-only, no action needed unless you want a guard.
</blockquote></details>

Deferred, no guard added, per your "note-only unless you want one." Pre-existing representational ambiguity.
___
*[`packages/platform/src/editor/adaptors/usj-marker-action.utils.ts` line 715 at r1](https://reviewable.io/reviews/eten-tech-foundation/scripture-editors/529#-OzRmHVRDp9kfABWsF-R:-OzSGS8V43VFnj169viJ:b-cvdvm6) ([raw file](https://github.com/eten-tech-foundation/scripture-editors/blob/92ce3882a5832721da34d9dd7383a64759c0323d/packages/platform/src/editor/adaptors/usj-marker-action.utils.ts#L715)):*

NIT **Severity: Low (trigger unconfirmed)** 

The NBSP trim removes only one leading NBSP, which could strand a second one._ Under editable mode each text child gets one NBSP prepended; if Lexical reconciliation merges two adjacent prefixed runs into a single node, `slice(NBSP.length)` leaves one NBSP behind as exporter-visible text. I couldn't confirm a CharNode ever carries two mergeable text children — worth a quick check, and a test if it can.
</blockquote></details>

Deferred but **not** dismissed, and I want to be straight that I did _not_ confirm or rule out the trigger. One thing that makes me less comfortable than the other nits: commit `4c61942c` ("don't insert a space between adjacent text nodes of the same run") suggests adjacent text children within a run do occur, which is the precondition for the merge you describe. If it can happen, `slice(NBSP.length)` leaves exporter-visible NBSP in the USJ — a real data bug, not presentation. Third of the three I'd track, and the highest priority of them.
___
*[`packages/platform/src/editor/adaptors/usj-marker-action.utils.ts` line 715 at r1](https://reviewable.io/reviews/eten-tech-foundation/scripture-editors/529#-OzRnp1F74ZfszKT1b35:-OzSGTyFDN6CDR_vcT-t:bx8x7e9) ([raw file](https://github.com/eten-tech-foundation/scripture-editors/blob/92ce3882a5832721da34d9dd7383a64759c0323d/packages/platform/src/editor/adaptors/usj-marker-action.utils.ts#L715)):*

NIT

NBSP-strip logic duplicated in `editor-usj.adaptor.ts` `createCharMarker`
</blockquote></details>

Deferred. Agreed the prepend in `usj-editor.adaptor.ts:361-363` and the strip here should share a helper so they can't drift — relevant to #26.

katherinejensen00 (approved):
> @katherinejensen00 reviewed 11 files and all commit messages, and resolved 16 discussions.
> *[Reviewable](https://reviewable.io/reviews/eten-tech-foundation/scripture-editors/529)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved.

PR #530, opened 2026-08-06
PT-3780: Put the insertion point inside a marker inserted in a footnote

### LINKED ISSUE #1: add volta
- JS tool manager

**Note:** recreates Pull Request history from previous repository.

katherinejensen00:
https://github.com/user-attachments/assets/c30bbf23-72b7-4247-9c66-5c3e0933a264

## What & why

Fixes [PT-3780](https://paratextstudio.atlassian.net/browse/PT-3780). When inserting a marker while editing a footnote, the insertion point (caret) landed **before** the new marker instead of inside it. Investigation surfaced a second, related defect in the same code path: when the caret was inside an existing footnote char, the new marker **escaped the note** into the surrounding paragraph.

## The fix

All in the NoteNode-insertion branch of `getUsjMarkerAction` ([usj-marker-action.utils.ts](packages/platform/src/editor/adaptors/usj-marker-action.utils.ts)):

1. **Caret inside the new marker.** The branch now sets the selection (via `selectEnd` on the new content char) — every sibling insertion branch already did this; this one was the outlier that didn't.
2. **Marker stays in the note.** The branch now also fires when the caret is inside one of the note's CharNodes (previously only when it was on the note's own spacer text), inserting the new marker as a sibling **within** the note so it can't escape into the paragraph.
3. **No caret-steal via a double spacer.** The trailing spacer is added only when one doesn't already follow. Otherwise, inserting between a char and its existing spacer left two adjacent spacers, which a NoteNode transform later collapsed with a `selectEnd` that yanked the caret back out onto the spacer — defeating fix #1 end-to-end.

## Testing

- **Unit (adaptor level):** caret on a note spacer, and caret inside an existing footnote char (asserts the marker stays in the note, the existing text is preserved, and the caret lands inside the new marker).
- **End-to-end with the real plugins/transforms mounted:** caret in footnote text, caret on a note spacer, and the **marker-menu option-click** path. These are the tests that catch the double-spacer caret-steal, which is invisible to the plugin-less adaptor tests.

Note: the floating typeahead menu itself can't be opened/positioned in jsdom (floating-ui + portals are browser-only), so the popover test drives the real menu's list component and its option→action seam — wired to the same `getUsjMarkerAction` the editor gives the menu — rather than the keyboard-open + positioning. True browser-level menu behavior would need an e2e harness (not present in this repo).

## Follow-up

[PT-4315](https://paratextstudio.atlassian.net/browse/PT-4315) tracks unifying caret placement across all of `getUsjMarkerAction`'s branches so a future branch can't silently regress the caret again.

- - -
This change is [Reviewable](https://reviewable.io/reviews/eten-tech-foundation/scripture-editors/530)

### DIFF
--- a/packages/platform/src/editor/Editor.test.tsx
+++ b/packages/platform/src/editor/Editor.test.tsx
@@ -7,8 +7,25 @@ import Editorial from "../Editorial";
 import { flushQueuedEvents } from "./editor-test.utils";
 import { ContentJsonPath, Usj } from "@eten-tech-foundation/scripture-utilities";
 import { useLexicalComposerContext } from "@lexical/react/LexicalComposerContext";
-import { act, render } from "@testing-library/react";
-import { KEY_DOWN_COMMAND, LexicalEditor } from "lexical";
+// Deep import: the marker-menu list component isn't exposed from shared-react's package entry.
+// eslint-disable-next-line @nx/enforce-module-boundaries
+import { NodeSelectionMenu, OptionItem } from "../../../../libs/shared-react/src/plugins/NodesMenu";
+import { getUsjMarkerAction } from "./adaptors/usj-marker-action.utils";
+import { act, fireEvent, render } from "@testing-library/react";
+import {
+  $createPoint,
+  $createRangeSelection,
+  $getRoot,
+  $getSelection,
+  $isElementNode,
+  $isRangeSelection,
+  $isTextNode,
+  $setSelection,
+  KEY_DOWN_COMMAND,
+  LexicalEditor,
+  LexicalNode,
+} from "lexical";
+import { $isCharNode, $isNoteNode } from "shared";
 import { createRef, RefObject, useEffect } from "react";
 import { vi } from "vitest";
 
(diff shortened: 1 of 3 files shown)


captaincrazybro:
<details><summary><i>Previously, captaincrazybro (Levi Wenger) wrote…</i></summary><blockquote>

In the fix, when the caret splits an existing footnote `CharNode` mid-text, the tail half is rebuilt as:

```ts
const tailChar = $createCharNode(caretChar.getMarker(), caretChar.getUnknownAttributes());
tailChar.append(...charTail);
```

This only copies `marker` + `unknownAttributes`. It drops `indent`, `format`, `style`, `direction`, and — critically — **`charIdState` (`cid`)**.

I know this matters because `main` already hit this exact bug elsewhere in the same file and fixed it with a helper, `$createCharNodeLike` (built on Lexical's `$copyNode`), whose doc comment says almost verbatim what applies here:

> "Don't hand-roll this: `CharNode.insertNewAfter` and `ParaNode.insertNewAfter` model a manual copy... The cid matters: without it `$charNodeTransform`'s `$hasSameCharAttributes` check would refuse to re-merge the halves later."

`$hasSameCharAttributes` (`node.utils.ts:316`) requires both sides to have the _same_ `cid` or _both_ have none — a front half that keeps the original `cid` and a tail half with none will never satisfy that. Practical consequence: if the inserted marker is later removed and the CharNodePlugin's adjacent-merge transform tries to recombine the two `ft` halves, it won't — the footnote stays permanently split into two char nodes, and if `cid` correlates to anything (e.g. a comment anchor), that association is silently dropped from the tail half. Any direction/format/style set on the original char is also silently lost.

**Suggestion**: after rebasing, reuse `$copyNode`/`$createCharNodeLike` (or the equivalent full clone) for `tailChar` instead of the hand-rolled `$createCharNode(marker, unknownAttributes)`.
</blockquote></details>

I'm not sure this is an actual problem but this is something Claude thought was a big problem.

PR #531, opened 2026-08-07
fix(platform): apply annotations issued while a load is in flight

### LINKED ISSUE #515: setAnnotation before the first content commit silently no-ops
From the Fluent RTE PoC (fluent-web#375, PR fluent-web#376).

Calling `editorRef.setAnnotation(...)` immediately after mounting `Editorial` with `defaultUsj` (or immediately after `setUsj`) silently does nothing: Lexical hasn't committed the content yet, `$getRangeFromUsjSelection` finds no nodes, the plugin logs to the (optional) logger and returns. No error surfaces to the caller, and there is no signal to await.

In Fluent we had to build a retry loop that watches the DOM for `mark` elements and re-applies until they land — workable, but fragile and easy for other consumers to miss (the symptom is just "annotations sometimes don't show").

### Proposal (either would do)

- queue annotation calls internally until the first content commit, then apply them; or
- expose a "content ready" signal (callback or promise) so hosts know when the annotation API is safe to use.

henrique221:
Fixes #515.

## Demo

A 2-minute screencast (with narration) is attached in the comments. It runs a small consumer app that behaves like a checking tool: a single line in a mount effect asks for one word to be underlined, with no waiting and no retry. The same app is shown against a build of `main` — badge reads "no annotation applied", no underline, nothing thrown or logged — and then against this branch, where the annotation lands. The only thing that differs between the two halves is the package build.

https://github.com/user-attachments/assets/a83149bf-2ec0-4643-a9d5-330367b08d90

## What was wrong

`setAnnotation` did nothing if it was called before the editor's first content commit — from a consumer's mount effect, or in the same tick as `setUsj`. `LoadStatePlugin` commits USJ from a `queueMicrotask`, so at that moment the document is still the empty initial state, and the annotation was lost in one of three ways:

1. `annotationRef.current` is still `null` that early, so `Editor`'s own optional-chained call vanished without even reaching the plugin;
2. `$getRangeFromUsjSelection` couldn't resolve the range, so the plugin logged `"Failed to find start or end node of the annotation."` to an *optional* logger and returned;
3. a range that *did* resolve was then thrown away wholesale by the load's `editor.setEditorState()`.

None of them surface to the caller — no throw, no return value — so it presents as "annotations sometimes just don't show". In Fluent we ended up polling the DOM for `<mark>` elements and re-applying up to three times; this removes the need for that.

## The change

`LoadStatePlugin` gains one optional prop, `onLoadingChange?: (isLoading: boolean) => void`, reporting `true` when a load starts and `false` once it settles — committed, skipped, or failed. Calls are always balanced (a `finally` covers the early return and any throw).

`Editor` uses that to defer `setAnnotation` / `removeAnnotation` until the load in flight has settled. **When nothing is loading they run immediately, exactly as before.** `setUsj` also marks a load as requested, because React hasn't re-rendered yet at that point, so the plugin hasn't reported the load.

Removal is gated too: if a queued `setAnnotation` is about to run, an un-gated `removeAnnotation` would be a no-op and the mark would survive.

## The subtle part, worth a careful look

The settle signal is deliberately **not** the update's `onUpdate` option, which is the obvious choice and is wrong here. `setEditorState` runs an intermediate `$commitPendingUpdates` of the *outgoing* state from inside the update, and that commit drains Lexical's deferred callbacks — so `onUpdate` fires while the old document is still installed. Verified against lexical 0.43 with a standalone probe:

```
onUpdate option:                          settle sees OLD
microtask queued after editor.update():   settle sees NEW
```

Lexical schedules the real commit with its own `queueMicrotask` before `update()` returns, so a microtask queued after it lands strictly later. There's a comment at the call site so nobody "simplifies" it back.

## Not swallowing real failures

A queued operation runs at the end of the load it was queued behind — never waiting on a *future* load — so a range that can never resolve still fails loudly, one load later, with the same log. That's pinned by a test.

## Verification

New `annotation-before-load.test.tsx` drives the real `Editor` through its ref, the way an app does, and covers the mount-effect case, the `setUsj` case, and the never-resolves guard. With the fix reverted, the first two fail and the guard still passes.

```
shared            123 passed
platform-editor   194 passed
shared-react     1323 passed
lint, typecheck   clean
extract-api       API report identical (no public API change)
```

Backward compatible: `EditorRef`, `EditorProps` and `MarginalRef` are untouched, the new plugin prop is optional (scribe doesn't pass it), and existing tests that `await` the mount are unaffected because the gate is already open by the time they use the ref.

Happy to adjust if this collides with the editor work your team has planned — it's confined to the load/annotation path.

- - -
This change is [Reviewable](https://reviewable.io/reviews/eten-tech-foundation/scripture-editors/531)

### DIFF
--- a/libs/shared-react/src/plugins/usj/LoadStatePlugin.tsx
+++ b/libs/shared-react/src/plugins/usj/LoadStatePlugin.tsx
@@ -20,6 +20,7 @@ export function LoadStatePlugin<TLogger extends LoggerBasic>({
   nodeOptions,
   editorAdaptor,
   viewOptions,
+  onLoadingChange,
   logger,
 }: {
   scripture?: unknown;
@@ -27,6 +28,16 @@ export function LoadStatePlugin<TLogger extends LoggerBasic>({
   nodeOptions?: NodeOptions;
   editorAdaptor: EditorAdaptor;
   viewOptions?: unknown;
+  /**
+   * Called `true` when a load starts and `false` once it settles — committed, skipped (nothing to
+   * serialize), or failed. Calls are always balanced, so a caller may count them, and `false` is
+   * never reported before the loaded document is live. Use it to defer work that has to address
+   * the loaded content.
+   *
+   * Must be referentially stable (e.g. `useCallback` with stable dependencies): it is a dependency
+   * of the load effect, so a fresh identity each render would reload the document each render.
+   */
+  onLoadingChange?: (isLoading: boolean) => void;
   logger?: TLogger;
 }): null {
   const [editor] = useLexicalComposerContext();
@@ -40,32 +51,63 @@ export function LoadStatePlugin<TLogger extends LoggerBasic>({
     // otherwise fall back to the prop value
     const currentScripture = scriptureRef?.current ?? scripture;
 
-    editorAdaptor.reset?.();
-    const serializedEditorState = editorAdaptor.serializeEditorState(currentScripture, viewOptions);
-    if (serializedEditorState == null) {
-      logger?.warn(
-        "LoadStatePlugin: serializedEditorState was null or undefined. Skipping editor update.",
-      );
-      return;
-    }
+    onLoadingChange?.(true);
+    // Every exit from this effect must report `false` exactly once, including a throw out of the
+    // adaptor — otherwise a caller that gates work on the load waits forever.
+    let settled = false;
+    const settle = () => {
+      if (settled) return;
+      settled = true;
+      onLoadingChange?.(false);
+    };
+    // Set once the commit path owns settling (it reports from its own microtask, after the
+    // loaded document is live); until then the `finally` below is responsible.
+    let settlesAfterCommit = false;
 
     try {
+      editorAdaptor.reset?.();
+      const serializedEditorState = editorAdaptor.serializeEditorState(
+        currentScripture,
+        viewOptions,
+      );
+      if (serializedEditorState == null) {
+        logger?.warn(
+          "LoadStatePlugin: serializedEditorState was null or undefined. Skipping editor update.",
+        );
+        return;
+      }
+
       const editorState = editor.parseEditorState(serializedEditorState);
+      settlesAfterCommit = true;
       // Use queueMicrotask to defer the editor update outside of React's lifecycle,
       // preventing flushSync warnings when this is triggered by a parent component update
       queueMicrotask(() => {
-        editor.update(
-          () => {
-            editor.setEditorState(editorState);
-            editor.dispatchCommand(CLEAR_HISTORY_COMMAND, undefined);
-          },
-          { tag: EXTERNAL_USJ_MUTATION_TAG },
-        );
+        try {
+          editor.update(
+            () => {
+              editor.setEditorState(editorState);
+              editor.dispatchCommand(CLEAR_HISTORY_COMMAND, undefined);
+            },
+            { tag: EXTERNAL_USJ_MUTATION_TAG },
+          );
+          // Deliberately NOT the update's `onUpdate` option: `setEditorState` runs an intermediate
+          // commit of the OUTGOING state from inside the update, and that commit drains Lexical's
+          // deferred callbacks — so `onUpdate` fires while the old document is still installed.
+          // Lexical schedules the real commit with its own microtask before `update` returns, so a
+          // microtask queued here lands strictly after the loaded document is live.
+          queueMicrotask(settle);
+        } catch {
+          logger?.error("LoadStatePlugin: error setting editor state.");
+          settle();
+        }
       });
     } catch {
       logger?.error("LoadStatePlugin: error parsing or setting editor state.");
+    } finally {
+      // Covers the early return and any throw; the committed path settles from its own microtask.
+      if (!settlesAfterCommit) settle();
     }
-  }, [editor, editorAdaptor, logger, scripture, scriptureRef, viewOptions]);
+  }, [editor, editorAdaptor, logger, onLoadingChange, scripture, scriptureRef, viewOptions]);
 
   return null;
 }
--- a/packages/platform/src/editor/Editor.tsx
+++ b/packages/platform/src/editor/Editor.tsx
@@ -7,6 +7,7 @@ import {
   isUsjMarkerSupported,
 } from "./adaptors/usj-marker-action.utils";
 import { EditorOptions, EditorProps, EditorRef } from "./editor.model";
+import { useLoadGate } from "./use-load-gate.hook";
 import editorTheme from "./editor.theme";
 import { ActiveTextPlugin } from "./ActiveTextPlugin";
 import { ParaMarkerPrefixGuardPlugin } from "./ParaMarkerPrefixGuardPlugin";
@@ -143,6 +144,8 @@ const Editor = forwardRef(function Editor<TLogger extends LoggerBasic>(
   const [usj, setUsj] = useState(defaultUsj);
   const [loadTrigger, setLoadTrigger] = useState(0);
   const [contextMarker, setContextMarker] = useState<string>();
+  // Annotations address the loaded document, so they wait for the load in flight (#515).
+  const { handleLoadingChange, noteLoadRequested, runWhenLoaded } = useLoadGate();
 
   const {
     isReadonly = false,
@@ -215,6 +218,10 @@ const Editor = forwardRef(function Editor<TLogger extends LoggerBasic>(
         editedUsjRef.current = incomingUsj;
         // This can happen when using `applyUpdate` since `usj` won't change.
         const shouldForceReload = deepEqual(usj, incomingUsj);
+        // A load is now certain, but React hasn't re-rendered yet, so LoadStatePlugin hasn't
+        // reported it. Without this, an annotation set in this same tick would address the
+        // outgoing document and be discarded by the load (#515).
+        noteLoadRequested();
         setUsj(incomingUsj);
         if (shouldForceReload) setLoadTrigger((prev) => prev + 1);
       }
@@ -287,18 +294,22 @@ const Editor = forwardRef(function Editor<TLogger extends LoggerBasic>(
         onMouseLeave = fourth.onMouseLeave;
       }
 
-      annotationRef.current?.setAnnotation(
-        selection,
-        externalTypedMarkType(type),
-        id,
-        onClick,
-        onRemove,
-        onMouseEnter,
-        onMouseLeave,
+      runWhenLoaded(() =>
+        annotationRef.current?.setAnnotation(
+          selection,
+          externalTypedMarkType(type),
+          id,
+          onClick,
+          onRemove,
+          onMouseEnter,
+          onMouseLeave,
+        ),
       );
     },
     removeAnnotation(type, id) {
-      annotationRef.current?.removeAnnotation(externalTypedMarkType(type), id);
+      // Also gated: removing an annotation that a queued call is about to add must happen after
+      // it, not before, or the removal is a no-op and the mark survives.
+      runWhenLoaded(() => annotationRef.current?.removeAnnotation(externalTypedMarkType(type), id));
     },
     formatPara(blockMarker) {
       editorRef.current?.update(() => {
@@ -466,6 +477,7 @@ const Editor = forwardRef(function Editor<TLogger extends LoggerBasic>(
             nodeOptions={nodeOptions}
             editorAdaptor={usjEditorAdaptor}
             viewOptions={viewOptions}
+            onLoadingChange={handleLoadingChange}
             logger={logger}
           />
           <OnSelectionChangePlugin onChange={onSelectionChange} />
--- a/packages/platform/src/editor/annotation-before-load.test.tsx
+++ b/packages/platform/src/editor/annotation-before-load.test.tsx
@@ -1,0 +1,175 @@
+/**
+ * #515: annotations applied before the editor's first content commit were silently lost.
+ *
+ * `LoadStatePlugin` commits USJ from inside a `queueMicrotask`, so a consumer that calls
+ * `setAnnotation` right after mount — or right after `setUsj` — runs while the document is
+ * still the empty initial state. Two ways that lost the annotation:
+ *   a) the range cannot resolve, so `$getRangeFromUsjSelection` returns undefined and the
+ *      plugin only logs "Failed to find start or end node of the annotation.";
+ *   b) the range does resolve against the outgoing document, and the load's
+ *      `editor.setEditorState()` then replaces the whole state, discarding the mark.
+ *
+ * Neither surfaces to the caller, so consumers resorted to polling the DOM for <mark> and
+ * re-applying. These tests drive the real Editor through its ref, exactly as an app does.
+ */
+import Editor from "./Editor";
+import { EditorRef } from "./editor.model";
+import { Usj, USJ_TYPE, USJ_VERSION } from "@eten-tech-foundation/scripture-utilities";
+import { act, render } from "@testing-library/react";
+import { createRef, useEffect, useRef } from "react";
+import { AnnotationRange } from "shared-react";
+import { vi } from "vitest";
+
+const v1Text = "The heavens, the earth, and all their vast array were finished.";
+const v2Text = "On the seventh day God finished his work which he had done.";
+
+const usjGen2: Usj = {
+  type: USJ_TYPE,
+  version: USJ_VERSION,
+  content: [
+    { type: "chapter", marker: "c", number: "2" },
+    {
+      type: "para",
+      marker: "p",
+      content: [
+        { type: "verse", marker: "v", number: "1" },
+        v1Text,
+        { type: "verse", marker: "v", number: "2" },
+        v2Text,
+      ],
+    },
+  ],
+};
+
+/** A second document, to exercise the same race on a later `setUsj` load. */
+const usjGen3: Usj = {
+  type: USJ_TYPE,
+  version: USJ_VERSION,
+  content: [
+    { type: "chapter", marker: "c", number: "3" },
+    {
+      type: "para",
+      marker: "p",
+      content: [{ type: "verse", marker: "v", number: "1" }, v1Text],
+    },
+  ],
+};
+
+const jsonPath = "$.content[1].content[1]";
+const earthStart = v1Text.indexOf("earth");
+const earthRange: AnnotationRange = {
+  start: { jsonPath, offset: earthStart },
+  end: { jsonPath, offset: earthStart + "earth".length },
+};
+
+function createLoggerMock() {
+  return vi.fn();
+}
+
+function markTexts() {
+  return [...document.querySelectorAll("mark")].map((mark) => mark.textContent);
+}
+
+/**
+ * A consumer that annotates from its own mount effect — the realistic shape of this bug. The
+ * effect runs right after the editor commits, while the load microtask is still pending, and it
+ * is the first moment a consumer can legitimately touch the ref.
+ */
+function AnnotatingHost({
+  annotate,
+  logError,
+}: {
+  annotate: (editor: EditorRef) => void;
+  logError: () => void;
+}) {
+  const ref = useRef<EditorRef>(null);
+  useEffect(() => {
+    if (ref.current) annotate(ref.current);
+  }, [annotate]);
+  return (
+    <Editor
+      ref={ref}
+      defaultUsj={usjGen2}
+      logger={{
+        error: logError,
+        warn: createLoggerMock(),
+        info: createLoggerMock(),
+        debug: createLoggerMock(),
+      }}
+    />
+  );
+}
+
+describe("annotating before the first content commit (#515)", () => {
+  it("applies an annotation requested from the consumer's mount effect", async () => {
+    const logError = createLoggerMock();
+
+    await act(async () => {
+      render(
+        <AnnotatingHost
+          logError={logError}
+          annotate={(editor) => editor.setAnnotation(earthRange, "spelling", "a1")}
+        />,
+      );
+    });
+
+    expect(logError).not.toHaveBeenCalled();
+    expect(markTexts()).toEqual(["earth"]);
+  });
+
+  it("applies an annotation requested in the same tick as setUsj", async () => {
+    const logError = createLoggerMock();
+    const ref = createRef<EditorRef>();
+
+    await act(async () => {
+      render(
+        <Editor
+          ref={ref}
+          defaultUsj={usjGen2}
+          logger={{
+            error: logError,
+            warn: createLoggerMock(),
+            info: createLoggerMock(),
+            debug: createLoggerMock(),
+          }}
+        />,
+      );
+    });
+
+    // Swap the document and annotate immediately, without awaiting the reload.
+    await act(async () => {
+      ref.current?.setUsj(usjGen3);
+      ref.current?.setAnnotation(earthRange, "spelling", "a2");
+    });
+
+    expect(logError).not.toHaveBeenCalled();
+    expect(markTexts()).toEqual(["earth"]);
+  });
+
+  it("still reports a range that can never resolve, instead of queueing it forever", async () => {
+    const logError = createLoggerMock();
+
+    await act(async () => {
+      render(
+        <AnnotatingHost
+          logError={logError}
+          annotate={(editor) =>
+            editor.setAnnotation(
+              {
+                start: { jsonPath: "$.content[99].content[0]", offset: 0 },
+                end: { jsonPath: "$.content[99].content[0]", offset: 3 },
+              },
+              "spelling",
+              "bad",
+            )
+          }
+        />,
+      );
+    });
+
+    // Deferring must not swallow genuine failures: once the document is live the annotation is
+    // attempted and fails loudly, exactly as it does today.
+    expect(logError).toHaveBeenCalled();
+    expect(markTexts()).toEqual([]);
+  });
+});
--- a/packages/platform/src/editor/use-load-gate.hook.ts
+++ b/packages/platform/src/editor/use-load-gate.hook.ts
@@ -1,0 +1,62 @@
+import { useCallback, useRef } from "react";
+
+/**
+ * Defers imperative work that has to address the loaded document until the load carrying it has
+ * settled.
+ *
+ * `LoadStatePlugin` commits USJ from a microtask, and `editor.setEditorState()` replaces the whole
+ * state, so an operation issued while a load is in flight either can't resolve its target or is
+ * thrown away by the commit that follows. Callers shouldn't have to know that: they hold a ref and
+ * may reasonably use it the moment they have one.
+ *
+ * When no load is pending the operation runs immediately, i.e. exactly as before.
+ */
+export function useLoadGate() {
+  // A load always follows mount, and this hook runs before LoadStatePlugin's effect reports it.
+  const isLoadRequestedRef = useRef(true);
+  const activeLoadCountRef = useRef(0);
+  const queuedOpsRef = useRef<(() => void)[]>([]);
+
+  const drain = useCallback(() => {
+    const ops = queuedOpsRef.current;
+    queuedOpsRef.current = [];
+    for (const op of ops) op();
+  }, []);
+
+  /** Report a load starting or settling. Calls must be balanced. */
+  const handleLoadingChange = useCallback(
+    (isLoading: boolean) => {
+      if (isLoading) {
+        isLoadRequestedRef.current = false;
+        activeLoadCountRef.current += 1;
+        return;
+      }
+      activeLoadCountRef.current = Math.max(0, activeLoadCountRef.current - 1);
+      if (activeLoadCountRef.current > 0 || isLoadRequestedRef.current) return;
+      // A microtask, not a direct call: this runs from inside the settle path, and deferring keeps
+      // the queued work out of it (and off React's commit phase).
+      queueMicrotask(drain);
+    },
+    [drain],
+  );
+
+  /**
+   * Note that a load is certain but hasn't started yet — e.g. `setUsj` was just called, so React
+   * hasn't re-rendered and `LoadStatePlugin` hasn't reported the load. Without this, work issued
+   * in the same tick as `setUsj` would run against the outgoing document.
+   */
+  const noteLoadRequested = useCallback(() => {
+    isLoadRequestedRef.current = true;
+  }, []);
+
+  /** Run now if the document is settled, otherwise once the in-flight load finishes. */
+  const runWhenLoaded = useCallback((op: () => void) => {
+    if (!isLoadRequestedRef.current && activeLoadCountRef.current === 0) {
+      op();
+      return;
+    }
+    queuedOpsRef.current.push(op);
+  }, []);
+
+  return { handleLoadingChange, noteLoadRequested, runWhenLoaded };
+}


packages/platform/src/editor/use-load-gate.hook.ts
```diff
@@ -0,0 +1,62 @@
...
+  // A load always follows mount, and this hook runs before LoadStatePlugin's effect reports it.
+  const isLoadRequestedRef = useRef(true);
+  const activeLoadCountRef = useRef(0);
+  const queuedOpsRef = useRef<(() => void)[]>([]);
+
+  const drain = useCallback(() => {
+    const ops = queuedOpsRef.current;
+    queuedOpsRef.current = [];
+    for (const op of ops) op();
+  }, []);
```
irahopkinson: **Two separate problems in `drain` — the first makes claim 6 in the description untrue.**

**(a) `drain` never re-checks the gate.** The microtask is scheduled at settle time (line 38) and then runs unconditionally, re-reading `queuedOpsRef.current` at call time. So an op pushed *after* a subsequent `noteLoadRequested()` but *before* the already-scheduled drain fires gets executed against the outgoing document.

The description says:

> A queued operation runs at the end of the load it was queued behind — never waiting on a *future* load

What the code implements is closer to "runs at whatever moment the gate last opened". Suggested fix: re-test `isLoadRequestedRef.current` / `activeLoadCountRef.current` inside `drain` and re-queue if the gate has closed again.

**(b) No per-op isolation.** `for (const op of ops) op();` — any throw from one op aborts the rest of the queue *and*, because this runs inside a bare `queueMicrotask`, becomes an uncaught global error rather than reaching the caller. Before this change a throw propagated synchronously to whoever called `setAnnotation`. Wrapping each `op()` and routing failures to `logger` would keep the queue draining and keep failures visible.

_(For the record: I first suspected `AnnotationPlugin`'s reserved-type throw at `AnnotationPlugin.tsx:176` as a live trigger, but it is not reachable through `Editor` — `externalTypedMarkType` prefixes `external-` and the only reserved type is `internal-comment`. Other throws are still reachable, e.g. `AnnotationPlugin.tsx:43`.)_

packages/platform/src/editor/use-load-gate.hook.ts
```diff
@@ -0,0 +1,62 @@
...
+ * thrown away by the commit that follows. Callers shouldn't have to know that: they hold a ref and
+ * may reasonably use it the moment they have one.
+ *
+ * When no load is pending the operation runs immediately, i.e. exactly as before.
+ */
+export function useLoadGate() {
+  // A load always follows mount, and this hook runs before LoadStatePlugin's effect reports it.
+  const isLoadRequestedRef = useRef(true);
+  const activeLoadCountRef = useRef(0);
+  const queuedOpsRef = useRef<(() => void)[]>([]);
```
irahopkinson, line 24: **Queued ops are silently discarded on unmount.**

`queuedOpsRef` has no cleanup effect and no unmount awareness, and `queueMicrotask(drain)` is not cancellable. Unmounting (or a route change) while a load is in flight drops every queued annotation with no log and no rejection.

That is the same failure class #515 opens with:

> No error surfaces to the caller, and there is no signal to await.

The pre-change code at least logged "Failed to find start or end node of the annotation." Worth a cleanup that surfaces anything still queued at teardown.

packages/platform/src/editor/use-load-gate.hook.ts
```diff
@@ -0,0 +1,62 @@
...
+    const ops = queuedOpsRef.current;
+    queuedOpsRef.current = [];
+    for (const op of ops) op();
+  }, []);
+
+  /** Report a load starting or settling. Calls must be balanced. */
+  const handleLoadingChange = useCallback(
+    (isLoading: boolean) => {
+      if (isLoading) {
+        isLoadRequestedRef.current = false;
```
irahopkinson, line 62: **`isLoadRequestedRef` is a latch only `LoadStatePlugin` can clear — and the failure mode is silence.**

`noteLoadRequested()` sets this `true`; the *only* thing that clears it is `handleLoadingChange(true)`. Today that always follows, so this is correct as written. But if any future change stops the load effect re-firing — exactly the kind of deep-equality memoisation already applied to `viewOptions` and `logger` at `Editor.tsx:168-195` — every `setAnnotation` / `removeAnnotation` queues forever with no warning, no timeout and no escape hatch.

Judgement call, but given the bug being fixed here *is* "annotations silently do not show", a dev-mode warning when the gate stays closed beyond a tick would keep the next regression loud.

packages/platform/src/editor/use-load-gate.hook.ts
```diff
@@ -0,0 +1,62 @@
...
+  }, []);
+
+  /** Run now if the document is settled, otherwise once the in-flight load finishes. */
+  const runWhenLoaded = useCallback((op: () => void) => {
+    if (!isLoadRequestedRef.current && activeLoadCountRef.current === 0) {
+      op();
+      return;
+    }
+    queuedOpsRef.current.push(op);
+  }, []);
```
irahopkinson: **Scope note:** #515 is scoped to the *first* content commit —

> `setAnnotation` before the first content commit silently no-ops

...but this gate applies to *every* load for the lifetime of the component. That is a defensible and cheap generalisation and I am not asking for it to be narrowed — just flagging that it is broader than the issue asked for, and that it changes timing for every later reload (see my note on the un-gated ref methods in `Editor.tsx`).

libs/shared-react/src/plugins/usj/LoadStatePlugin.tsx
```diff
@@ -40,32 +51,63 @@ export function LoadStatePlugin<TLogger extends LoggerBasic>({
...
+              editor.dispatchCommand(CLEAR_HISTORY_COMMAND, undefined);
+            },
+            { tag: EXTERNAL_USJ_MUTATION_TAG },
+          );
+          // Deliberately NOT the update's `onUpdate` option: `setEditorState` runs an intermediate
+          // commit of the OUTGOING state from inside the update, and that commit drains Lexical's
+          // deferred callbacks — so `onUpdate` fires while the old document is still installed.
+          // Lexical schedules the real commit with its own microtask before `update` returns, so a
+          // microtask queued here lands strictly after the loaded document is live.
+          queueMicrotask(settle);
```
irahopkinson: **This is the load-bearing assumption of the whole fix, and this repo's own code documents a case where it does not hold.**

The comment states that Lexical schedules the real commit with its own microtask before `update` returns. That holds for a **top-level** update, but an update issued while another is already in progress is queued and commits later. This repo says so explicitly:

> `Editor.tsx:369` — `// `discrete` so the update runs now rather than being deferred behind an in-progress one, which would leave `didRemove` reporting `false` for a removal that did happen. Same reason `applyUpdate` above uses it.`

`applyUpdate` and `removeCharacterMarker` both pass `{ discrete: true }` for precisely this reason. The load update here does not. On that path `queueMicrotask(settle)` reports `false` **before** the loaded document is live and the gate drains against the outgoing state — the exact bug this PR fixes, reintroduced through the settle signal.

The regression mode is silent and there is no `LoadStatePlugin` test to catch it. Would `{ discrete: true }` here (or asserting the state actually changed before settling) close it?

libs/shared-react/src/plugins/usj/LoadStatePlugin.tsx
```diff
@@ -40,32 +51,63 @@ export function LoadStatePlugin<TLogger extends LoggerBasic>({
...
 
-    editorAdaptor.reset?.();
-    const serializedEditorState = editorAdaptor.serializeEditorState(currentScripture, viewOptions);
-    if (serializedEditorState == null) {
-      logger?.warn(
-        "LoadStatePlugin: serializedEditorState was null or undefined. Skipping editor update.",
-      );
-      return;
-    }
+    onLoadingChange?.(true);
```
irahopkinson: Minor: `onLoadingChange?.(true)` sits **outside** the `try`, so if this call itself throws, `settle` never runs and the gate wedges permanently. Today `handleLoadingChange` is only ref arithmetic so it cannot throw, but the prop is public and a consumer-supplied callback can. Moving this inside the `try` would make the balance guarantee hold unconditionally.

libs/shared-react/src/plugins/usj/LoadStatePlugin.tsx
```diff
@@ -20,13 +20,24 @@ export function LoadStatePlugin<TLogger extends LoggerBasic>({
...
+  /**
+   * Called `true` when a load starts and `false` once it settles — committed, skipped (nothing to
+   * serialize), or failed. Calls are always balanced, so a caller may count them, and `false` is
+   * never reported before the loaded document is live. Use it to defer work that has to address
+   * the loaded content.
+   *
+   * Must be referentially stable (e.g. `useCallback` with stable dependencies): it is a dependency
+   * of the load effect, so a fresh identity each render would reload the document each render.
+   */
+  onLoadingChange?: (isLoading: boolean) => void;
```
irahopkinson, line 41: Two things on this new prop:

**1. Missing from the `@param` block (lines 6-16).** Every other prop on this component has an `@param` line; `onLoadingChange` is documented only here in the type literal. That block is the file's own established convention.

**2. The stability requirement is documented but not enforced.** `onLoadingChange` is a dependency of the load effect (line 110) and `LoadStatePlugin` is publicly re-exported (`plugins/usj/index.ts:12`). A consumer passing an inline lambda reloads the document *every render* — precisely the silent undo/redo wipe that the `viewOptions` / `logger` deep-equality memos at `Editor.tsx:168-195` were written to prevent. The JSDoc warns, but nothing stops it. Taking a ref internally (or deep-comparing) would make the footgun unreachable rather than merely documented.

libs/shared-react/src/plugins/usj/LoadStatePlugin.tsx
```diff
@@ -40,32 +51,63 @@ export function LoadStatePlugin<TLogger extends LoggerBasic>({
...
+        } catch {
+          logger?.error("LoadStatePlugin: error setting editor state.");
+          settle();
+        }
       });
     } catch {
       logger?.error("LoadStatePlugin: error parsing or setting editor state.");
+    } finally {
+      // Covers the early return and any throw; the committed path settles from its own microtask.
+      if (!settlesAfterCommit) settle();
```
irahopkinson: **The skip and parse-failure paths settle with no document live, which contradicts the invariant this prop advertises.**

The prop doc (lines 32-35) promises:

> `false` is never reported before the loaded document is live.

But the `serializedEditorState == null` early return (line 77) and the parse-throw (line 104) both settle synchronously from this `finally`, with the *outgoing* document still installed. A gate that trusts the documented invariant will drain against the wrong state.

The balance guarantee itself is fine — I traced every path and each settles exactly once. It is the "never before the document is live" half that is too strong. Either narrow the wording, or add a third state so callers can distinguish "committed" from "skipped/failed".

libs/shared-react/src/plugins/usj/LoadStatePlugin.tsx
```diff
@@ -40,32 +51,63 @@ export function LoadStatePlugin<TLogger extends LoggerBasic>({
...
+    // adaptor — otherwise a caller that gates work on the load waits forever.
+    let settled = false;
+    const settle = () => {
+      if (settled) return;
+      settled = true;
+      onLoadingChange?.(false);
+    };
+    // Set once the commit path owns settling (it reports from its own microtask, after the
+    // loaded document is live); until then the `finally` below is responsible.
+    let settlesAfterCommit = false;
```
irahopkinson: Judgement call: three mutable flags now coordinate one boolean signal — `settled`, `settlesAfterCommit`, and the `if (!settlesAfterCommit) settle()` guard in the `finally`. `settlesAfterCommit` names a *responsibility handoff* rather than a fact about the load, and the protocol needs a dozen lines of comment to defend itself.

Not a blocker, but if the commit path returned a `Promise<void>` (or the settle decision were hoisted into one small named helper), the reader would not have to hold two flags and a `finally` in their head at once.

packages/platform/src/editor/Editor.tsx
```diff
@@ -249,6 +252,10 @@ const Editor = forwardRef(function Editor<TLogger extends LoggerBasic>(
         editedUsjRef.current = incomingUsj;
         // This can happen when using `applyUpdate` since `usj` won't change.
         const shouldForceReload = deepEqual(usj, incomingUsj);
+        // A load is now certain, but React hasn't re-rendered yet, so LoadStatePlugin hasn't
+        // reported it. Without this, an annotation set in this same tick would address the
+        // outgoing document and be discarded by the load (#515).
+        noteLoadRequested();
```
irahopkinson, line 467: **This is the only `noteLoadRequested()` call, but `setUsj` is not the only thing that starts a load — so the original bug is still reachable.**

A reload also fires when `viewOptions` or `logger` genuinely changes; both are dependencies of the load effect (`LoadStatePlugin.tsx:110`). The comment at `Editor.tsx:176-179` in this same file confirms that is intended:

> Comparing by value keeps the reference stable across such re-renders while still producing a new one - correctly triggering a reload - when a view option genuinely changes.

Sequence that still loses an annotation: a consumer toggles a view option and calls `setAnnotation` in the same tick. The gate is open (nothing in flight), so the op runs immediately against the current document; React then re-renders, `viewOptions` identity changes, the effect re-runs, and `setEditorState` discards the mark. That is #515 verbatim:

> the range does resolve against the outgoing document, and the load's `setEditorState` then replaces the whole state

Calling `noteLoadRequested()` wherever a reload-triggering input changes (or deriving the gate from the load inputs rather than from `setUsj`) would close this.

packages/platform/src/editor/Editor.tsx
```diff
@@ -321,18 +328,22 @@ const Editor = forwardRef(function Editor<TLogger extends LoggerBasic>(
...
+        annotationRef.current?.setAnnotation(
+          selection,
+          externalTypedMarkType(type),
+          id,
+          onClick,
+          onRemove,
+          onMouseEnter,
+          onMouseLeave,
+        ),
       );
```
irahopkinson, line 561: **Gating only the two annotation methods reorders them against the un-gated ref methods.**

`setSelection`, `insertMarker`, `insertNote`, `applyUpdate` and `formatPara` all still run immediately. So during a load, a caller doing `setAnnotation(); insertMarker();` now gets them applied in the opposite order to the one they wrote.

The description says:

> When nothing is loading they run immediately, exactly as before.

True — but the corollary is that when something *is* loading, relative ordering across the ref API changes. Worth either stating that in the `EditorRef` docs or routing all document-addressing ref methods through the same gate.

packages/platform/src/editor/Editor.tsx
```diff
@@ -321,18 +328,22 @@ const Editor = forwardRef(function Editor<TLogger extends LoggerBasic>(
...
     },
     removeAnnotation(type, id) {
-      annotationRef.current?.removeAnnotation(externalTypedMarkType(type), id);
+      // Also gated: removing an annotation that a queued call is about to add must happen after
+      // it, not before, or the removal is a no-op and the mark survives.
+      runWhenLoaded(() => annotationRef.current?.removeAnnotation(externalTypedMarkType(type), id));
     },
```
irahopkinson, line 568: The ordering argument in this comment is correct — a single FIFO queue does preserve queued-set-then-remove — but **nothing tests it**. The new test file covers `setAnnotation` only.

This is the paranext Insert-Comment path, which the header of `annotation-pending-rewrap.test.tsx` already documents as fragile, so a test for "queued set, then remove, during one load" seems worth having.

packages/platform/src/editor/annotation-before-load.test.tsx
```diff
@@ -0,0 +1,175 @@
...
+import { vi } from "vitest";
+
+const v1Text = "The heavens, the earth, and all their vast array were finished.";
+const v2Text = "On the seventh day God finished his work which he had done.";
+
+const usjGen2: Usj = {
+  type: USJ_TYPE,
+  version: USJ_VERSION,
+  content: [
+    { type: "chapter", marker: "c", number: "2" },
```
irahopkinson: **Duplicated fixture** — this is near-verbatim from `annotation-pending-rewrap.test.tsx` in the same directory: `v1Text` (its line 32), `usjGen2` (34), `createLoggerMock` (199), the four-method logger literal (213-215), and the `[...document.querySelectorAll("mark")].map(...)` shape (229).

Worth extracting to a shared test helper beside the two files rather than carrying a second copy that can drift.

packages/platform/src/editor/annotation-before-load.test.tsx
```diff
@@ -0,0 +1,175 @@
...
+const jsonPath = "$.content[1].content[1]";
+const earthStart = v1Text.indexOf("earth");
+const earthRange: AnnotationRange = {
+  start: { jsonPath, offset: earthStart },
+  end: { jsonPath, offset: earthStart + "earth".length },
+};
+
+function createLoggerMock() {
+  return vi.fn();
+}
```
irahopkinson: `createLoggerMock` returns a single mock *method*, not a logger, so the name points at the wrong thing (the call sites below assemble the logger object by hand).

The sibling file also types this more tightly: `type LoggerMock = ReturnType<typeof vi.fn<(...params: unknown[]) => void>>` (`annotation-pending-rewrap.test.tsx:197`). Here the prop is `logError: () => void`, which drops the argument types — so an assertion like `toHaveBeenCalledWith(...)` would not typecheck against the real logger signature.

packages/platform/src/editor/annotation-before-load.test.tsx
```diff
@@ -0,0 +1,175 @@
...
+    });
+
+    // Swap the document and annotate immediately, without awaiting the reload.
+    await act(async () => {
+      ref.current?.setUsj(usjGen3);
+      ref.current?.setAnnotation(earthRange, "spelling", "a2");
+    });
+
+    expect(logError).not.toHaveBeenCalled();
+    expect(markTexts()).toEqual(["earth"]);
```
irahopkinson, line 106: **This test can pass without the second load having happened.** `earthRange` resolves against *both* `usjGen2` and `usjGen3` (both contain `v1Text` at the same `jsonPath`), and the only assertions are "no error" plus `markTexts() === ["earth"]` — all of which hold if `setUsj(usjGen3)` silently did nothing.

Asserting that the Gen 3 document is actually installed (chapter "3" rendered, or `getUsj()` reflecting it) would make this pin what it claims to pin.

packages/platform/src/editor/annotation-before-load.test.tsx
```diff
@@ -0,0 +1,175 @@
...
+              "bad",
+            )
+          }
+        />,
+      );
+    });
+
+    // Deferring must not swallow genuine failures: once the document is live the annotation is
+    // attempted and fails loudly, exactly as it does today.
+    expect(logError).toHaveBeenCalled();
```
irahopkinson: `expect(logError).toHaveBeenCalled()` passes on *any* error from anywhere in the stack, including one unrelated to range resolution. The file header already names the expected message:

> the plugin only logs "Failed to find start or end node of the annotation."

Asserting that string (`toHaveBeenCalledWith`) would stop this test going green for the wrong reason.

packages/platform/src/editor/annotation-before-load.test.tsx
```diff
@@ -0,0 +1,175 @@
...
+    await act(async () => {
+      ref.current?.setUsj(usjGen3);
+      ref.current?.setAnnotation(earthRange, "spelling", "a2");
+    });
+
+    expect(logError).not.toHaveBeenCalled();
+    expect(markTexts()).toEqual(["earth"]);
+  });
+
+  it("still reports a range that can never resolve, instead of queueing it forever", async () => {
```
irahopkinson, line 141: This pins "not queued forever", which is valuable. But the description also claims the stronger property:

> A queued operation runs at the end of the load it was queued behind — never waiting on a *future* load ... That's pinned by a test.

Nothing here exercises **two sequential loads**, so that attribution property is unpinned — and per my comment on `use-load-gate.hook.ts:20-24` I believe it is also currently untrue. A test that queues an op behind load A, starts load B, and asserts the op ran against A would cover it.

packages/platform/src/editor/use-load-gate.hook.ts
henrique221: both fixed in e396b08a. `drain` re-checks the gate now, so an op queued behind load A whose drain finds load B already requested stays queued and runs once, after B — against a document that is actually live. and each deferred op runs behind its own try/catch reporting to the logger; the immediate path is deliberately still unwrapped so a throw reaches the caller exactly as it did before the gate existed. both are pinned in `use-load-gate.hook.test.ts`. the description claim was wrong and now says what the code does: an op runs at the first settle with nothing loading or requested.

packages/platform/src/editor/use-load-gate.hook.ts
henrique221: added a teardown report in e396b08a — anything still queued when the editor unmounts is logged with a count instead of vanishing. it is deferred by a microtask and re-checked, so React's dev-only double-invoke (unmount then immediate remount, with the consumer re-queueing in the second pass) isn't reported as a loss.

packages/platform/src/editor/use-load-gate.hook.ts
henrique221: e396b08a stops it being a `setUsj`-only latch: it is set from a layout effect keyed on the inputs of the plugin's reload effect, and those deps are deliberately a subset of the plugin's, so the load that clears it is always the same commit's. I skipped the dev-mode timer — a macrotask check can land before React's scheduled render, which would drain early and reintroduce the bug. the silent case is covered by the unmount report instead.

packages/platform/src/editor/use-load-gate.hook.ts
henrique221: deliberate, and it got broader rather than narrower: e396b08a keys the gate on every reload trigger, per your note on Editor.tsx. the timing change for later reloads is now written down on `EditorRef.setAnnotation`, including that only the annotation pair is deferred.

libs/shared-react/src/plugins/usj/LoadStatePlugin.tsx
henrique221: took it — c7a65e35 passes `{ discrete: true }`. for the record I could not build a case where the microtask version settles early: a microtask can't run while an update is in progress, so the nested-update path isn't reachable from there. the reachable half is narrower and still real — `$beginUpdate` only schedules the commit when that update cloned the editor state — and `discrete` removes the dependency on Lexical's ordering entirely. `LoadStatePlugin.test.tsx` now has the missing tests: drop `discrete` and three of them fail with the outgoing document still installed.

libs/shared-react/src/plugins/usj/LoadStatePlugin.tsx
henrique221: moved inside the try in c7a65e35, with a test that a consumer callback throwing on the way in still gets its balanced `false`.

libs/shared-react/src/plugins/usj/LoadStatePlugin.tsx
henrique221: both done in c7a65e35. the `@param` line is there, and the callback is held in a ref and dropped from the load effect's deps, so the footgun is unreachable rather than documented — an inline lambda can't reload the document for any consumer now. a test pins that a changed identity alone doesn't re-serialize, and that the latest callback is the one a real reload reports to.

libs/shared-react/src/plugins/usj/LoadStatePlugin.tsx
henrique221: narrowed the wording rather than adding a third state: a settle means the document this load leaves behind is live — the loaded one, or the previous one when the load was skipped or failed. that is what the gate actually needs, since a load that changed nothing leaves its queued work addressing a document that is still there. both paths are pinned in c7a65e35.

libs/shared-react/src/plugins/usj/LoadStatePlugin.tsx
henrique221: gone in c7a65e35. with `discrete` the commit path settles in a plain `finally`, so `settlesAfterCommit` and the guard disappeared — one `settled` flag, one settle per exit, and about ten lines of comment less.

packages/platform/src/editor/Editor.tsx
henrique221: you're right, and fixed in e396b08a. a layout effect keyed on the plugin's reload inputs closes the gate, and layout effects run before every passive effect and before the consumer's own, so the view-option sequence is covered. `setUsj` still notes the load itself for the window before React re-renders.

packages/platform/src/editor/Editor.tsx
henrique221: documented it on `EditorRef` in e396b08a rather than gating everything. `applyUpdate` and `removeCharacterMarker` read their result synchronously and `insertMarker` validates by throwing to its caller, so deferring those changes their contract rather than just their timing. the annotation pair stays ordered with itself.

packages/platform/src/editor/Editor.tsx
henrique221: added in 47030088, and writing it turned something up: a same-tick set-then-remove leaves the mark either way, because `AnnotationPlugin` resolves a removal through a map its node-mutation listener fills at commit, which the set hasn't reached yet. so the gate can't fix that pair, it only keeps the order — the test pins gated behaving exactly as un-gated. I rewrote the comment at this line, which claimed more than that. drain order itself is pinned in the hook test.

packages/platform/src/editor/annotation-before-load.test.tsx
henrique221: extracted to `annotation.test-helpers.tsx` in 47030088 — fixture, ranges, logger mocks, the mark readers and the real-Editor harness, both suites on it now.

packages/platform/src/editor/annotation-before-load.test.tsx
henrique221: renamed and typed in 47030088: `createLogMock` is the method, `createLoggerMock` returns the logger, both `(...params: unknown[]) => void` like the real one. the tighter type immediately caught a mistyped prop in this file.

packages/platform/src/editor/annotation-before-load.test.tsx
henrique221: right — 47030088 asserts the chapter number, so a `setUsj` that silently did nothing fails it now.

packages/platform/src/editor/annotation-before-load.test.tsx
henrique221: asserts the exact message in 47030088.

packages/platform/src/editor/annotation-before-load.test.tsx
henrique221: the claim was wrong, so I fixed both. the property is now that a queued op runs at the first settle with nothing loading or requested, and the two-load race is pinned in `use-load-gate.hook.test.ts`: queued behind A, B requested before A settles, runs once and after B. that one drives the hook directly — both loads and their settles are microtasks React owns, so the race isn't schedulable through a mounted editor.

libs/shared-react/src/plugins/usj/LoadStatePlugin.tsx
irahopkinson: Both verified. `@param onLoadingChange` at `:14`; the ref at `:49-52` with the callback dropped from the load effect's deps at `:134`.

Making it unreachable rather than documented is the better fix and I should have proposed that in the first place — a JSDoc warning only helps the consumer who reads it, and this one is publicly re-exported. `does not reload when only the callback's identity changes` pins the important half, and asserting `serializeEditorState` call counts is the right probe: it fails if someone puts the callback back in the deps, which a test on the reported values alone would not catch.

libs/shared-react/src/plugins/usj/LoadStatePlugin.tsx
irahopkinson: Verified — `LoadStatePlugin.tsx:77`, inside the `try`, with `settles when a consumer's own callback throws on the way in` pinning it. That test is the useful half: the move alone would be easy to revert without noticing.

libs/shared-react/src/plugins/usj/LoadStatePlugin.tsx
irahopkinson: Confirmed — `settlesAfterCommit` and the `finally` guard are gone, leaving one `isSettled` flag and a plain `finally` at `:130-132`. Nice second-order win from taking `discrete`: the flag existed only to describe who owned settling, and once the commit is synchronous there is no ownership question left to name.

libs/shared-react/src/plugins/usj/LoadStatePlugin.tsx
irahopkinson: You are right and I was wrong on the mechanism — worth saying plainly. A microtask cannot run while an update is in progress, so the nested-update path I pointed at is not reachable from inside that `queueMicrotask`. My citation of `Editor.tsx:369` established that Lexical *defers updates behind in-progress ones*, which is true, but it does not get you to the settle firing early from where this code sits. I over-read it.

The reason you landed on — `$beginUpdate` only schedules the commit when that update cloned the editor state — is the real one, and it is narrower than what I claimed.

Verified the fix regardless: `{ discrete: true }` at `LoadStatePlugin.tsx:126`, and your "drop it and three fail" claim is exact. I removed it and got precisely three:

```
× reports the load balanced, and the loaded document is live at the settle
× reports each reload the same way, with the new document live
× settles with the previous document live when there is nothing to serialize
```

The `settledWith.push(liveText())` shape is a good way to pin this — it asserts the thing that actually matters (which document is installed at settle) rather than a proxy for it. 8/8 green with `discrete` restored.

libs/shared-react/src/plugins/usj/LoadStatePlugin.tsx
irahopkinson: Narrowing the wording is the right resolution — better than the third state I floated. "The document this load leaves behind is live" is true on all three paths and is exactly the guarantee the gate consumes; a third state would have made every caller handle a distinction only one of them cares about.

Verified at `:33-36`, with `settles with the previous document live when there is nothing to serialize` and `settles when the adaptor throws` pinning the two paths that were the problem.

packages/platform/src/editor/annotation-before-load.test.tsx
irahopkinson: Verified — `annotation.test-helpers.tsx`, with both suites on it. The header note about it being a plain helper module rather than an export from one of the suites (so importing it does not re-register the other suite's tests) is a genuinely non-obvious trap to have written down.

packages/platform/src/editor/annotation-before-load.test.tsx
irahopkinson: Verified — `LogMock` at `annotation.test-helpers.tsx:62`, with `createLogMock` for the method and `createLoggerMock` for the logger, both `(...params: unknown[]) => void`. The split resolves the naming complaint properly rather than just renaming around it.

Worth noting that the tighter type immediately caught a mistyped prop, which is the argument for doing it.

packages/platform/src/editor/annotation-before-load.test.tsx
irahopkinson: Verified — `chapterNumbers()` asserted at `annotation-before-load.test.tsx:105` and again at `:137`. The fixture comment spelling out *why* ("otherwise a load that never happened would look identical") is what stops someone dropping the assertion later as redundant.

packages/platform/src/editor/annotation-before-load.test.tsx
irahopkinson: Fixing the claim rather than the code was the right way round here — "runs at the first settle with nothing loading or requested" is both true and the property the gate actually provides.

Agreed on the placement too: driving the hook directly is the only way to schedule that race deterministically, since through a mounted `Editor` both loads and their settles are microtasks React owns. The header note in `use-load-gate.hook.test.ts` explaining that is worth keeping.

One caveat on the two-load test specifically — it does not pin what it says it pins, for a reason that is easy to miss. Details in my reply on the `drain` thread and in a new discussion on the test file; the fix itself is correct.

packages/platform/src/editor/annotation-before-load.test.tsx
irahopkinson: Verified — `expect(logError).toHaveBeenCalledWith(NO_RANGE_LOGGED)` at `:165`. Hoisting the expected string to a shared constant is better than inlining it: the test now fails if the plugin's message changes, instead of quietly passing on some unrelated error.

packages/platform/src/editor/Editor.tsx
irahopkinson: Verified — `useLayoutEffect` at `Editor.tsx:311`, with `setUsj` keeping its own `noteLoadRequested()` at `:567` for the window before React re-renders. The layout-effect ordering argument holds: layout effects run before any passive effect in the commit, so the gate closes ahead of both `LoadStatePlugin`'s load effect and the consumer's own mount effect. That closes the view-option sequence I was worried about, which was the one where the annotation ran immediately against a document that was about to be replaced.

packages/platform/src/editor/Editor.tsx
irahopkinson: Documenting rather than gating is the right trade, and your reason is the convincing part: `applyUpdate` and `removeCharacterMarker` read their result synchronously and `insertMarker` validates by throwing, so deferring those changes their *contract*, not just their timing. Gating everything, which is what I loosely suggested, would have broken all three.

Verified at `editor.model.ts:202-207`.

packages/platform/src/editor/Editor.tsx
irahopkinson: This is the most valuable outcome in the review, and it came from writing the test rather than from either of us reasoning about it. A same-tick set-then-remove leaves the mark either way, because `AnnotationPlugin` resolves the removal through a map its node-mutation listener only fills at commit — so the gate preserves order but cannot make the pair work.

Rewriting the comment to claim only what is true is the right response. `keeps a set-then-remove pair behaving exactly as it does with no load in flight` is a good shape for it too: asserting gated equals un-gated pins the actual guarantee without baking in the current (arguably wrong) outcome as desirable.

If that pair is something a host would reasonably expect to work, it seems worth its own issue against `AnnotationPlugin` rather than anything further here.

packages/platform/src/editor/use-load-gate.hook.ts
irahopkinson: Verified — `use-load-gate.hook.ts:99-119`. The StrictMode guard is the part I would have got wrong: deferring the report by a microtask and re-checking `isTornDownRef` means the dev-only unmount/remount pair does not produce a phantom "operations dropped" error, while a real teardown has nothing to reset the flag. Both directions are pinned (`reports operations still waiting...` and `says nothing at unmount when the queue is empty`), which is what stops the guard from silently swallowing the real case later.

packages/platform/src/editor/use-load-gate.hook.ts
irahopkinson: Both verified in the code, and I ran them.

**(a) Re-check** — `use-load-gate.hook.ts:43`. I mutated it out and the suite goes red (2 failures), so the guard is genuinely load-bearing.

**(b) Isolation** — `:47-55`, per-action `try`/`catch` to the logger, with the immediate path left unwrapped at `:88-91`. Agreed that keeping the synchronous path throwing is the right call.

One thing worth knowing about the test that names the re-check, though: **its coverage is incidental.** `use-load-gate.hook.test.ts:78` sequences `noteLoadRequested()` *before* the settle, which the scheduling-time guard in `handleLoadingChange` already catches on its own — no drain is ever scheduled. It only goes red without the re-check because `renderOpenGate` leaves an unflushed drain microtask behind from its own setup.

Proof: with the re-check removed **and** an `await flushMicrotasks()` added right after `renderOpenGate()`, that test passes. So a perfectly reasonable tidy-up to the helper would silently delete the regression guard.

A sequence that pins it directly — settle first so the drain is actually scheduled with the gate open, *then* close it:

```js
act(() => {
  result.current.handleLoadingChange(true);
  result.current.runWhenLoaded(action);
  result.current.handleLoadingChange(false); // settles: drain scheduled, gate open
  result.current.noteLoadRequested();        // gate closes before the drain runs
});
await flushMicrotasks();

expect(action).not.toHaveBeenCalled();
```

I ran this one both ways: red with the re-check removed, green with it in. Raised separately on the test file so it does not get lost in this thread. Not a blocker — the fix itself is right.

packages/platform/src/editor/use-load-gate.hook.ts
irahopkinson: Good call skipping the dev-mode timer — your reason is better than my suggestion. A macrotask check landing before React's scheduled render would drain early, which is the bug rather than a warning about it.

I checked the subset claim rather than taking it on trust. Layout effect deps are `[loadTrigger, noteLoadRequested, stableLogger, usj, viewOptions]` (`Editor.tsx:311`); the plugin's reload deps are `[editor, editorAdaptor, logger, scripture, scriptureRef, viewOptions]` plus the `key={loadTrigger}` remount. Every gate-closing trigger maps to one that really does produce a load (`usj`→`scripture`, `stableLogger`→`logger`, `viewOptions`, `loadTrigger`→remount), and the ones you left out (`editor`, `editorAdaptor`, `scriptureRef`) can only cost you the same-tick window, never wedge the gate. That is the safe direction, and the comment at `:305-310` says so.

packages/platform/src/editor/use-load-gate.hook.ts
irahopkinson: Happy with it going broader. My note was only ever "this is more than #515 asked for, say so somewhere" — and now `EditorRef.setAnnotation` does, including the part I cared about most (that only the annotation pair is deferred, so ordering against the rest of the ref API is not preserved). That is the bit a host would otherwise have discovered the hard way.

irahopkinson (commented):
Re-reviewed against `7a2b8fbf`. **Both blockers are resolved**, and I verified each response against the code and by running the suites rather than taking the replies at their word.

| Was | Now |
| --- | --- |
| `drain` did not re-check the gate | `use-load-gate.hook.ts:43`; mutating it out turns the suite red |
| Settle rested on Lexical's commit ordering | `{ discrete: true }`, `LoadStatePlugin.tsx:126` |
| No per-action isolation | per-action `try`/`catch` to the logger, `:47-55`; immediate path deliberately left throwing |
| Queue vanished on unmount | teardown report with a StrictMode guard, `:99-119` |
| `setUsj`-only latch | `useLayoutEffect` on a verified subset of the plugin's reload deps, `Editor.tsx:311` |
| `EditorRef` docs silent on deferral | `editor.model.ts:202-207`, `:251-252`, including the cross-API ordering caveat |
| No `LoadStatePlugin` / gate tests | `LoadStatePlugin.test.tsx` (8), `use-load-gate.hook.test.ts` (9) |
| Duplicated fixture, weak assertions | `annotation.test-helpers.tsx`; `chapterNumbers()` and exact-message assertions |

**28/28 green** across `use-load-gate.hook.test.ts`, `annotation-before-load.test.tsx`, `annotation-pending-rewrap.test.tsx` and `LoadStatePlugin.test.tsx`. Your "drop `discrete` and three fail" claim is exact — I removed it and got precisely those three.

**One correction to my own review.** My nested-update argument was wrong: a microtask cannot run while an update is in progress, so that path is not reachable from inside the `queueMicrotask`. I cited `Editor.tsx:369` to establish that Lexical defers updates behind in-progress ones, which is true but does not get you to an early settle from where this code sits. The reason you identified — `$beginUpdate` only schedules the commit when the update cloned the editor state — is the real and narrower one. You took the fix anyway and your own tests demonstrate the reachable case.

**Two things still open, neither a blocker:**

1. The test named for the drain re-check does not actually pin it — its coverage rides on an unflushed microtask left by `renderOpenGate`. Removing the guard *and* adding a flush after setup makes it pass. Direct replacement proposed on the test file, verified red-then-green.
2. The `op` → `action` rename did not reach `use-load-gate.hook.test.ts`, which still uses `op` in six places.

Separately, the set-then-remove finding is the best thing to come out of this round, and it came from writing the test rather than from reasoning — `AnnotationPlugin` resolving removals through a commit-time map means the gate can preserve order but cannot make that pair work. Probably deserves its own issue.

Good, careful work on the responses — several of them improved on what I suggested rather than just implementing it.

> @irahopkinson+AI made 21 comments and resolved 2 discussions.
> *[Reviewable](https://reviewable.io/reviews/eten-tech-foundation/scripture-editors/531#-:-P18P6vI3wBYXepx-ZsC:b-lucvbx)* status: 0 of 9 files reviewed, 18 unresolved discussions (waiting on henrique221 and irahopkinson).

___
*[`libs/shared-react/src/plugins/usj/LoadStatePlugin.tsx` line 57 at r1](https://reviewable.io/reviews/eten-tech-foundation/scripture-editors/531#-P-lVK-PEGYv-xHXsp8G:-P18OwXhECsrZaGuY9nj:bbvmjsc) ([raw file](https://github.com/eten-tech-foundation/scripture-editors/blob/3c88029ea8735290a6f4679b76ae7959eb7433ab/libs/shared-react/src/plugins/usj/LoadStatePlugin.tsx#L57)):*

BTW style tip: use a verb to help read and recognize booleans - change to `isSettled`.
</blockquote></details>

Done in `7a2b8fbf` — `let isSettled = false;` at `LoadStatePlugin.tsx:68`, with the `settle()` closure reading it at `:70-71`. No `settled` identifier left in the file; the only remaining occurrences of the word are prose in the hook's TSDoc ("until the load carrying it has settled", "the document is settled"), which reads correctly as English rather than as a boolean name.
___
*[`packages/platform/src/editor/use-load-gate.hook.ts` line 53 at r1](https://reviewable.io/reviews/eten-tech-foundation/scripture-editors/531#-P-lZPD6AEhS_4aGjULx:-P18OyDA5REpIwS6R3tF:b-se53a9) ([raw file](https://github.com/eten-tech-foundation/scripture-editors/blob/3c88029ea8735290a6f4679b76ae7959eb7433ab/packages/platform/src/editor/use-load-gate.hook.ts#L53)):*

This repo uses `ops` specifically in reference to OT (Operational Transform operations). I thought its use here was self contained to this hook but it leaks out here and the TSDoc doesn't explain it. Change the argument to non-abbreviated `operation` or `action` or similar. That change has flow on effects to other vars.
</blockquote></details>

Done in the source, and the OT-collision reason for it is worth the change — `use-load-gate.hook.ts` now uses `action` throughout: `runWhenLoaded(action: () => void)` at `:87`, `queuedActionsRef` at `:24`, `for (const action of actions)` at `:47`. The public surface no longer leaks `op`, which was the actual complaint.

**The flow-on you predicted did not reach the new test file, though.** `use-load-gate.hook.test.ts` still uses `op` as its local in six places — `:41`, `:44`, `:47`, `:62`, `:66`, `:68`, `:75`, `:83`, `:87`, `:92`, `:101`, `:106`, `:111`, `:115`, `:122`. That file is new in this PR, so it is in scope, and it is the file a reader lands on to understand what the hook's vocabulary means.

Purely cosmetic and test-local, so I have not marked it blocking — your call whether it is worth another push, but leaving it means the term you objected to survives in the PR.
___
*[`packages/platform/src/editor/use-load-gate.hook.test.ts` line 78 at r3](https://reviewable.io/reviews/eten-tech-foundation/scripture-editors/531#-P18P1vp8apnOPnBCKIH:-P18P1vq4Pb4UliHjq0y:bh3deri) ([raw file](https://github.com/eten-tech-foundation/scripture-editors/blob/7a2b8fbfb3a15df6e480e7c074e35c3cbf537a7a/packages/platform/src/editor/use-load-gate.hook.test.ts#L78)):*
> ```XML
>   });
> 
>   it("re-checks the gate when draining: an operation waits for the load that is next to commit", async () => {
> ```

**This test does not pin the re-check it is named for.** The `drain` guard it is guarding is correct — this is only about the test.

The sequence here puts `noteLoadRequested()` *before* the settle. At that settle, `handleLoadingChange` hits its own scheduling-time guard (`use-load-gate.hook.ts:67`) and never schedules a drain at all — so the re-check inside `drain` is never reached. That guard predates this change.

It still goes red without the re-check, but for an unrelated reason: `renderOpenGate` performs a settle in its own synchronous `act()`, which schedules a drain that is still unflushed when the test body runs. That stale microtask is what the re-check catches.

Two experiments against this branch:

1. Remove `if (!isDocumentSettled()) return;` from `drain` → this test and `drains overlapping loads only once they have all settled` fail. Looks like real coverage.
2. Remove it **and** add `await flushMicrotasks()` immediately after `renderOpenGate()` → **this test passes.**

So the coverage depends on setup leakage. Adding a flush to `renderOpenGate`, or switching it to an async helper — both natural tidy-ups — would silently delete the regression guard for the one behaviour this PR's headline claim rests on.

A sequence that exercises it directly: settle first, so the drain really is scheduled with the gate open, then close the gate before the microtask runs.

```js
it("re-checks the gate when draining: a load requested after the settle defers the action", async () => {
  const { result } = renderOpenGate();
  await flushMicrotasks(); // drop the setup's own pending drain, so this pins the re-check alone
  const action = vi.fn();

  act(() => {
    result.current.handleLoadingChange(true);
    result.current.runWhenLoaded(action);
    result.current.handleLoadingChange(false); // settles: drain scheduled, gate open
    result.current.noteLoadRequested();        // gate closes before the drain runs
  });
  await flushMicrotasks();

  expect(action).not.toHaveBeenCalled();
});
```

I ran this both ways: red with the re-check removed, green with it in — so it pins the guard rather than the setup.

Worth keeping the existing test too; it covers the scheduling-time guard, which is also worth having. It is the name and comment that currently over-claim.

(Using `action` above rather than `op`, per the naming thread — the rename did not reach this file.)

irahopkinson (commented):
> @irahopkinson reviewed 9 files and all commit messages, and resolved 16 discussions.
> *[Reviewable](https://reviewable.io/reviews/eten-tech-foundation/scripture-editors/531)* status: all files reviewed, 2 unresolved discussions (waiting on henrique221).

henrique221:
Fixed in 40118a6c. All 10 load-gate tests pass. Removing the drain guard makes the new regression test fail; restoring it makes it pass.

> ___
> *[`packages/platform/src/editor/use-load-gate.hook.ts` line 53 at r1](https://reviewable.io/reviews/eten-tech-foundation/scripture-editors/531#-P-lZPD6AEhS_4aGjULx:-P18OyDA5REpIwS6R3tF:b-se53a9) ([raw file](https://github.com/eten-tech-foundation/scripture-editors/blob/3c88029ea8735290a6f4679b76ae7959eb7433ab/packages/platform/src/editor/use-load-gate.hook.ts#L53)):*

Done. Renamed the test actions.

Satisfied

> ___
> *[`packages/platform/src/editor/use-load-gate.hook.test.ts` line 78 at r3](https://reviewable.io/reviews/eten-tech-foundation/scripture-editors/531#-P18P1vp8apnOPnBCKIH:-P18P1vq4Pb4UliHjq0y:bh3deri) ([raw file](https://github.com/eten-tech-foundation/scripture-editors/blob/7a2b8fbfb3a15df6e480e7c074e35c3cbf537a7a/packages/platform/src/editor/use-load-gate.hook.test.ts#L78)):*

Done. Verified red, then green.

Satisfied

irahopkinson (commented):
Re-reviewed `40118a6c`. **Both remaining items are closed, and I have nothing further.**

One commit, one file, and it resolves both cleanly:

1. **Drain regression isolated.** Verified by mutation: `re-checks the gate when draining after another load is requested` fails with the guard removed, while `waits for a load requested before the previous load settles` now *passes* with it removed — which is the proof the setup leakage is gone and each test covers one behaviour. You also renamed the old test and dropped its over-claiming comment, which I had not asked for and which is the right call.
2. **`op` → `action`** complete in the test file; the term is now absent from the PR.

**29/29 green** across all four suites (`use-load-gate.hook.test.ts` 10, `annotation-before-load.test.tsx` 4, `annotation-pending-rewrap.test.tsx` 7, `LoadStatePlugin.test.tsx` 8).

For the record on how this went: of my original findings, one was outright wrong on its mechanism (the nested-update argument — you corrected me and were right), and one of my follow-ups was right about the defect but reached it through a hypothesis I had to disprove by experiment before I could state it properly. The substantive corrections in this review came from you writing tests, twice — the `AnnotationPlugin` commit-time map behind set-then-remove, and the `$beginUpdate` clone condition. Worth recording, since the PR description now reflects what the code does rather than what it was hoped to do.

No blockers from me. Two optional follow-ups, neither in scope here:

- An issue against `AnnotationPlugin` for the same-tick set-then-remove pair, if a host would reasonably expect it to work.
- The other half of #515 — hosts still have no awaitable "content ready" signal, only internal libs do.

@irahopkinson — your `blocking` disposition on the `op` naming thread is the last thing outstanding; the rename is complete, so it just needs your ack.

> @irahopkinson+AI made 3 comments.
> *[Reviewable](https://reviewable.io/reviews/eten-tech-foundation/scripture-editors/531#-:-P1BoAs41D0XyHBkJ1wW:b-b2yn60)* status: 8 of 9 files reviewed, 1 unresolved discussion (waiting on irahopkinson).

___
*[`packages/platform/src/editor/use-load-gate.hook.ts` line 53 at r1](https://reviewable.io/reviews/eten-tech-foundation/scripture-editors/531#-P-lZPD6AEhS_4aGjULx:-P1Bo78R2ekAg4HlLJO4:bke6fbx) ([raw file](https://github.com/eten-tech-foundation/scripture-editors/blob/3c88029ea8735290a6f4679b76ae7959eb7433ab/packages/platform/src/editor/use-load-gate.hook.ts#L53)):*
<details><summary><i>Previously, henrique221 (HenriqueCode) wrote…</i></summary><blockquote>

Done. Renamed the test actions.



> ___
> 
</blockquote></details>

Verified — `use-load-gate.hook.test.ts` is fully renamed in `40118a6c`; `grep` for `\bop\b|\bops\b` across the file returns nothing. Combined with the source rename in `7a2b8fbf`, `op` is gone from the PR entirely, so the OT-vocabulary collision @irahopkinson raised is closed on my side.

Note this thread still carries @irahopkinson's own `blocking` disposition, which I cannot clear on their behalf — it needs their ack to fully resolve.
___
*[`packages/platform/src/editor/use-load-gate.hook.test.ts` line 78 at r3](https://reviewable.io/reviews/eten-tech-foundation/scripture-editors/531#-P18P1vp8apnOPnBCKIH:-P1Bo6A2AHwz3jp5xUih:bfmboue) ([raw file](https://github.com/eten-tech-foundation/scripture-editors/blob/7a2b8fbfb3a15df6e480e7c074e35c3cbf537a7a/packages/platform/src/editor/use-load-gate.hook.test.ts#L78)):*
<details><summary><i>Previously, henrique221 (HenriqueCode) wrote…</i></summary><blockquote>

Done. Verified red, then green.
</blockquote></details>

Verified in `40118a6c`, and you solved this better than I proposed.

I suggested adding an isolated test. You did that *and* fixed the original: renaming it to `waits for a load requested before the previous load settles`, dropping the `#515` claim from its comment, and adding the flush after `renderOpenGate()`. So the old test now honestly covers only the scheduling-time guard, and the new one covers the drain re-check. Two separate behaviours, two separate tests — which is the outcome I wanted but had not thought to ask for.

Mutation run against this commit:

| Test | Guard removed |
| --- | --- |
| `re-checks the gate when draining after another load is requested` | **fails** — real, isolated coverage |
| `waits for a load requested before the previous load settles` | **passes** — confirming the leakage is gone and it now tests only what its name says |

That second row is the part that proves the fix: before your change that test failed under the mutation for the wrong reason, and now it does not. 10/10 green with the guard restored.

One harmless leftover, no action needed: `drains overlapping loads only once they have all settled` did not get a flush, so it still fails under the mutation via the setup's pending drain. That is now just redundant coverage rather than the only coverage, so it does not matter — flagging it only so it is not a surprise if someone adds a flush there later and sees the failure count drop from 2 to 1.

irahopkinson (approved):
:lgtm: Thanks for this fix

> @irahopkinson reviewed 1 file and all commit messages, made 1 comment, and resolved 1 discussion.
> *[Reviewable](https://reviewable.io/reviews/eten-tech-foundation/scripture-editors/531#-:-P1BopD9789DN7vQ7FSN:b-3y7yzy)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved.

PR #532, opened 2026-08-08
PT-4201: Standard-view copy/paste USFM fidelity

tjcouch-sil:
Implements PT-4201 (WI-14): copy/paste USFM fidelity in Standard view. Targets `standard-view-pt-4187` (stacked; rebased onto its tip b6ce230d).

## What changed

- **Copy (S1):** new `$selectionToUsfmText` walker for the Standard-view `text/plain` clipboard leg — emits note callers (`\f - …`, previously dropped and rewritten to `+` on re-paste), suppresses display-only separator NBSPs (no more phantom spaces after `\xo*`/`\xt*`/before `\f*`), traverses `AttributeRunNode`-wrapped runs transparently. Byte pins from the 2026-08-07 live repros.
- **Paste (S3):** `$handlePasteForStandardView` now claims every external paste (no same-namespace Lexical payload) — kills the live tilde-corruption bug (`\f~ \fr~1:1 …`) caused by the async clipboard read dropping the private Lexical flavor. Three-part positional NBSP rule (after-marker → space, before-marker → drop, interior → `~`), `\r` normalization, `\c`/`\id` strip anywhere in a line (a pasted `\c` previously poisoned the save loop with "Multiple chapter markers present"). Shared verbatim by the in-note multi-line paste path.
- **Paste-scoped paragraph-prefix dedup:** a pasted line owning its own paragraph marker no longer leaves a stray empty host paragraph; typed input deliberately keeps the pre-existing split-with-empty behavior (P9 parity), pinned both ways.
- **Semantics documented:** `docs/superpowers/specs/2026-08-06-clipboard-semantics.md` (S1–S7, P9 reference, accepted asymmetries, deferred items, test-mapping table).
- **Pins:** clipboard copy fidelity, marker-paste fidelity, plain-text-paste equivalence, corpus round-trip sweep (4 documented lossy constructs incl. a structural `\esb` Tier-2 granularity deferral), hidden-view gate tests, WI-2 char-span boundary armor, one-undo-step pins.

## Verification

- Full workspace suite green: 96 files / 2577 tests, typecheck + lint clean (post-rebase, cache-skipped re-run).
- **Outstanding:** live in-app QA and the paranext-core E2E clipboard spec run are pending an app slot (companion PR: paranext-core `standard-view-pt-4201`).

AI-assisted — Claude Code (Fable 5), full task/review ledger in `.superpowers/sdd/` (untracked).

- - -
This change is [Reviewable](https://reviewable.io/reviews/eten-tech-foundation/scripture-editors/532)


tjcouch-sil:
# Code Review Summary

**Branch**: `standard-view-pt-4201` (both repos)

**Base**: `main` — paranext-core `origin/main`, scripture-editors `eten-tech-foundation/main`

**Date**: 2026-09-15

**Review model**: Claude Opus 5 (1M context)

**Files changed**: 1 (paranext-core, committed) + 40 (scripture-editors)

**PRs**: [paranext-core #2656](https://github.com/paranext/paranext-core/pull/2656) · [scripture-editors #532](https://github.com/eten-tech-foundation/scripture-editors/pull/532)

## Overview

PT-4201 makes Standard-view copy/paste round-trip USFM faithfully. The editor half (scripture-editors #532) makes `text/plain` byte-faithful USFM, stops an empty copy from overwriting the clipboard with `@lexical/clipboard`'s `#` placeholder, keeps opaque constructs (figure/sidebar/periph/ref/optbreak) intact across the internal clipboard flavor, replaces a blanket NBSP→`~` mapping with a positional rule, strips pasted `\c`/`\id` before they poison the save loop, and routes attribute-run pastes as plain text. The paranext-core half (#2656) adds one Playwright spec driving that round trip through the real OS clipboard.

This was an automated review: five `review-analyzer` passes (api/style/compliance across both repos), each finding verified against the code before being acted on. Every finding below is fixed, refuted, dismissed with a reason, or explicitly left open. Two of the analyzers' Important findings turned out to be genuine defects the branch's own 4,200-test suite could not see, and both now have pins that go red when the fix is reverted.

## API Changes

**paranext-core**: None. The only committed file is a new e2e spec.

**scripture-editors** — the published surface is `packages/platform` only; `libs/shared`, `libs/shared-react`, `libs/test-data` are `"private": true`.

- `packages/platform/src/index.ts` and `packages/platform/etc/platform-editor.api.md`: **no change**.
- `EditorRef.copy()` / `EditorRef.cut()`: signatures unchanged, **behavior changed** — with nothing selected they now leave the clipboard untouched instead of writing `#`.
- Private workspace libs: `shared-react` gained `copySelection`, `cutSelection`, `registerEmptyCopyGuard`, `$hasCopyableSelection`; `shared` exported `tableCellMarkerWithSpan`.
- Package-internal: `normalizePastedNbsp`, `stripPastedChapterAndBookId`, `$selectionToUsfmText` added; `PastePayload` narrowed to `{text, isInternal}`; `MarkerEditContext` gained `pasteRebuildArmed` / `pastePendedKeys` / `structureProtectionMode`; `isParaKindMarker` / `isCharKindMarker` moved to a new leaf module `markerKind.utils.ts`.

## Findings

None.

**scripture-editors**

- [x] **`stripPastedChapterAndBookId` destroyed pasted user content past the chapter number.** `/\\(?:c|id)(?![\w-])[^\n\\]*/g` ran to the next `\` or newline, so pasting `x \c 5 y` stripped to `x ` — the `y` silently dropped. The tokenizer takes only the next whitespace-delimited word as the chapter number (`getNextWord`, `usfmFragmentToUsj.ts:372-376`), so everything after it is ordinary content. _(fixed during review: split into `CHAPTER_TOKEN` (marker + separator + one word, stopping at a backslash) and `BOOK_ID_TOKEN` (unchanged — an `\id` line's whole remainder genuinely is payload); updated the pin in `markerPasteFidelity.test.tsx` that had encoded the loss, and added direct pins. Reverting the regex turns the new pin red with `expected 'x ' to be 'x  y'`.)_
- [x] **`Editor.tsx`'s `structureProtectionMode` pass-through had no test — deleting the line left all 4,234 tests green.** Both `Editor` and `MarkerEditPlugin` default the prop to `"off"`, so a dropped pass-through silently starves `StructureKeyboardPlugin`'s sanitizer on a protected document's paste. _(fixed during review: threaded the option through `mountStandardViewEditor` and added a real-`Editor` paste pin. Verified falsifiable — deleting `Editor.tsx:1160` turns it red while the plugin-level pin stays green. Note the option lives on `EditorOptions`, not as a top-level `<Editor>` prop.)_
- [x] **The empty-copy rule was re-implemented inline instead of reusing the predicate this branch created.** `$handleCopyForStandardView` spelled `(!selection || selection.isCollapsed())` — provably the exact negation of shared-react's `$hasCopyableSelection()` — leaving two layers whose agreement is load-bearing resting on two independently-spelled predicates in two packages. _(fixed during review: the platform layer now calls `!$hasCopyableSelection()`, which also gives that export its first consumer outside its own module.)_
- [x] **Four new marker regexes bypassed `markerName.pattern.ts` ("the single authority"), and two of the spellings already disagreed.** `AFTER_MARKER_NBSP`/`BEFORE_MARKER_NBSP` used `[a-z0-9-]` + `i`; `LEADING_MARKER_LITERAL` and the `\c`/`\id` token used `[\w-]`. `\w` includes `_`, so `\my_marker` was a marker to one and not the other — its separator NBSP survived into content as a literal `~` exactly where the tokenizer expects a space. _(fixed during review: all four derive from `ENGINE_MARKER_NAME_BYTES`. New pin goes red on revert with `expected '\my_marker~light' to be '\my_marker light'`.)_
- [x] **Spec-document breadcrumbs and provenance narration in comments** — `Deferred item 2`, `S2's own documented case`, `measured`, `confirmed by reading LexicalClipboard.dev.js`, `Regression classes closed by`, `inherited rather than introduced`, `rather than papered over`. The repo's own `CLAUDE.md:275` forbids spec section numbers in code. _(fixed during review: rules kept, provenance and IDs stripped, across `whitespaceDisplay.plugin.utils.ts`, `UnknownNode.ts`, `attributeContextPasteFidelity.test.tsx`.)_
- [x] **`$`-sigil on two pure `string → string` functions.** `CLAUDE.md:269` requires every exported `$`-prefixed function to document its calling context; neither had one, because neither has a context. _(fixed during review: renamed to `normalizePastedNbsp` / `stripPastedChapterAndBookId` across sources, tests, and the semantics doc.)_

**paranext-core**

- [ ] **The spec cannot pass against the editor `main` builds, so merge order matters.** `$stripPastedChapterAndBookId` and `$selectionToUsfmText` do not exist on `origin/platform-yalc`. CI is unaffected (`test.yml` runs only `test:e2e:smoke`), but after this merges to `main`, `npm run test:e2e:isolated scripture-editor` fails for every developer until #532 is on `platform-yalc`. **Land #532 and move `platform-yalc` first.**
- [x] **Hard-coded `Control+` chords made the spec unrunnable on macOS.** `Control+C/V/Z` and `Control+Home/End` are OS editing chords, not app accelerators — the only other bare `Control+` uses in the suite (`verse-navigation-shortcuts.spec.ts`) genuinely are accelerators. _(fixed during review: per-platform key constants following `navigation-history.spec.ts:43-45`.)_
- [x] **The copy step could pass on clipboard contents it did not write.** Nothing cleared the OS clipboard before `Control+C`, and a retry attempt that clears the poll then fails the one-shot `\p ` check leaves real USFM there for the next attempt. _(fixed during review: a sentinel is written immediately before the copy.)_
- [x] **The undo step was three negative assertions with no positive control.** An empty or mid-replacement editor satisfies all of them, and the PDP's debounced save echoes by replacing content wholesale. _(fixed during review: added `toContainText('Amittai')` as the positive control, plus the missing closer-count assertion.)_
- [x] **No negative control (the Revert Test).** _(addressed during review by static analysis rather than a rebuild: `$selectionToUsfmText` and `$stripPastedChapterAndBookId` are absent from `origin/platform-yalc` while `$handleCopyForStandardView` is present, and the pre-fix paste gate returns `false` for any payload without an NBSP. So **step 4 (`\c` strip) is the true fix-pin; steps 1–3 are characterization pins** that likely pass pre-fix too. Recorded here so nobody mistakes the spec for a four-way regression net.)_

### Minor — Consider

**paranext-core**

- [x] `.first().toBeAttached()` could not catch glyph doubling on paste _(fixed: `toHaveCount(1)` on both opener and closer — Jonah has zero `\nd` going in, so the exact count is free)_
- [x] The undo step checked only the opening span, so a half-reverted pair would pass _(fixed: both halves asserted back to 0)_
- [x] The OS clipboard leak was undocumented _(fixed: new bullet in `e2e-tests/CLAUDE.md`'s "State that leaks between runs" — it is machine-wide and **not** contained by `isolatedProjectRoot` or the temp `userDataDir`, unlike everything else the spec touches)_
- [x] `tests/isolated/README.md` documented every subset except `scripture-editor/` _(fixed; pre-existing, one line)_
- [x] Backward-facing comment: `"Yahweh" — the word used in earlier drafts of this spec` _(fixed: keeps the durable fact that scripture vocabulary cannot serve as a sentinel, drops the drafting history)_
- [ ] **The `Control+Z` is not synchronized with the PDP's ~700ms debounced save/echo.** Mitigated by the positive control added above, but the ordering hazard is real and unresolved. See "In-Review Quality Check" for what two live runs showed.
- [ ] ~~The `press('Home')` between paste and undo should be dropped~~ _(dismissed: settles are history-merged via `$settleWithoutOwnUndoEntry`, which the analyzer itself verified, so it cannot split the undo step; the comment already marks it belt-and-braces.)_
- [ ] ~~`dev-packages.json` redirects every developer install and CI leg to a feature branch~~ _(refuted: working-tree-only — `git diff <MERGE_BASE> HEAD -- dev-packages.json` is empty, nothing staged. Both core analyzers looked at this and the second reached the same conclusion. Deliberately excluded from the review commit.)_

**scripture-editors**

- [x] `$hasCopyableSelection` was an export with no consumer outside its own module _(resolved by the reuse fix above)_
- [x] `isParaKindMarker` / `isCharKindMarker` were byte-identical but for one `MarkerType` comparison, in a module that exists precisely so the two tiers cannot disagree _(fixed: one private `isKindMarker` body, two thin wrappers; the both-kinds-true unknown fallback is preserved and now documented)_
- [x] `$selectionToUsfmText` mirrors `RangeSelection.getTextContent()` without naming the version it was mirrored from _(fixed: pinned to lexical 0.43.0 with an instruction to re-check on upgrade)_
- [x] **The clipboard semantics doc lived in a gitignored directory.** `.gitignore:67` ignores `docs/superpowers`; the file was tracked only by a force-add, so future edits to the durable contract would never appear in `git status`. _(fixed: `git mv` to `docs/clipboard-semantics.md`, alongside `docs/standard-view-invariants.md`, matching the repo's flat topic-named docs; the one code reference now carries the real path.)_
- [x] Deferred item 6 pointed at `docs/superpowers/specs/2026-07-07-standard-view-followups.md`, which does not exist _(fixed: replaced with a plain "tracked separately")_
- [x] The test-mapping table still described the E2E spec as "written, not yet run in this task" _(fixed: it runs and passes)_
- [ ] ~~`UnknownNode.ts`'s new block is 88% comment~~ _(dismissed: the block encodes a genuinely subtle invariant — why child-count is the test, and why excluding a wrapper makes `$appendNodesToJSON` hoist its children into a convincing lie. This repo is deliberately comment-heavy and the reasoning is not inferable from the code.)_
- [ ] `$noteCallerText`'s `caller || GENERATOR_NOTE_CALLER` fallback is untested — no fixture has a note without an explicit caller
- [ ] `$isNoteInternalDisplaySeparator` and `$collapsedNoteCategoryBytes` run the same three-step probe, differing only in the last check's sign — deferred as a refactor beyond this PR's scope
- [ ] `$handlePasteForStandardView` has four positional params, three optional only so tests can call it with one — an options object would read better; deferred
- [ ] `whitespaceDisplay.plugin.utils.ts` is now 787 lines with 10 of 11 exports being clipboard concerns — splitting out `standardViewClipboard.utils.ts` would match the module boundary to the concern; deferred (large diff churn on a branch already under review)
- [ ] **`EditorRef.copy()`/`cut()` behavior change deserves a release note** — a host that calls `copy()` at a collapsed caret used to get a clipboard write (`#`) and now gets nothing. Strictly better, but there is no compile-time signal for a consumer that built a workaround.

### Template Propagation

#### Shared Regions Modified

None. No changed file in either repo contains a `#region shared with` marker.

#### Extension Config Changes

- [n/a] paranext-core: no files under `extensions/` changed.
- [n/a] scripture-editors: no `extensions/` tree, and no build/tooling config changed.

## Positive Observations

- **Load-bearing claims in the doc comments hold up against the installed Lexical source.** An analyzer independently re-derived them: `$appendNodesToHTML` returns early on a null `element` before consuming the `excludeFromCopy` value; `$appendNodesToJSON` passes `'html'` and hoists excluded nodes' children; `copyToClipboard(editor, null)` really does append `createTextNode('#')` while `$copyToClipboardEvent` bails before `preventDefault()` on a collapsed selection. That is the exact `#`-leak mechanism the guard closes.
- **Falsifiability across the new tests is unusually good.** Copy-side pins assert exact full USFM strings the old blanket `getTextContent().replaceAll(NBSP, " ")` cannot produce, and three places carry explicit negative controls that isolate the fix as the only variable (`tier2Rebuild.utils.test.tsx:405`, `markerPasteFidelity.test.tsx:521`, `attributeContextPasteFidelity.test.tsx:251`).
- **The 2SA fixtures are exact regeneration output.** Re-running `GENERATE_TEST_DATA=1` plus prettier reproduced the committed files byte-identically, and `2sa.lexical.hidden.ts` correctly shows no change.
- Putting the empty-copy guard on `COPY_COMMAND`/`CUT_COMMAND` at `COMMAND_PRIORITY_LOW` is the right seam — one registration covers the shortcuts, the context menu and `EditorRef`, and it verifiably sits above rich-text's `EDITOR`-priority fallback and below Standard view's own `HIGH` claim.
- `markerKind.utils.ts` is a genuinely good extraction: it removes the Tier 1 ↔ Tier 2 import cycle and documents why a narrower `type === MarkerType.Paragraph` check would have been wrong.
- The e2e spec's step 4 is exemplary — a mid-line `${token} \c 99 ` payload matching the deliberately-unanchored strip regex, with three independent probes, one of them (the toolbar reference) read entirely outside the editor DOM.
- Every locator in the spec resolves against real editor code, and `nd` is a genuinely clean sentinel (the bundled Jonah SFM contains zero `\nd`), so the attach and zero-count assertions are non-vacuous.
- No `eslint-disable`, `@ts-expect-error`, type assertions, `console.*`, `.only`, or untracked TODOs added anywhere in either diff.

## Interview Notes

Run in automated mode at the author's request — no interview was conducted. The author asked for each finding to be considered on its merits, verified, and then fixed if small-to-medium, low-risk and worthwhile; otherwise refuted or dismissed with a reason. Judgment calls made autonomously:

- **Fixed** anything that was a verified correctness defect, a stated-convention violation, or a missing/weak test — including two defects the existing 4,200-test suite could not see (the `\c` over-strip and the untested `Editor` pass-through), both now pinned with tests confirmed to go red on revert.
- **Deferred** pure refactors with large diff churn and no correctness content (the module split, the options-object signature, the note-separator merge) as out of scope for a branch already in review.
- **Did not re-home PR #532.** Active scripture-editors development moved to `paranext/scripture-editors` under PT-4500 — that repo carries the `pt-4313`, `pt-4370` and `pt-4615` branches and commits `packages/*/dist` — while #532 still targets `eten-tech-foundation`. Moving a PR across repos is the author's call and the transfer guide is theirs to exercise.
- **Did not widen shared e2e timeouts** in response to the flake found below; the gate is shared infrastructure used by every isolated spec.

## In-Review Quality Check

**scripture-editors** — all gates green after every change:

- `nx run-many -t test` (`--skip-nx-cache --pool=forks --maxWorkers=4`): **shared 697 passed · shared-react 1674 passed / 1 skipped · platform-editor 1864 passed / 4 skipped**. Baseline before this review was 697 / 1674+1 / 1857+4, so the review added **7 tests** and broke none.
- `nx run-many -t typecheck lint`: clean — **0 errors**. The 28 lint warnings are pre-existing `no-console` / `no-loop-func` in files this branch does not touch (cross-referenced against the changed-file list).
- `nx format:check`: clean.
- Each new pin was **verified falsifiable** by reverting its fix and watching it go red with the expected message.

**paranext-core**:

- `tsc -p e2e-tests/tsconfig.json`, `eslint`, `prettier --check`: all clean.
- **The e2e spec was run twice against a live app.** Run 1: **passed, 4.0m**. Run 2: **failed then passed on retry — Playwright reports it flaky.**
- **The flake is not in the clipboard logic.** It failed at `clipboard-usfm-round-trip.spec.ts:77` (`makeSampleProjectEditable()`), in the *shared* backend-readiness gate at `e2e-tests/fixtures/helpers.ts:942`: `PAPI method "object:platform.Paratext-pdpf.getProjectDataProviderId" not listed in rpc.discover within 120000ms`. That is the C# data provider failing to register its PDP factory in time on a machine under heavy load, before any clipboard step runs. It affects every isolated spec that calls that helper, and is not caused by this branch.
- Worth noting for the budget: `test.slow()` gives 360s, and a passing run took 240s while the setup gate alone can consume 120s. The margin is thinner than the previously-recorded 1.7m suggests.

## Suggested Review Focus

- [ ] **Merge order is a hard constraint**: scripture-editors #532 must land *and* reach `platform-yalc` before paranext-core #2656, or `npm run test:e2e:isolated scripture-editor` goes red on `main` for every developer. CI will not catch this (only `test:e2e:smoke` runs there).
- [ ] **Which repo should #532 land in?** PT-4500 moved active development to `paranext/scripture-editors`; this PR still targets `eten-tech-foundation`. Transferring it also means rebuilding the committed `packages/*/dist` that the paranext repo carries and `eten-tech-foundation` does not.
- [ ] **The narrowed `\c` strip is a deliberate behavior change.** `x \c 5 y` now keeps the `y`. Confirm that keeping trailing content is what you want in every paste shape, not just the tested ones.
- [ ] **The e2e flake.** It is a pre-existing shared-fixture readiness gate, but this spec's 360s budget leaves less headroom than the others. Decide whether to raise the gate, or accept the retry.
- [ ] **`EditorRef.copy()`/`cut()`** is a host-visible behavior change with no compile-time signal — does it need a release note?
- [ ] Remaining documented deferrals in `docs/clipboard-semantics.md` (items 10–14: the stranded-decorator class across figure/table/sidebar/periph/ref, table/milestone corpus losses, the dropped html-carrier NBSP rule) are unchanged by this review and still open.

tjcouch-sil:
Moved to **https://github.com/paranext/scripture-editors/pull/13**, where Paratext 10 now develops `platform-editor` (see `docs/transferring-work-from-eten-tech-foundation.md`).

The branch was rebased with `--onto origin/main` from this PR's base (`768f6476`), replaying only its own 19 commits — clean, no conflicts, every PT-4201 change verified intact. The new PR additionally rebuilds the committed `packages/*/dist` that the paranext repo carries and this one does not.

The `standard-view-pt-4201` branch here is left in place rather than deleted, so this trail stays followable.

